PR #23, opened 2024-01-15
Stackdriver - using opentelemetry 0.21.0 and opentelemetry_sdk 0.21.2

ivan-brko:
Stackdriver - using opentelemetry 0.21.0 and opentelemetry_sdk 0.21.2 and fixing compile issues

## Fixes
Removes OTEL git Cargo dependencies and replaces them with Cargo crates.
Fixes compile issues.

## Changes
N/A

### DIFF
--- a/opentelemetry-stackdriver/Cargo.toml
+++ b/opentelemetry-stackdriver/Cargo.toml
@@ -16,10 +16,8 @@ hex = "0.4"
 http = "0.2"
 hyper = "0.14.2"
 hyper-rustls = { version = "0.24", optional = true }
-# TODO: Replace with opentelemetry version before release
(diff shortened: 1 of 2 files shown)


djc:
As mentioned in my comment in the issue. I think we should stick with compatible git revisions from the upstream crates for now, instead of reverting changes made to become compatible with the next version of upstream crates. If we end up actually merging fixes and deciding that we need/want to do a maintenance release, we can always backport the fixes to a release branch.

ivan-brko:
Aaah, ok, then I misread what you wrote in the thread 🤦🏻 
I didn't realize that the changes in this repo are waiting for an upstream release, thought that it was the other way around (`stackdriver` using newer dependencies than it supports in the code). 

I'll close the PR then, it doesn't make sense to have it.

Issue #3, opened 2020-11-25
Add Google Trace propagator


lucacasonato:
Should this be added to `opentelemetry` or `opentelemetry-contrib`? Unless anyone else wants to implement this, I will do it.

Go implementation for opencensus: https://github.com/census-instrumentation/opencensus-go/blob/v0.22.5/exporter/stackdriver/propagation/http.go

cijothomas:
Closed via https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/25

PR #25, opened 2024-01-17
Stackdriver - adding support for Google X-Cloud-Trace-Context propagation format

ivan-brko:
## Fixes 
Issue https://github.com/open-telemetry/opentelemetry-rust-contrib/issues/3

## Changes

Added `GoogleTraceContextPropagator` to `opentelemetry-stackdriver` crate, behind a feature flag `propagator`.
Implementation is based on the [official GCP golang library implementation](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/propagator/propagator.go) and Rust `opentelemetry` [implementation of W3C Propagator](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/propagation/text_map_propagator.rs).

### DIFF
--- a/opentelemetry-stackdriver/src/lib.rs
+++ b/opentelemetry-stackdriver/src/lib.rs
@@ -51,6 +51,9 @@ use yup_oauth2::authenticator::Authenticator;
 #[allow(clippy::derive_partial_eq_without_eq)] // tonic doesn't derive Eq for generated types
 pub mod proto;
 
+#[cfg(feature = "propagator")]
+pub mod google_trace_context_propagator;
+
 const HTTP_HOST: &str = "http.host";
 const HTTP_PATH: &str = "http.path";
 const HTTP_USER_AGENT: &str = "http.user_agent";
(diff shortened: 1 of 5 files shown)


opentelemetry-stackdriver/src/lib.rs
```diff
@@ -51,6 +51,9 @@ use yup_oauth2::authenticator::Authenticator;
 #[allow(clippy::derive_partial_eq_without_eq)] // tonic doesn't derive Eq for generated types
 pub mod proto;
 
+#[cfg(feature = "propagator")]
```
cijothomas, line 54: curious, why this needs to be under feature flag? The user must set the global propagator themselves anyway for this to take effect, right?
(Or feature flag is done to keep extra dependencies as optional? )

ivan-brko: Yeah, I placed it under the feature flag so that users who don't need it don't unnecessarily get new dependencies in their projects after upgrading to a new library version.

TommyCpp (approved):
Looks good. Thanks

PR #39, opened 2024-02-26
opentelemetry 0.22 et al updates

djc:


shaun-cox (approved):
Nice cleanup, thank you.

cijothomas:
@djc @hdost We need changelog entry for the crates before releasing them? I think we need to release all the crates, since we never did a release since moving to contrib..

hdost:
I had that thought after merging we can address during the release of them.

cijothomas:
> I had that thought after merging we can address during the release of them. I'll come through with a cleanup PR so we can get a release ready.

Thanks! That'll also confirm that you have the necessary rights to publish as well.

PR #47, opened 2024-03-05
stackdriver: parse trace context without regex

djc:
## Changes

Simplify parsing trace context to avoid the complexity of regular expressions (and the added dependency).

Follow-up for #25.


cijothomas (approved):
A changelog entry might be good to callout the reduces deps.

djc:
The code that is changed here was added in #25 (from January), which was merged after the latest release (0.18.0, from November). So from a downstream perspective, there is no change in this PR.

lalitb (approved):

PR #59, opened 2024-04-04
feat: add host.arch resource attribute

rogercoll:
# Changes

Adds the "host.arch" resource attribute to the Host detector.

### DIFF
--- a/opentelemetry-resource-detectors/src/host.rs
+++ b/opentelemetry-resource-detectors/src/host.rs
@@ -63,7 +63,6 @@ impl Default for HostResourceDetector {
     }
 }
 
-#[cfg(target_os = "linux")]
 #[cfg(test)]
 mod tests {
     use super::HostResourceDetector;
@@ -71,6 +70,7 @@ mod tests {
     use opentelemetry_sdk::resource::ResourceDetector;
     use std::time::Duration;
 
+    #[cfg(target_os = "linux")]
     #[test]
     fn test_host_resource_detector() {
         let resource = HostResourceDetector::default().detect(Duration::from_secs(0));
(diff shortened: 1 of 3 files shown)


opentelemetry-resource-detectors/src/host.rs
```diff
@@ -56,18 +67,24 @@ impl Default for HostResourceDetector {
...
             .get(Key::from_static_str(
                 opentelemetry_semantic_conventions::resource::HOST_ID
             ))
-            .is_some())
+            .is_some());
+        assert_eq!(
+            resource.get(Key::from_static_str(
+                opentelemetry_semantic_conventions::resource::HOST_ARCH
+            )),
+            Some(Value::from("x86_64"))
```
lalitb, line 87: nit - do we need this test to run for windows too ?

opentelemetry-resource-detectors/src/host.rs
```diff
@@ -4,30 +4,41 @@
...
-/// This detector looks up the host id using the sources defined
-/// in the OpenTelemetry semantic conventions [`host.id from non-containerized systems`].
+/// This resource detector returns the following information:
 ///
-/// [`host.id from non-containerized systems`]: https://opentelemetry.io/docs/specs/semconv/resource/host/#collecting-hostid-from-non-containerized-systems
+/// - [`host.id from non-containerized systems`]: https://opentelemetry.io/docs/specs/semconv/resource/host/#collecting-hostid-from-non-containerized-systems
+/// - Host architecture (host.arch).
 pub struct HostResourceDetector {
     host_id_detect: fn() -> Option<String>,
 }
```
lalitb, line 20: I understand it is not related to this PR, but using a function pointer is bit unconventional for Rust.  Closure or trait object could be more idiomatic. Something like (not tested):

```rust
pub struct HostResourceDetector {
    host_id_detect: Box<dyn HostIdDetector>,
}

trait HostIdDetector {
    fn detect(&self) -> Option<String>;
}
#[cfg(target_os = "linux")]
struct LinuxHostIdDetector;

#[cfg(target_os = "linux")]
impl HostIdDetector for LinuxHostIdDetector {
    fn detect(&self) -> Option<String> {
       Some("my_host_id")
   }
}
#[cfg(not(target_os = "linux"))]
struct NonLinuxHostIdDetector;

#[cfg(not(target_os = "linux"))]
impl HostIdDetector for NonLinuxHostIdDetector {
    fn detect(&self) -> Option<String> {
        None
    }
}
```
But not for this PR. Commented just that we know it can be improved :)

lalitb (approved):
LGTM. Thanks for the PR :)

opentelemetry-resource-detectors/src/host.rs
rogercoll: Definitely, I just added a platform agnostic test: https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/59/commits/35b7860c16a03beb3b945e251028ebc22bdea4bc

hdost (approved):

PR #84, opened 2024-06-25
Implement journald exporter for OpenTelemetry

lalitb:
## Changes

This adds `Journald` log exporter for OpenTelemetry, allowing logs to be sent to Journald. Note that this exporter is experimental, and breaking changes are expected. The performance and stability improvements are required, and contributions are welcome.

### Features:

- Export OpenTelemetry logs to journald.
- Optionally serialize logs and attributes as JSON.
- Configurable message size limit based on the journald configuration : As of now this is just used to return an error, in future `memfd` can be used for sending large messages.
- Configurable attribute prefix.

### Dependencies:
This implementation relies on the `libsystemd` library for sending log entries to the journald daemon. It uses `sd_journal_sendv` API method provided by the library.

### DIFF
--- a/opentelemetry-journald-logs/Cargo.toml
+++ b/opentelemetry-journald-logs/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "opentelemetry-journald-logs"
+version = "0.1.0"
+edition = "2021"
+keywords = ["opentelemetry", "journald", "logs", "tracing"]
+description = "OpenTelemetry logs exporter for journald"
+repository = "https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-journald-logs"
+homepage = "https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-journald-logs"
+license = "Apache-2.0"
+
+[dependencies]
+opentelemetry = { workspace = true, features = ["logs"] }
+opentelemetry_sdk = { workspace = true, features = ["logs"] }
+libc = "0.2"
+async-trait = "0.1"
+serde = { version = "1.0", features = ["derive"], optional =true}
+serde_json = {version = "1.0", optional = true}
+
+[dev-dependencies]
+opentelemetry-appender-tracing = { workspace = true }
+tracing = { version = "0.1", default-features = false, features = ["std"] }
+tracing-core = "0.1.31"
+tracing-subscriber = { version = "0.3.0", default-features = false, features = ["registry", "std"] }
+
+[features]
+logs_level_enabled = ["opentelemetry/logs_level_enabled", "opentelemetry_sdk/logs_level_enabled"]
+default = ["logs_level_enabled"]
+json = ["serde", "serde_json"]
--- a/opentelemetry-journald-logs/examples/basic-logs.rs
+++ b/opentelemetry-journald-logs/examples/basic-logs.rs
@@ -1,0 +1,29 @@
+// cargo run --example basic-logs
+
+use opentelemetry_appender_tracing::layer;
+use opentelemetry_journald_logs::JournaldLogExporter;
+use opentelemetry_sdk::logs::LoggerProvider;
+use tracing::info;
+use tracing_subscriber::prelude::*;
+
+fn init_logger() -> LoggerProvider {
+    let exporter = JournaldLogExporter::builder()
+        .identifier("opentelemetry-journal-exporter")
+        .message_size_limit(4 * 1024)
+        .attribute_prefix(Some("OTEL_".to_string()))
+        .json_format(true) //uncomment to log in json format
+        .build()
+        .expect("Failed to build JournaldLogExporter");
+
+    LoggerProvider::builder()
+        .with_simple_exporter(exporter)
+        .build()
+}
+
+fn main() {
+    let logger_provider = init_logger();
+    let layer = layer::OpenTelemetryTracingBridge::new(&logger_provider);
+    tracing_subscriber::registry().with(layer).init();
+
+    info!(event_id = 1234, user_id = 5678, "my test message");
+}
--- a/opentelemetry-journald-logs/src/lib.rs
+++ b/opentelemetry-journald-logs/src/lib.rs
@@ -1,0 +1,288 @@
+use async_trait::async_trait;
+use libc::{c_int, c_void};
+use opentelemetry::logs::AnyValue;
+use std::ffi::CString;
+use std::fmt::Debug;
+
+use opentelemetry::logs::LogError;
+use opentelemetry_sdk::export::logs::{LogData, LogExporter};
+
+use opentelemetry::logs::Severity;
+
+#[cfg(feature = "json")]
+use serde::Serialize;
+
+extern "C" {
+    fn sd_journal_sendv(iov: *const libc::iovec, n: libc::c_int) -> libc::c_int;
+}
+
+#[derive(Default)]
+pub struct JournaldLogExporterBuilder {
+    identifier: Option<String>,
+    message_size_limit: Option<usize>,
+    attribute_prefix: Option<String>,
+    json_format: bool,
+}
+
+impl JournaldLogExporterBuilder {
+    pub fn identifier(mut self, identifier: &str) -> Self {
+        self.identifier = Some(identifier.to_string());
+        self
+    }
+
+    pub fn message_size_limit(mut self, message_size_limit: usize) -> Self {
+        self.message_size_limit = Some(message_size_limit);
+        self
+    }
+
+    pub fn attribute_prefix(mut self, attribute_prefix: Option<String>) -> Self {
+        if let Some(prefix) = attribute_prefix {
+            self.attribute_prefix = Some(prefix.to_uppercase());
+        } else {
+            self.attribute_prefix = None;
+        }
+        self
+    }
+
+    pub fn json_format(mut self, json_format: bool) -> Self {
+        self.json_format = json_format;
+        self
+    }
+
+    pub fn build(self) -> Result<JournaldLogExporter, &'static str> {
+        let identifier = self.identifier.ok_or("Identifier is required")?;
+        let message_size_limit = self
+            .message_size_limit
+            .ok_or("Message size limit is required")?;
+        Ok(JournaldLogExporter {
+            identifier: CString::new(identifier).map_err(|_| "Invalid identifier")?,
+            message_size_limit,
+            attribute_prefix: self.attribute_prefix,
+            json_format: self.json_format,
+        })
+    }
+}
+
+#[derive(Debug)]
+pub struct JournaldLogExporter {
+    identifier: CString,
+    message_size_limit: usize,
+    attribute_prefix: Option<String>,
+    json_format: bool,
+}
+
+impl JournaldLogExporter {
+    pub fn builder() -> JournaldLogExporterBuilder {
+        JournaldLogExporterBuilder::default()
+    }
+
+    fn send_to_journald(&self, iovecs: &[libc::iovec]) -> Result<(), std::io::Error> {
+        let ret = unsafe { sd_journal_sendv(iovecs.as_ptr(), iovecs.len() as c_int) };
+        if ret < 0 {
+            Err(std::io::Error::last_os_error())
+        } else {
+            Ok(())
+        }
+    }
+
+    fn send_log_to_journald(&self, log_data: &LogData) -> Result<(), std::io::Error> {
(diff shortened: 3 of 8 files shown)


lalitb:
@TommyCpp had an question on comparing it with the subscriber provided by Tokio tracing library - https://github.com/tokio-rs/tracing/blob/master/tracing-journald/ - 

- The key distinction in the implementation is that the tracing journald subscriber writes directly to the Unix-domain socket monitored by the `journald` daemon. In contrast, this PR leverages the API provided by the `libsystemd` library (formerly `libsystemd-journald`) to perform the same operation. @Ralith (apologies for the tag) added the initial tracing implementation, so he might offer some background on the direct socket approach and any concerns related to using libsystemd?
- The tracing subscriber supports writing large logs (beyond limit enforced by journald config) using `memfd`, which can be added in subsequent implementation.
- I haven't done the benchmarks yet to compare, need to do this.

Ralith:
> background on the direct socket approach and any concerns related to using libsystemd

This was motivated by the improved downstream developer experience from being pure Rust, and the simplicity of the journald client protocol. It's not significantly more complicated than FFIing to libsystemd, and it dramatically reduces the sensitivity of downstream crates to build environment issues like missing pkg-config files, linking during cross-compilation, etc.

lalitb:
> This was motivated by the improved downstream developer experience from being pure Rust, and the simplicity of the journald client protocol. It's not significantly more complicated than FFIing to libsystemd, and it dramatically reduces the sensitivity of downstream crates to build environment issues like missing pkg-config files, linking during cross-compilation, etc.

Thanks, @Ralith, for the background. The concerns regarding build complexities with FFI to C are valid, and we will consider documenting this for the developers. My motivation for using libsystemd is that it internally handles sending large data through memfd, and so abstracting these low-level details. And hopefully being future-proof with any low-level protocol changes :)

opentelemetry-journald-logs/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+        self.identifier = Some(identifier.to_string());
+        self
+    }
+
+    pub fn message_size_limit(mut self, message_size_limit: usize) -> Self {
+        self.message_size_limit = Some(message_size_limit);
+        self
+    }
+
+    pub fn attribute_prefix(mut self, attribute_prefix: Option<String>) -> Self {
```
utpilla, line 38: It would be more user friendly to just ask them for a string slice instead of having them to call `to_string()` and wrapping it with an `Option`.

```suggestion
    pub fn attribute_prefix(mut self, attribute_prefix: &str) -> Self {
```

opentelemetry-journald-logs/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+}
+
+impl JournaldLogExporterBuilder {
+    pub fn identifier(mut self, identifier: &str) -> Self {
+        self.identifier = Some(identifier.to_string());
+        self
+    }
+
+    pub fn message_size_limit(mut self, message_size_limit: usize) -> Self {
+        self.message_size_limit = Some(message_size_limit);
```
utpilla, line 34: Add some validation? Something like `message_size_limit > 0`?

opentelemetry-journald-logs/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+            #[cfg(feature = "json")]
+            {
+                // Serialize message and attributes as JSON
+                let log_entry = LogEntry::from_log_data(log_data, self.attribute_prefix.clone());
+                let message_str = format!("MESSAGE={}", serde_json::to_string(&log_entry).unwrap());
+                message = Some(CString::new(message_str).unwrap());
+            }
+            #[cfg(not(feature = "json"))]
+            {
+                return Err(std::io::Error::new(
```
utpilla, line 113: Move this validation to `json_format` method where the user configures this bool variable?

opentelemetry-journald-logs/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    pub fn json_format(mut self, json_format: bool) -> Self {
+        self.json_format = json_format;
+        self
+    }
+
+    pub fn build(self) -> Result<JournaldLogExporter, &'static str> {
+        let identifier = self.identifier.ok_or("Identifier is required")?;
+        let message_size_limit = self
+            .message_size_limit
+            .ok_or("Message size limit is required")?;
```
utpilla, line 56: Could we have a default value for `message_size_limit` instead of requiring the user to always specify this?

opentelemetry-journald-logs/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+        }
+        self
+    }
+
+    pub fn json_format(mut self, json_format: bool) -> Self {
+        self.json_format = json_format;
+        self
+    }
+
+    pub fn build(self) -> Result<JournaldLogExporter, &'static str> {
```
utpilla, line 52: Could we simply return `JournadLogExporter` instead of a `Result`?

```suggestion
    pub fn build(self) -> JournaldLogExporter {
```

opentelemetry-journald-logs/examples/basic-logs.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use opentelemetry_appender_tracing::layer;
+use opentelemetry_journald_logs::JournaldLogExporter;
+use opentelemetry_sdk::logs::LoggerProvider;
+use tracing::info;
+use tracing_subscriber::prelude::*;
+
+fn init_logger() -> LoggerProvider {
+    let exporter = JournaldLogExporter::builder()
+        .identifier("opentelemetry-journal-exporter")
```
utpilla, line 11: Assign `with` prefix to the builder methods?

opentelemetry-journald-logs/examples/basic-logs.rs
```diff
@@ -0,0 +1,29 @@
...
+        .message_size_limit(4 * 1024)
+        .attribute_prefix(Some("OTEL_".to_string()))
+        .json_format(true) //uncomment to log in json format
```
utpilla, line 14: The comment is unclear.

opentelemetry-journald-logs/src/lib.rs
lalitb: Done

opentelemetry-journald-logs/src/lib.rs
lalitb: Done

opentelemetry-journald-logs/src/lib.rs
lalitb: Done

opentelemetry-journald-logs/src/lib.rs
lalitb: Yes, default is 0 now, which means no limit is enforced.

opentelemetry-journald-logs/src/lib.rs
lalitb: Done

opentelemetry-journald-logs/examples/basic-logs.rs
lalitb: Done

opentelemetry-journald-logs/examples/basic-logs.rs
lalitb: Good catch, fixed now.

cijothomas:
@lalitb Are you planning to resurrect this? I think it makes sense to have a working prototype and merge it in, and mark the status as alpha.
Improvements can be done iteratively, so as to not block progress.

lalitb:
> @lalitb Are you planning to resurrect this? I think it makes sense to have a working prototype and merge it in, and mark the status as alpha. Improvements can be done iteratively, so as to not block progress.

Yes, let me fix the CI, and make it ready. This would also help to get community contributions.

opentelemetry-journald-logs/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "opentelemetry-journald-logs"
+version = "0.1.0"
+edition = "2021"
+keywords = ["opentelemetry", "journald", "logs", "tracing"]
+description = "OpenTelemetry logs exporter for journald"
```
utpilla, line 6:
```suggestion
description = "OpenTelemetry Logs Exporter for journald"
```

opentelemetry-journald-logs/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+            let entries: Vec<String> = map
+                .iter()
+                .map(|(k, v)| format!("{}: {}", k, format_any_value(v)))
+                .collect();
+            format!("{{{}}}", entries.join(", "))
+        }
+    }
+}
+
+fn get_priority(severity: &Severity) -> i32 {
```
utpilla, line 250:
```suggestion
const fn get_priority(severity: &Severity) -> i32 {
```

opentelemetry-journald-logs/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+}
+
+fn sanitize_field_name(name: &str) -> String {
+    name.chars()
+        .map(|c| if c == '.' { '_' } else { c })
+        .skip_while(|&c| c == '_')
+        .filter(|&c| c == '_' || c.is_ascii_alphanumeric())
+        .collect::<String>()
+        .to_uppercase()
```
utpilla, line 292: I think we could avoid an allocation related to uppercase conversion by mapping the characters to upper case before collecting them into a `String`.

```rust
name.chars()
        .map(|c| if c == '.' { '_' } else { c })
        .skip_while(|&c| c == '_')
        .filter(|&c| c == '_' || c.is_ascii_alphanumeric())
        .map(|c| c.to_ascii_uppercase())
        .collect::<String>()
```

opentelemetry-journald-logs/examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+To export all log properties and attributes as JSON instead of key-value pairs, uncomment the `.json_format(true)` line in the init_logger function and use the following command:
+
+```rust
+fn init_logger() -> LoggerProvider {
+    let exporter = JournaldLogExporter::builder()
+        .identifier("opentelemetry-journal-exporter")
+        .message_size_limit(4 * 1024)
+        .attribute_prefix(Some("OTEL_".to_string()))
+        .json_format(true) // uncomment to log in JSON format
```
utpilla, line 23: Same as https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/84#discussion_r1661643763

opentelemetry-journald-logs/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+        });
+
+        let total_size: usize = iovecs.iter().map(|iov| iov.iov_len).sum();
+        let size_exceeded = total_size > self.message_size_limit;
+
+        // Try to send to journald regardless of the size
+        let send_result = self.send_to_journald(&iovecs);
+
+        if size_exceeded {
+            return Err(std::io::Error::new(
```
utpilla, line 177: Since the default message size limit is zero, does it mean that for a user who did not configure the message size limit explicitly, this method would always return an error?

opentelemetry-journald-logs/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "opentelemetry-journald-logs"
```
cijothomas, line 2:
```suggestion
name = "opentelemetry-journald"
```

utpilla: How about `opentelemetry-exporter-journald`?

```suggestion
name = "opentelemetry-exporter-journald"
```

opentelemetry-journald-logs/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+description = "OpenTelemetry logs exporter for journald"
+repository = "https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-journald-logs"
+homepage = "https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-journald-logs"
+license = "Apache-2.0"
+
+[dependencies]
+opentelemetry = { workspace = true, features = ["logs"] }
+opentelemetry_sdk = { workspace = true, features = ["logs"] }
+libc = "0.2"
+async-trait = "0.1"
```
utpilla, line 15: Is this needed?

opentelemetry-journald-logs/Cargo.toml
cijothomas: I like specific names, so opentelemetry-exporter-journald sounds good

PR #86, opened 2024-07-15
Update `opentelemetry-user_events-*` and `opentelemetry-etw-*` exporters to use otel v0.24

lalitb:
Fixes #
- Update `opentelemetry-user_events-*` and `opentelemetry-etw-*` exporters to use otel v0.24
- `opentelemetry-datadog` and `opentelemetry-stackdriver` continue using otel v0.23. Need help from the users/contributors for these repos to do the upgrade.
- `opentelemetry-zpages` and `opentelemetry-aws` don't need any significant change for the new versions, so they are implicitly bumped.

## Changes


opentelemetry-aws/Cargo.toml
```diff
@@ -31,5 +31,5 @@ opentelemetry_sdk = { workspace = true, optional = true }
 opentelemetry_sdk = { workspace = true, features = ["testing"] }
```
cijothomas, line 31: given the update to these crates, we need a changelog entry. or force them to use 0.23 itself

lalitb: Oh yes, there were few more which actually don't need any change. Updated their changelog.

opentelemetry-zpages/CHANGELOG.md
```diff
@@ -1,6 +1,16 @@
 # Changelog
 
-## vNext
+## Unreleased
+
+### Changed
```
cijothomas, line 5: 0.7.0 is repeated. Also both api and sdk versions are bumped?

cijothomas (approved):
LGTM.
Have left a qn about Zpages changes that should be fixed before merge.

opentelemetry-zpages/CHANGELOG.md
lalitb: Good catch. Fixed both.

PR #97, opened 2024-07-30
Upgrade hyper and otel dependencies

kcking-openai:
## Changes
Upgrade otel and hyper dependencies, change HttpClient example to hyper over unix socket.

I removed writing the resource for every span, as the field was [removed from `SpanData`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/CHANGELOG.md?plain=1#L58).

### DIFF
--- a/opentelemetry-datadog/CHANGELOG.md
+++ b/opentelemetry-datadog/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## vNext
 
+## v0.12.0
+
+### Changed
+
+- Bump opentelemetry and opentelemetry_sdk version to 0.24
+- Bump hyper to version 1
+
 ## v0.10.0
 
 ### Added
@@ -16,7 +23,6 @@
 
 - allow send all traces to `datadog-agent` with `agent-sampling` feature.
 - allow `datadog-agent` generate metrics from spans for [APM](https://docs.datadoghq.com/tracing/metrics/).
-
 
 ## v0.9.0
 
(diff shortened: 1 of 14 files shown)


lalitb:
> I removed writing the resource for every span, as the field was [removed from SpanData](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/CHANGELOG.md?plain=1#L58).

@kcking-openai thanks for the PR. Regarding resource, while it was removed from SpanData, [SpanExporter::set_resource()](https://github.com/open-telemetry/opentelemetry-rust/blob/7a9076032d270ae283c0549c7ef2f03833bbd2e5/opentelemetry-sdk/src/export/trace.rs#L68) method was introduced for one time setting of resource in Exporter. Default implementation was added for it in the SpanExporter trait so that custom exporters don't break. You may want to implement the `DatadogExporter ::set_resource()` method instead of removing resource from the code.

kcking-openai:
Thank you @lalitb, I have completed the CLA and pushed a fix for the lint. If you would like an alternative solution to the `too_many_arguments` lint I could wrap them in a struct, but I wasn't sure it was worth the effort at this point.

kcking-openai:
And I just pushed a fix for some lints that had broken in a couple unrelated directories.

lalitb (approved):
Thanks @kcking-openai. This looks good. Will publish a release shortly.

lalitb:
https://crates.io/crates/opentelemetry-datadog/0.12.0

Issue #104, opened 2024-09-07
Etw-Metrics - Improvements required


cijothomas:
Tracking item for improvements to the Etw-Metrics exporter:
https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-etw-metrics

1. Avoid heap allocation. It currently [uses a vec!](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-etw-metrics/src/exporter/mod.rs#L68) for each export. We could re-use the vec!, or use array (allocate maximum etw size - 65360).
2. It currently serialized the entire batch of metrics and sends them as one etw event. This makes it very easy to hit the 65360 limit on ETW. We need to send one metric at a time to ETW. This is how [OTel .NET exports metrics.](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs#L224). This should significantly reduce the probability of hitting ETW size limits.
3. Unit testing - it should be possible to write unit-tests where an ETW listener is setup and it can validate the metrics are flowing as expected.

cijothomas:
@mattbodd @psandana Either of you have bandwidth to help with the above, please reply to this issue. Thanks in advance!

mattbodd:
@cijothomas I'd be interested in taking this on!

cijothomas:
Thanks @mattbodd 2 is higher priority as users are hitting the limit very easily currently. 1 is more like nice-to-have, so lower priority than 2.

cijothomas:
@utpilla has corrected me that OTel .NET exports every metric point in one etw call, not one Metric. So the same can be followed here too, as that'd make it very safe (from hitting etw limits), even when future improvements like Exemplars come into picture!

mattbodd:
Thanks @cijothomas for calling out priorities and sharing how the .NET implementation approaches this issue

cijothomas:
> @utpilla has corrected me that OTel .NET exports every metric point in one etw call, not one Metric. So the same can be followed here too, as that'd make it very safe (from hitting etw limits), even when future improvements like Exemplars come into picture!

The above need to be revisited as too many calls to ETW might overwhelm the receiver. So we need more smarter splitting strategy. Following has some ideas worth exploring.
[...]

PR #105, opened 2024-09-09
etw-exporter-metrics updates to opentelemetry, better error logging

cijothomas:
Related to https://github.com/open-telemetry/opentelemetry-rust-contrib/issues/104, just adding better logging. Actual fixes are not part of this PR.
Also upgrades opentelemetry dependencies to latest by hand. Will update workspace in a future PR.


ThomsonTan (approved):

opentelemetry-etw-metrics/src/exporter/mod.rs
```diff
@@ -70,14 +70,22 @@ impl PushMetricsExporter for MetricsExporter {
...
             global::handle_error(MetricsError::Other(format!(
-                "Failed to write ETW event with error code: {}",
-                result
+                "Exporting failed due to event size {} exceeding the maximum size of {} bytes",
+                byte_array.len(),
+                etw::MAX_EVENT_SIZE
             )));
+        } else {
+            let result = etw::write(&byte_array);
+            // println!("Exported {} bytes to ETW", byte_array.len());
```
lalitb, line 81: Nit - remove the comment?

lalitb (approved):

opentelemetry-etw-metrics/src/exporter/mod.rs
cijothomas: I am really looking to get this https://github.com/open-telemetry/opentelemetry-rust/issues/1146 resolved, so we can liberally do internally logging. At this point, we are limited to logging in Error situations. (So this was my layman's attempt to figure out the serialized byte size, in various scenarios.!)

utpilla: I think it would be better to add a TODO statement here instead of having the `println!` comment in that case.

PR #102, opened 2024-08-28
Refresh generated code in opentelemetry-stackdriver

fiadliel:
`Send` and `Sync` in code generated by tonic 0.12.2 is fully qualified as `std::marker::Send` and `std::marker::Sync`.

## Changes

Refreshes generated code in opentelemetry-stackdriver to fix "generated_code_is_fresh" test.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Unit tests added/updated (if applicable)
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [x] Changes in public API reviewed (if applicable)

PR #108, opened 2024-09-10
Regenerate proto files in opentelemetry-stackdriver.

fiadliel:
## Changes

https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/102 updated proto files, but they've diverged again.
Probably these should remain `#[ignore]`d. Maybe run a separate (scheduled?) build check for them?


lalitb:
> Probably these should remain #[ignore]d. Maybe run a separate (scheduled?) build check for them?

I think [ignore]d would hide any required proto changes. But if this happens too often, we may have to revisit again.

lalitb (approved):

PR #111, opened 2024-09-18
Emit metric points as unique ETW events

### LINKED ISSUE #104: Etw-Metrics - Improvements required

mattbodd:
Part of #104

## Changes
* Constructs `opentelemetry_sdk::metrics::data::ResourceMetrics` for each `opentelemetry_sdk::metrics::data::Metric` data point and emits these as ETW events


opentelemetry-etw-metrics/src/exporter/mod.rs
```diff
@@ -63,32 +66,278 @@ impl Debug for MetricsExporter {
...
+                        resource_metrics.push(resource_metric);
+                    }
+                } else {
+                    global::handle_error(MetricsError::Other(format!(
+                        "Unsupported aggregation type: {:?}",
+                        data
+                    )));
+                }
+
+                for resource_metric in resource_metrics {
```
cijothomas, line 311: the approach used in this PR is extremely inefficient, as it clones the entire metrics and recreates them into a format that the opentelemetry-proto is able to serialize. This is okay for alpha, and will help resolve the size limit issue, but this must be done in a efficient way before stabilizing the component. (that might require support from upstream sdk)

mattbodd: Agreed. I'm working on a more performant solution but am proposing this PR as an immediate fix for #104

cijothomas: This is okay for an immediate fix.

cijothomas (approved):
LGTM as a mitigation for the ETW size limit issue. Definitely need to redesign this to avoid the clone/copy works.
Please add an entry to changelog.md file and we can merge.

PR #112, opened 2024-09-30
Add a derive crate to allow deriving conversions into common types

Sufflope:
Moved from https://github.com/open-telemetry/opentelemetry-rust/pull/1686

## Changes

Implement derive macros to convert user types into common Key/Value related types.

See https://github.com/open-telemetry/opentelemetry-rust/issues/1685


Sufflope:
I changed the intra-doc links to common types to direct links to docs.rs to avoid adding the `opentelemetry` crate to dependencies, which would then be added to users' projects dependency tree in the workspace version, possibly another version than the one they use.

The two uncovered lines are weird false negatives, is it blocking?

Sufflope:
Hello, I updated this MR to latest `main`. Arguably the package could be renamed `opentelemetry-macros` in case someone needs non-derive proc macros sometime. Also I updated to edition 2024, but all the other packages are on 2021, IDK your policy about this. And I needed to use this crate so I already published a 0.1.0 so it should probably be bumped to 0.2.0 or even 1.0.0 if it is actually officially merged.

opentelemetry-derive/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+//! ```rust
+//! use opentelemetry_derive::Value;
+//!
+//! #[derive(Value)]
+//! #[otel(variant = i64)]
+//! struct Counter {
+//!     count: i64,
+//! }
+//!
+//! impl From<&Counter> for i64 {
```
cijothomas, line 43: I am unsure how much value this actually gives. It does save some lines of code, but it's not a lot. Having to learn another crate, its macros is probably not worth it.
 
I'd prefer to wait to get more signs about the need of this (https://github.com/open-telemetry/opentelemetry-rust/issues/1685#issuecomment-4557809762) before accepting this PR, even in the contrib repo.

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/112/changes#r3313223046
I do not think we should take this now without seeing more demand.

Most PRs in this repo was not getting any attention for years, sorry about that. Just getting back to old PRs that are not auto-closed by bot!

PR #130, opened 2024-11-18
Prepare crates for otel v0.27.0

lalitb:
Fixes #

## Changes


cijothomas (approved):

opentelemetry-etw-logs/src/logs/converters.rs
```diff
@@ -17,6 +17,7 @@ impl IntoJson for AnyValue {
             AnyValue::Bytes(_value) => todo!("No support for AnyValue::Bytes yet."),
             AnyValue::ListAny(value) => value.as_json_value(),
             AnyValue::Map(value) => value.as_json_value(),
+            &_ => Value::Null,
```
psandana, line 20: I rather we add a branch that explicitly calls `todo!`. This way, upgrades to opentelemetry that adds a new value will fail to compile, forcing us to add support for it. Returning `Null` make it easy to forget.
```suggestion
            &_ => {
                // If we reach this line is because we are upgrading opentelemetry to a version that contains a new type.
                // We must add support for the new type.
                todo!("Add support for new types");
            }
```

lalitb: @psandana  The goal is to ensure that adding a new enum variant in OpenTelemetry does not break custom exporters.

PS: I believe the suggested change won't cause a compile-time failure but will result in a runtime panic, which we want to avoid entirely.

opentelemetry-etw-logs/src/logs/exporter.rs
```diff
@@ -401,6 +401,7 @@ fn add_attribute_to_event(event: &mut tld::EventBuilder, key: &Key, value: &AnyV
                 0,
             );
         }
+        &_ => {}
```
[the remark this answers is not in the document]
lalitb: @psandana  same as the earlier comment.

opentelemetry-etw-logs/src/logs/converters.rs
psandana: @lalitb you are right. IMO, if compile error is preferred, then I would just remove the `&_` entry. If opentelemetry ever adds a new item to the enum, it should come with code that handles it.

psandana: My bad. Forgot this is non-exhaustive, so you must use a default branch.

PR #134, opened 2024-12-03
Improve efficiency of ETW metrics exporter

### LINKED ISSUE #104: Etw-Metrics - Improvements required

mattbodd:
Fixes #104 
Design Decisions:
* Follow .Net and C++ implementations and use helper methods to serialize fragments of metric points into a coherent, exportable buffer
* Construct a new `ResourceMetrics` for each metric point and serialize each into an exportable buffer

## Changes

### DIFF
--- a/opentelemetry-etw-metrics/Cargo.toml
+++ b/opentelemetry-etw-metrics/Cargo.toml
@@ -17,7 +17,8 @@ opentelemetry-proto = { workspace = true, features = ["gen-tonic", "metrics"] }
 async-trait = "0.1"
 prost = "0.13"
 tracelogging = "1.2.1"
-tracing = {version = "0.1", optional = true}
+tracing = { version = "0.1", optional = true }
+criterion = { workspace = true, features = ["html_reports"] }
 
 [dev-dependencies]
 tokio = { version = "1.0", features = ["full"] }
@@ -28,3 +29,11 @@ default = ["internal-logs"]
 
 [package.metadata.cargo-machete]
 ignored = ["tracing"]
+
+[[bench]]
+name = "etw"
+harness = false
+
+[[bench]]
+name = "exporter"
+harness = false
--- a/opentelemetry-etw-metrics/benches/exporter.rs
+++ b/opentelemetry-etw-metrics/benches/exporter.rs
@@ -13,17 +13,10 @@ RAM: 64.0 GB
 | exporter                       | 847.38µs    |
 */
 
-use opentelemetry::metrics::Gauge;
 use opentelemetry::{metrics::MeterProvider as _, KeyValue};
-use opentelemetry::{InstrumentationScope, InstrumentationScopeBuilder};
 use opentelemetry_etw_metrics::MetricsExporter;
-use opentelemetry_sdk::metrics::reader::MetricReader;
-use opentelemetry_sdk::metrics::{ManualReader, PeriodicReader, SdkMeterProvider};
 use opentelemetry_sdk::{
-    metrics::{
-        data::{ResourceMetrics, ScopeMetrics},
-        exporter::PushMetricExporter,
-    },
+    metrics::{reader::MetricReader, PeriodicReader, SdkMeterProvider},
     runtime, Resource,
 };
 
--- a/opentelemetry-etw-metrics/benches/exporter.rs
+++ b/opentelemetry-etw-metrics/benches/exporter.rs
@@ -5,12 +5,12 @@
 /*
 The benchmark results:
 criterion = "0.5.1"
-OS:
-Hardware:
-RAM:
+OS: Windows 11 Enterprise N, 23H2, Build 22631.4460
+Hardware: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz   2.79 GHz, 16vCPUs
+RAM: 64.0 GB
 | Test                           | Average time|
 |--------------------------------|-------------|
-|                                |             |
+| exporter                       | 1.2927ms    |
 */
 
 use opentelemetry_etw_metrics::MetricsExporter;
--- a/opentelemetry-etw-metrics/benches/exporter.rs
+++ b/opentelemetry-etw-metrics/benches/exporter.rs
@@ -13,18 +13,26 @@ RAM:
 |                                |             |
 */
 
+use opentelemetry_etw_metrics::MetricsExporter;
+use opentelemetry_sdk::{metrics::{data::{ResourceMetrics, ScopeMetrics}, exporter::PushMetricExporter}, Resource};
+
 use criterion::{criterion_group, criterion_main, Criterion};
 
-fn fibonacci(n: u64) -> u64 {
-    match n {
-        0 => 0,
-        1 => 1,
-        _ => fibonacci(n - 1) + fibonacci(n - 2),
-    }
+fn export() {
+    let exporter = MetricsExporter::new();
+    let mut resource_metrics = ResourceMetrics {
+        resource: Resource::default(),
+        scope_metrics: vec![ScopeMetrics::default(), ScopeMetrics::default()],
+    };
+
+    let runtime = tokio::runtime::Runtime::new().unwrap();
+    runtime.block_on(async {
+        exporter.export(&mut resource_metrics).await.unwrap();
+    });
 }
 
 fn criterion_benchmark(c: &mut Criterion) {
-    c.bench_function("fib 20", |b| b.iter(|| fibonacci(20)));
+    c.bench_function("export", |b| b.iter(|| { export()}));
 }
 
 criterion_group!(benches, criterion_benchmark);
--- a/opentelemetry-etw-metrics/benches/exporter.rs
+++ b/opentelemetry-etw-metrics/benches/exporter.rs
@@ -1,0 +1,31 @@
+//! run with `$ cargo bench --bench exporter -- --exact <test_name>` to run specific test for logs
+//! So to run test named "fibonacci" you would run `$ cargo bench --bench exporter -- --exact fibonacci`
+//! To run all tests for logs you would run `$ cargo bench --bench exporter`
+//!
+/*
+The benchmark results:
+criterion = "0.5.1"
+OS:
+Hardware:
+RAM:
+| Test                           | Average time|
+|--------------------------------|-------------|
+|                                |             |
+*/
+
+use criterion::{criterion_group, criterion_main, Criterion};
+
+fn fibonacci(n: u64) -> u64 {
+    match n {
+        0 => 0,
+        1 => 1,
+        _ => fibonacci(n - 1) + fibonacci(n - 2),
+    }
+}
+
+fn criterion_benchmark(c: &mut Criterion) {
+    c.bench_function("fib 20", |b| b.iter(|| fibonacci(20)));
+}
+
+criterion_group!(benches, criterion_benchmark);
+criterion_main!(benches);
--- a/opentelemetry-etw-metrics/benches/exporter.rs
+++ b/opentelemetry-etw-metrics/benches/exporter.rs
@@ -10,29 +10,55 @@ Hardware: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz   2.79 GHz, 16vCPUs
 RAM: 64.0 GB
 | Test                           | Average time|
 |--------------------------------|-------------|
-| exporter                       | 1.2927ms    |
+| exporter                       | 847.38µs    |
 */
 
+use opentelemetry::metrics::Gauge;
+use opentelemetry::{metrics::MeterProvider as _, KeyValue};
+use opentelemetry::{InstrumentationScope, InstrumentationScopeBuilder};
 use opentelemetry_etw_metrics::MetricsExporter;
-use opentelemetry_sdk::{metrics::{data::{ResourceMetrics, ScopeMetrics}, exporter::PushMetricExporter}, Resource};
+use opentelemetry_sdk::metrics::reader::MetricReader;
+use opentelemetry_sdk::metrics::{ManualReader, PeriodicReader, SdkMeterProvider};
+use opentelemetry_sdk::{
+    metrics::{
+        data::{ResourceMetrics, ScopeMetrics},
+        exporter::PushMetricExporter,
+    },
+    runtime, Resource,
+};
 
 use criterion::{criterion_group, criterion_main, Criterion};
 
 fn export() {
-    let exporter = MetricsExporter::new();
-    let mut resource_metrics = ResourceMetrics {
-        resource: Resource::default(),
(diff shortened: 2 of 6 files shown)


opentelemetry-etw-metrics/benches/exporter.rs
```diff
@@ -0,0 +1,58 @@
...
+use opentelemetry::{metrics::MeterProvider as _, KeyValue};
+use opentelemetry_etw_metrics::MetricsExporter;
+use opentelemetry_sdk::{
+    metrics::{reader::MetricReader, PeriodicReader, SdkMeterProvider},
+    runtime, Resource,
+};
+
+use criterion::{criterion_group, criterion_main, Criterion};
+
+fn export() {
```
cijothomas, line 25: this is not measuring the cost of export - it is creating new runtime, create meter provider, meter, instrument, record 10,000 measurements, flush, and drop...

mattbodd:
Comparing the latest `exporter` benchmark results between the proposed `export` implementation and the previous `export` implementation, there is a ~26% improvement.

opentelemetry-etw-metrics/benches/exporter.rs
mattbodd: Thanks for the comment. I tried to address this in my latest change.

opentelemetry-etw-metrics/benches/exporter.rs
```diff
@@ -0,0 +1,80 @@
...
+fn criterion_benchmark(c: &mut Criterion) {
+    let runtime = tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(1)
+        .enable_all()
+        .build()
+        .unwrap();
+
+    c.bench_function("export", |b| {
+        b.to_async(&runtime)
+            .iter(|| export(create_resource_metrics()))
```
cijothomas, line 75: no need to create resource_metrics in each iteration. create once and repeatedly export it.
same with exporter - create once only.

opentelemetry-etw-metrics/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-mod etw;
+pub mod etw;
```
cijothomas, line 1: we won't want to make this public. lets revert if this was done for some testing.

opentelemetry-etw-metrics/benches/exporter.rs
```diff
@@ -0,0 +1,80 @@
...
+        data_points: vec![data_point.clone(), data_point.clone(), data_point],
+        temporality: Temporality::Delta,
+        is_monotonic: true,
+    };
+
+    let resource_metrics = ResourceMetrics {
+        resource: Resource::new(vec![KeyValue::new("service.name", "my-service")]),
+        scope_metrics: vec![ScopeMetrics {
+            scope: InstrumentationScope::default(),
+            metrics: vec![Metric {
```
cijothomas, line 54: I'd suggest to use 10 metrics, each with 2000 datapoints inside it, to get a realistic measure.

opentelemetry-etw-metrics/src/lib.rs
mattbodd: I made it public so that I could access `etw::write` in a benchmark. Is there a better way to achieve that?

cijothomas: no. and I don't that is something we need to test either. cost of writing to etw is indirectly covered by the other benchmark.

opentelemetry-etw-metrics/benches/exporter.rs
mattbodd: Added although I'm relying on a helper function to create metrics since `opentelemetry_sdk::metrics::Metric` doesn't implement `clone`

opentelemetry-etw-metrics/benches/exporter.rs
mattbodd: Exporter was trivial to reuse but I'm struggling to reuse the `ResourceMetric` because Rust complains about the captured variable (resource_metrics) escaping the FnMut closure body

opentelemetry-etw-metrics/src/lib.rs
mattbodd: Great, remove the `etw` benchmark and restored the `etw` module's privacy

mattbodd:
> Comparing the latest `exporter` benchmark results between the proposed `export` implementation and the previous `export` implementation, there is a ~26% improvement.

With a more realistic `ResourceMetrics`, the improvement looks to be closer to 38%

opentelemetry-etw-metrics/Cargo.toml
```diff
@@ -17,7 +17,8 @@ opentelemetry-proto = { workspace = true, features = ["gen-tonic", "metrics"] }
 async-trait = "0.1"
 prost = "0.13"
 tracelogging = "1.2.1"
-tracing = {version = "0.1", optional = true}
+tracing = { version = "0.1", optional = true }
+criterion = { workspace = true, features = ["html_reports"] }
```
utpilla, line 21: This should be a dev dependency.

mattbodd: Thanks, updated

psandana:
We noticed some `tonic` namespace usage along with types with `Tonic` prefix. Discussing with @cijothomas we got clarified OpenTelemetry is not using Tonic, they are just types coming from proto definitions.

Thus, this comment is not blocking as the change is unrelated to this PR, but we propose making those proto types more generic, thus we prevent future confusions regarding tonic usage or not.

opentelemetry-etw-metrics/src/exporter/mod.rs
```diff
@@ -39,265 +41,135 @@ impl Debug for MetricsExporter {
     }
 }
 
+fn emit_export_metric_service_request(
+    export_metric_service_request: &ExportMetricsServiceRequest,
+) -> MetricResult<()> {
+    let mut encoding_buffer = Vec::new();
```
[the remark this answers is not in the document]
mattbodd: Addressed in [128490a](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/134/commits/128490a8ce78b8ceb1d2c0e65c7d706268417820)

mattbodd:
> We noticed some `tonic` namespace usage along with types with `Tonic` prefix. Discussing with @cijothomas we got clarified OpenTelemetry is not using Tonic, they are just types coming from proto definitions.
> 
> Thus, this comment is not blocking as the change is unrelated to this PR, but we propose making those proto types more generic, thus we prevent future confusions regarding tonic usage or not.

I borrowed the `use` pattern set in [open-telemetry/opentelemetry-rust/opentelemetry-proto/src/transform/metrics.rs](https://github.com/open-telemetry/opentelemetry-rust/blob/6cc327d555ce1d2ef90d076a53393a66c29362ac/opentelemetry-proto/src/transform/metrics.rs#L20). @psandana or @cijothomas do you have suggestions for better generic names?

opentelemetry-etw-metrics/src/exporter/mod.rs
```diff
@@ -39,265 +41,135 @@ impl Debug for MetricsExporter {
     }
 }
 
+fn emit_export_metric_service_request(
+    export_metric_service_request: &ExportMetricsServiceRequest,
+) -> MetricResult<()> {
+    let mut encoding_buffer = Vec::with_capacity(export_metric_service_request.encoded_len());
```
utpilla, line 47: Not a blocker for this PR but we should be able to reuse the `vec` instead of allocating a new one for each export. `MetricExporter` struct can hold a `vec` as one of its fields and keeping using that `vec` for each export.

opentelemetry-etw-metrics/src/exporter/mod.rs
```diff
@@ -39,265 +41,135 @@ impl Debug for MetricsExporter {
...
+
+    export_metric_service_request
+        .encode(&mut encoding_buffer)
+        .map_err(|err| MetricError::Other(err.to_string()))?;
+
+    if (encoding_buffer.len()) > etw::MAX_EVENT_SIZE {
```
utpilla, line 53: It would be better to check `if export_metric_service_request.encoded_len() > etw::MAX_EVENT_SIZE` beforehand instead of always trying to encode the data into the `vec` and then checking for the length limit.

opentelemetry-etw-metrics/src/exporter/mod.rs
mattbodd: I make a tweak that's part of the way towards this improvement. Now `emit_export_metric_service_request` takes a `&mut Vec<u8>` as a parameter to avoid allocating a fresh `Vec<u8>` every invocation. Now a new vector is only initialized once per `export`.

opentelemetry-etw-metrics/src/exporter/mod.rs
mattbodd: Good catch. Updated in https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/134/commits/1fe84ff7a94f36539f97407e8f2a3841419bdc83

opentelemetry-etw-metrics/src/exporter/mod.rs
```diff
@@ -39,265 +41,146 @@ impl Debug for MetricsExporter {
...
 #[async_trait]
 impl PushMetricExporter for MetricsExporter {
     async fn export(&self, metrics: &mut ResourceMetrics) -> MetricResult<()> {
+        let schema_url: String = metrics
+            .resource
+            .schema_url()
+            .map(Into::into)
+            .unwrap_or_default();
+
+        let mut encoding_buffer = Vec::new();
```
lalitb, line 82: This can be optimized by using a pre-allocated thread-local vector of MAX_EVENT_SIZE size. Probably for future optimization.

opentelemetry-etw-metrics/benches/exporter.rs
```diff
@@ -0,0 +1,105 @@
...
+        b.iter_custom(|iters| {
+            let exporter = MetricsExporter::new();
+
+            let mut resource_metrics = create_resource_metrics();
+
+            let start = std::time::Instant::now();
+
+            for _i in 0..iters {
+                runtime.block_on(async {
+                    export(&exporter, &mut resource_metrics).await;
```
utpilla, line 96: It might be better to micro benchmark just the serialization and export of one data point instead of 20000. It would keep the calculations simple when comparing two benchmark results.

opentelemetry-etw-metrics/src/exporter/mod.rs
utpilla: Related to this https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/134#discussion_r1914018851.

I think it might be better to use a `Mutex` protected pre-allocated `vec` instead of a thread local `vec`. If the users are setting up this exporter with a runtime such as Tokio, then having a thread local `vec` would mean that potentially all worker threads would end up allocating a `vec` of size MAX_EVENT_SIZE. If we use a `Mutex` protect thread instead, we will only allocate this `vec` once and there would be no contention for the `Mutex` as we never call export concurrently.

utpilla (approved):

opentelemetry-etw-metrics/src/exporter/mod.rs
lalitb: Oh missed that comment. Btw, export can be called concurrently as this is the re-entrant processor right?

utpilla: No with metrics we always "batch" (aggregate the data) and export it. If an export is already in progress, the subsequent export would wait for the previous export to complete.

lalitb: Oh yes, was mixing logs and metrics :)

opentelemetry-etw-metrics/benches/exporter.rs
lalitb: I think this is coming from - https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/134#discussion_r1889202838

opentelemetry-etw-metrics/src/exporter/mod.rs
```diff
@@ -39,265 +41,146 @@ impl Debug for MetricsExporter {
...
+                            schema_url: schema_url.clone(),
+                        }],
+                        schema_url: schema_url.clone(),
+                    }],
+                };
+
+                if let Some(proto_data) = proto_data {
+                    match proto_data {
+                        TonicMetricData::Histogram(hist) => {
+                            for data_point in hist.data_points {
```
psandana, line 110: The contents of each arm looks repetitive. Can this be factored into a function? It would help also to improve readability and maintainability.

opentelemetry-etw-metrics/benches/exporter.rs
mattbodd: Yeah, let me know if there's a preference between a micro benchmark and what Cijo suggested.

cijothomas: not a blocker. Exporting Metrics is not in hot path, so it may not be worth perfecting this.

opentelemetry-etw-metrics/src/exporter/mod.rs
mattbodd: There is repetition that could probably be reduced and I'm open to suggestions. I don't see a clean way to abstract the repetitive logic into a function because specialization is needed to construct data points of different metric data types. We could probably add traits that allow for better abstraction, but I'm not sure I want to attempt that here.

opentelemetry-etw-metrics/src/exporter/mod.rs
cijothomas, line 82: @mattbodd Please add a TODO here to avoid this vec allocation, and instead use a mutex wrapped vec, pre-created in this struct.

lalitb (approved):

cijothomas (approved):

PR #136, opened 2024-12-10
Fix CI lint errors from new stable toolchain v1.83

garysassano:
See related issue: open-telemetry/opentelemetry-rust#2370

## Changes

Fixed the following lint errors:

```sh
error: use of deprecated method `opentelemetry_sdk::trace::Builder::with_config`: Config is becoming a private type. Use Builder::with_{config_name}(resource) instead. ex: Builder::with_resource(resource)
```

```sh
error: the following explicit lifetimes could be elided: 'a
```

```sh
error: empty line after doc comment
```

### DIFF
--- a/opentelemetry-contrib/benches/new_span.rs
+++ b/opentelemetry-contrib/benches/new_span.rs
@@ -13,7 +13,7 @@ use opentelemetry_contrib::trace::{
 };
 use opentelemetry_sdk::{
     export::trace::{ExportResult, SpanData, SpanExporter},
-    trace::{Config, Sampler, TracerProvider},
+    trace::{Sampler, TracerProvider},
 };
 #[cfg(not(target_os = "windows"))]
 use pprof::criterion::{Output, PProfProfiler};
--- a/opentelemetry-contrib/benches/new_span.rs
+++ b/opentelemetry-contrib/benches/new_span.rs
@@ -152,7 +152,7 @@ impl Display for Environment {
 
 fn parent_sampled_tracer(inner_sampler: Sampler) -> (TracerProvider, BoxedTracer) {
     let provider = TracerProvider::builder()
-        .with_config(Config::default().with_sampler(Sampler::ParentBased(Box::new(inner_sampler))))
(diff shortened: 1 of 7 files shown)


cijothomas (approved):

cijothomas:
Thank you ❤️

garysassano:
The deprecation warnings in the Datadog agent sampling example can't be fixed yet because the Datadog pipeline specifically expects a `trace::Config` in [its builder](https://github.com/apollographql/router/blob/f844b4e3c6ebef06507c1856a83c1ce046dd7c0c/apollo-router/src/plugins/telemetry/tracing/datadog_exporter/exporter/mod.rs#L155-L162).

While the OpenTelemetry SDK is moving away from `Config` in favor of using `TracerProvider::builder()` directly, we need to keep using the deprecated `Config` methods until the Datadog integration is updated to accept a `TracerProvider` instead.

For now, we're suppressing the deprecation warnings with `#[allow(deprecated)]` since this is something that needs to be fixed in the Datadog integration itself.

garysassano:
The deprecation warnings in the Datadog [agent sampling example](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/0c3771a41af47db7f79800890feb2c2aa8ad4f7f/opentelemetry-datadog/examples/agent_sampling.rs#L63-L65) can't be fixed yet because the Datadog pipeline specifically expects a `trace::Config` in [its builder](https://github.com/apollographql/router/blob/f844b4e3c6ebef06507c1856a83c1ce046dd7c0c/apollo-router/src/plugins/telemetry/tracing/datadog_exporter/exporter/mod.rs#L155-L162).

While the OpenTelemetry SDK is moving away from `Config` in favor of using `TracerProvider::builder()` directly, we need to keep using the deprecated `Config` methods until the Datadog integration is updated to accept a `TracerProvider` instead.

For now, we're suppressing the deprecation warnings with `#[allow(deprecated)]` since this is something that needs to be fixed in the Datadog integration itself.

PR #133, opened 2024-11-27
Add benchmark for user_events enabled check

lalitb:
## Changes

Initial benchmark to check if the tracepoint is enabled. Need to add more benchmark to write to the tracepoint. 

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #138, opened 2024-12-11
Add stress-test framework to contrib repo.

lalitb:
## Changes

I believe it would be benefitial to have stress framework copied from main to the contrib repo, to enable packages from this repo do necessary tests.

Have added the framework, along with the sample stress test for user-events enabled. The stress test result for user_events can be discussed separately in #133.

Ps. - It would be difficult to keep stress framework in sync between the two repos. I don't see any other easy way as of now, unless we publish the framework as a separate crate. But this would incur added responsibility in maintaining it.

### DIFF
--- a/stress/Cargo.toml
+++ b/stress/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 publish = false
 
-[[bin]] # Bin to run the metrics stress tests for Counter
+[[bin]] # Bin to run the metrics stress tests for Logs UserEvent Exporter
 name = "user-events-enabled"
 path = "src/user_events_enabled.rs"
 doc = false
--- a/stress/Cargo.toml
+++ b/stress/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "stress"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[[bin]] # Bin to run the metrics stress tests for Counter
+name = "user-events-enabled"
+path = "src/user_events_enabled.rs"
+doc = false
+
+[dependencies]
+ctrlc = "3.2.5"
+lazy_static = "1.4.0"
+num_cpus = "1.15.0"
+rand = { version = "0.8.4", features = ["small_rng"] }
+num-format = "0.4.4"
+sysinfo = { version = "0.32", optional = true }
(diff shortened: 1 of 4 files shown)


stress/src/user_events_enabled.rs
```diff
@@ -0,0 +1,38 @@
...
+// sudo -E ~/.cargo/bin/cargo rnu --bin user-events-enabled --release
+
+// TODO : Add stess result here.
+
+mod throughput;
+use eventheader_dynamic::{Provider, ProviderOptions};
+use lazy_static::lazy_static;
+
+// Global constants for level and keyword
+const LEVEL: u8 = 4; // Example level (Informational)
```
cijothomas, line 11: couple of suggestions
1. Lets add one test with and without tracepoint being enabled.
2. Instead of testing just the event header, lets test via OTel itself, so we know the overall cost/contention, not just that of EventHeader crate.

stress/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "stress"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[[bin]] # Bin to run the metrics stress tests for Counter
```
cijothomas, line 7:
```suggestion
[[bin]] # Bin to run the metrics stress tests for Logs UserEvent Exporter
```

stress/src/user_events_enabled.rs
lalitb: > Lets add one test with and without tracepoint being enabled.

Yes, I thought about it. This needs some thinking, as there needs to be some listener for the tracepoint. Can we do it as separate PR?

> Instead of testing just the event header, lets test via OTel itself, so we know the overall cost/contention, not just that of EventHeader crate.

This test was specifically for cost incurred for `eventheader` crate. Plan is to add separate test for the actual export and enabled check through Otel.

cijothomas: if this is only testing event header (that should be done by that repo!), lets rename to "eventheader_enabled" / similar.

cijothomas (approved):
It's okay to copy the stress test logic for now. (I think .NET GitHub also does this !)
Looking forward to seeing more OTel exporter code being stress tested.

(left some non blocking comments, okay to follow up)

PR #140, opened 2024-12-12
Update stress test results for eventheader

lalitb:
## Changes

update stress test for eventheader crate. The stats don't differ whether tracepoint is enabled or disabled- as it only involves a single bit read in the kernel VFS, so haven't added separate test for it.


lalitb:
Attaching the flamegraph with single thread, and concurrent threads ( equal to number of logical processors):

concurrent enabled check:

flamegraph_full

Single thread enabled check:
flamegraph

The relevant part in the stack trace in flamegraph is the frequent context switching in concurrent enabled graph, which is missing in single thread check, Will investigate more. 

However, the PR should be good to publish with these results. The investigation can follow further.

cijothomas (approved):
Thanks. I don't think we can invoke the check for each and every log, as it'll introduce contention to logging :( We can discuss some alternatives to avoid contention.

Also, we need to separately stress test if we see contention when actually exporting telemetry to user-events - it should be writing to per-cpu-buffer, so none is expected, but let us try and see if the assumptions hold true.

PR #150, opened 2025-01-15
Add back MSRV CI job and adjust MSRVs

gruebel:
## Changes

- Most of the MSRVs were not adjusted with the recent update to OTel 0.27 version (fixed that)
- Also added back the MSRV CI job and migrated to a very similar approach like `opentelemetry-rust`, just using `cargo-msrv` instead to verify it
- I had to set `1.71.1` in the CI job for all the packages, because `cargo-msrv` doesn't work with Rust <1.71 (also added a comment in the config file), but that doesn't matter much since the next OTel version will raise the MSRV to 1.75 anyway
- Also had to patch two dependencies otherwise it wouldn't work for that specific version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,25 +50,28 @@ jobs:
         args: --all -- --check
     - name: Lint
       run: ./scripts/lint.sh
-# TODO: re-add this test
-#  msrv:
-#    runs-on: ubuntu-latest
-#    steps:
-#    - uses: actions/checkout@v4
-#      with:
-#        submodules: true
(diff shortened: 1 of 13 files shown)


cijothomas (approved):
Thanks. We need changelog update for the msrv bump, but we can do it after bumping to 1.75 (from main repo).

gruebel:
> Thanks. We need changelog update for the msrv bump, but we can do it after bumping to 1.75 (from main repo).

yeah, sounds like a good plan. no need to do it now, because the ones I fixed had the wrong versions set anyway 😅

PR #152, opened 2025-01-15
Replace once_cell with OnceLock

gruebel:
## Changes

- Since the MSRV will become 1.70+ via #150 there is no need for `once_cell` anymore
- The implementation is identical to the OTel one
https://github.com/open-telemetry/opentelemetry-rust/blob/b53c19e2e6c4f8f18d2427c40556b2d0b574aa50/opentelemetry-sdk/src/propagation/trace_context.rs#L18-L23
-  I also renamed the header fields static to `TRACE_CONTEXT_HEADER_FIELDS` to keep identical through out the different packages (they are not public so no harm here)

Do you want me to add it to the changelog or not worth it? No user-facing changes 🙂  

### DIFF
--- a/opentelemetry-aws/Cargo.toml
+++ b/opentelemetry-aws/Cargo.toml
@@ -25,7 +25,6 @@ detector-aws-lambda = ["dep:opentelemetry-semantic-conventions"]
 internal-logs = ["tracing"]
 
 [dependencies]
-once_cell = "1.12"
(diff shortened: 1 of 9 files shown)


gruebel:
@cijothomas would be great, if you can also review this :slightly_smiling_face:

cijothomas (approved):
LGTM!

PR #156, opened 2025-01-23
Remove global set_tracer_provider from trace pipelines

gruebel:
## Changes

- Follow-up to https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/155/files/1f8b9f034b76adec7a69a30c3afc91b7f2d3e449#r1924545070
- Also adjusted all the examples I could find

### DIFF
--- a/opentelemetry-contrib/src/trace/exporter/jaeger_json.rs
+++ b/opentelemetry-contrib/src/trace/exporter/jaeger_json.rs
@@ -49,7 +49,7 @@ impl<R: JaegerJsonRuntime> JaegerJsonExporter<R> {
     }
 
     /// Install the exporter using the internal provided runtime
-    pub fn install_batch(self) -> Tracer {
+    pub fn install_batch(self) -> (Tracer, TracerProvider) {
         let runtime = self.runtime.clone();
         let provider_builder = TracerProvider::builder().with_batch_exporter(self, runtime);
         let provider = provider_builder.build();
@@ -58,9 +58,8 @@ impl<R: JaegerJsonRuntime> JaegerJsonExporter<R> {
             .with_schema_url(SCHEMA_URL)
             .build();
         let tracer = provider.tracer_with_scope(scope);
-        let _ = opentelemetry::global::set_tracer_provider(provider);
 
-        tracer
+        (tracer, provider)
     }
 }
 
--- a/opentelemetry-contrib/src/trace/exporter/jaeger_json.rs
+++ b/opentelemetry-contrib/src/trace/exporter/jaeger_json.rs
@@ -11,14 +11,11 @@ use opentelemetry::trace::SpanId;
     feature = "rt-tokio-current-thread"
 ))]
 use opentelemetry::trace::TraceError;
-use opentelemetry::trace::TracerProvider as _;
-use opentelemetry::InstrumentationScope;
 use opentelemetry_sdk::{
     export::trace::{ExportResult, SpanData, SpanExporter},
     runtime::RuntimeChannel,
-    trace::{Tracer, TracerProvider},
+    trace::TracerProvider,
 };
-use opentelemetry_semantic_conventions::SCHEMA_URL;
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::time::SystemTime;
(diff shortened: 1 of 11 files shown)


opentelemetry-contrib/src/trace/exporter/jaeger_json.rs
```diff
@@ -49,7 +49,7 @@ impl<R: JaegerJsonRuntime> JaegerJsonExporter<R> {
     }
 
     /// Install the exporter using the internal provided runtime
-    pub fn install_batch(self) -> Tracer {
+    pub fn install_batch(self) -> (Tracer, TracerProvider) {
```
cijothomas, line 52: this need to return the provider only. The caller then can create Tracer using the returned provider.

opentelemetry-datadog/src/exporter/mod.rs
```diff
@@ -295,7 +295,7 @@ impl DatadogPipelineBuilder {
             .with_attributes(None)
             .build();
         let tracer = provider.tracer_with_scope(scope);
-        let _ = global::set_tracer_provider(provider.clone());
+
```
cijothomas, line 298: return provider only, not tracer.

opentelemetry-contrib/src/trace/exporter/jaeger_json.rs
gruebel: Yeah, I thought about that, but my problem is the `InstrumentationScope` which is defined here and used to setup the tracer. Any advice on how to handle it? Or should I also remove it?

cijothomas: You can remove it. The instrumentation scope is expected to capture the component emitting the traces, so whoever is emitting traces, they should call provider.tracer("component-name"), and use that tracer, rather than creating a tracer with hardcoded name here. (OpenTelemetry-OTLP was also doing this mistake - it was returning a tracer with hardcode named "otel-otlp", which defeats the purpose of InstrumentationScope, as every telemetry would be associated incorrectly with "otel-otlp", instead of reflecting the component)

gruebel: hey @cijothomas I did as asked, please have an other look again.

opentelemetry-datadog/examples/agent_sampling.rs
```diff
@@ -66,6 +67,13 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
                 .with_id_generator(RandomIdGenerator::default()),
         )
         .install_simple()?;
+    global::set_tracer_provider(provider.clone());
+    let scope = InstrumentationScope::builder("opentelemetry-datadog")
```
cijothomas, line 71: opentelemetry-datadog is not the right name. Please include "example/demo" so people don't mistake it!

cijothomas, line 71:
```suggestion
    let scope = InstrumentationScope::builder("opentelemetry-datadog-demo")
```

opentelemetry-datadog/src/exporter/model/mod.rs
```diff
@@ -44,19 +44,21 @@ static DD_MEASURED_KEY: &str = "_dd.measured";
...
-///            .install_batch(opentelemetry_sdk::runtime::Tokio)?;
+///     let provider = new_pipeline()
+///         .with_service_name("my_app")
+///         .with_api_version(ApiVersion::Version05)
+///         // the custom mapping below will change the all spans' name to datadog spans
+///         .with_name_mapping(|span, model_config|{"datadog spans"})
+///         .with_agent_endpoint("http://localhost:8126")
+///         .install_batch(opentelemetry_sdk::runtime::Tokio)?;
+///     global::set_tracer_provider(provider.clone());
+///     let tracer = global::tracer("opentelemetry-datadog");
```
cijothomas, line 59:
```suggestion
///     let tracer = global::tracer("opentelemetry-datadog-demo");
```

cijothomas (approved):
left couple of nits, but LGTM overall.

lalitb (approved):

PR #165, opened 2025-02-08
fix and expose K8s resource detector

gruebel:
## Changes

- not sure why, but this was never exposed and it actually didn't work as intended. maybe that's why 😅 
- I was not sure, how to best mock the existence + content of a file under a specific path. if you have a good idea let me know, otherwise I leave the test as is


opentelemetry-resource-detectors/src/k8stest.rs
```diff
@@ -1,58 +0,0 @@
...
-                        KeyValue::new("k8s.node.name", "test-node"),
-                    ])
-                );
-            },
-        );
-    }
-
-    #[test]
-    fn test_k8s_resource_detector_with_missing_env_vars() {
-        // Temporarily set only one environment variable to test defaults
```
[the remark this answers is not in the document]
gruebel: sure, I added it back.

opentelemetry-resource-detectors/Cargo.toml
```diff
@@ -16,3 +16,6 @@ rust-version = "1.70.0"
 opentelemetry = { workspace = true }
 opentelemetry_sdk = { workspace = true }
 opentelemetry-semantic-conventions = { workspace = true }
+
+[dev-dependencies]
+temp-env = "0.3.6"
```
cijothomas, line 21: interesting...there were existing tests that used this somehow?

opentelemetry-resource-detectors/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## vNext
 
+### Changed
+
+- Expose `K8sResourceDetector`, which populates `k8s.pod.name` and `k8s.namespace.name`
```
cijothomas, line 7: I am curious about populating just these 2 attributes... Is that just a starting point, or we have limitations about capturing some other attributes from K8s?

gruebel: yeah, correct. These two are the easiest to retrieve without needing some adjustments of the Pod permissions. I planed to tackle the other ones in the future.

opentelemetry-resource-detectors/src/k8s.rs
```diff
@@ -1,24 +1,75 @@
...
-
-        // Create a Resource with Kubernetes attributes.
-        Resource::new(vec![
-            resource::K8S_POD_NAME.string(pod_name),
-            resource::K8S_NAMESPACE_NAME.string(namespace_name),
-            resource::K8S_NODE_NAME.string(node_name),
-        ])
+        let pod_name = env::var("HOSTNAME").ok();
+
+        let namespace = read_to_string(K8S_NAMESPACE_PATH).ok();
```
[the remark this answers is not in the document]
gruebel: yeah, this the official way to retrieve the namespace, which I need to leverage the API server in the future https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/

> Finally, the default namespace to be used for namespaced API operations is placed in a file at `/var/run/secrets/kubernetes.io/serviceaccount/namespace` in each container.

opentelemetry-resource-detectors/Cargo.toml
gruebel: yeah, they existed, but they were not exposed in the `lib.rs` and therefore never actually invoked 😅

opentelemetry-resource-detectors/CHANGELOG.md
gruebel: I'll create a separate issue, which you can assign to me

cijothomas (approved):

cijothomas:
@gruebel Oops! merged other PR so this got conflicts :(
Could you resolve!

PR #191, opened 2025-03-05
Fix userevents exporter

cijothomas:
Modeling Exporter as an actual `LogExporter`, and have Processor call it as if calling other exporter. Currently, the actual Export method was not invoked, but rather a different method was used.

No change in perf numbers.
(It was also tested and confirmed in main repo that invoking async method via executor has no measurable overhead than directly invoking a sync method.)

This will become relevant when https://github.com/open-telemetry/opentelemetry-rust/issues/1942 is addressed as it can avoid the dynamic dispatches.


lalitb (approved):
> (It was also tested and confirmed in main repo that invoking async method via executor has no measurable overhead than directly invoking a sync method.)

Nice if there are no perf impact :)

PR #196, opened 2025-03-07
UserEvents - Integration Test

cijothomas:
Might not work in CI, so can disable and run manually only.


lalitb (approved):
Nice work, Cijo!  I can see that troubleshooting and getting this to work must have taken time. We can build on this further :)

PR #198, opened 2025-03-08
feat: User_Events exporter - add support for serializing traceid and spanid

cijothomas:


opentelemetry-user-events-logs/src/lib.rs
```diff
@@ -133,6 +138,143 @@ mod tests {
         );
     }
 
+    #[ignore]
+    #[test]
+    fn integration_test_with_tracing() {
```
lalitb, line 143: Is earlier test subset of this - do we need them separate? This seems lots of repeated code in both tests, so checking.

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -193,14 +193,24 @@ impl UserEventsExporter {
...
-                        &chrono::DateTime::<chrono::Utc>::from(event_time),
-                    );
-                    eb.add_str("time", time, FieldFormat::Default, 0);
+
+                if let Some(trace_context) = log_record.trace_context() {
+                    cs_a_count += 1; // for ext_dt
+                    eb.add_struct("PartA", cs_a_count, 0);
+                    eb.add_struct("ext_dt", 2, 0);
+                    eb.add_str("traceId", trace_context.trace_id.to_string(), FieldFormat::Default, 0);
+                    eb.add_str("spanId", trace_context.span_id.to_string(), FieldFormat::Default, 0);
```
lalitb, line 206: traceFlags?

lalitb (approved):

opentelemetry-user-events-logs/src/logs/exporter.rs
cijothomas: to be added based on need. its optional.

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -193,14 +193,24 @@ impl UserEventsExporter {
...
+                }
+                else {
+                    eb.add_struct("PartA", cs_a_count, 0);
                 }
+
+                eb.add_str("time", time, FieldFormat::Default, 0);
```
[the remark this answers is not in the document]
cijothomas: good point. Maybe a good idea to avoid populating this in the upstream sdk itself (on a opt-in basis), as fetching time is not super cheap.

PR #202, opened 2025-03-17
Move `actix-web-opentelemetry` to contrib

jtescher:
## Changes

Moves the `actix-web-opentelemetry` crate from https://github.com/outtherelabs/actix-web-opentelemetry to this repo.


cijothomas (commented):
Many thanks for contributing this to OTel!
I'd like to discuss few things before merging
1. The name `actix-web-opentelemetry` might not be what OTel prefers. I'd suggest something like `opentelemetry-instrumentation-actix` (it should start with opentelemetry, in-line with rest of the crates). 
2. What should be the stability status? i.e, are we free to make breaking changes now and drive it to 1.0, at which point it can be marked stable with sem-ver guarantees?

I am okay with a simple lift-and-shift to the repo, and do the renaming+other changes(including breaking change) as future follow ups.
Once this is in, I can go ahead and modify both opentelemtry.io/docs and opentelemetry-demo project to use this component.

@open-telemetry/rust-approvers Please share your thoughts! 
(An additional background: almost all other OTel language implementations maintain instrumentation libraries like this in their contrib repos, Rust was so far the only repo without any!)

cijothomas:
Based on discussions on the community call yesterday,
1. We should accept as-is. (life-and-shift).
2. Publish using same name for at least one more release. During this one, make it clear that the name will change
3. See if there is a way to mark crate as deprecated and point to new name.

actix-web-opentelemetry/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+  "alloc",
+] }
+opentelemetry = { version = "0.28", default-features = false, features = [
+  "trace",
+] }
+opentelemetry-prometheus = { version = "0.28", optional = true }
+opentelemetry-semantic-conventions = { version = "0.28", features = [
+  "semconv_experimental",
+] }
+opentelemetry_sdk = { version = "0.28", optional = true, features = [
```
cijothomas, line 45: @lalitb raised an important point about this sdk dependency in the community call. Instrumentations should only rely on the opentelemetry crate.
However, lets do the lift-and-shift now, and then address issues in subsequent releases.

cijothomas (approved):
Many thanks!

PR #204, opened 2025-03-20
userevent logs - nit fix

cijothomas:
Remove emitting logs of its own to avoid rentrancy. The exporter is paired with a processor that expects exporter is safe from reentrancy. Exporter should not do any internal logs that can break this contract.


opentelemetry-user-events-logs/src/logs/reentrant_logprocessor.rs
```diff
@@ -26,6 +26,8 @@ impl<T: LogExporter> ReentrantLogProcessor<T> {
 impl<T: LogExporter> opentelemetry_sdk::logs::LogProcessor for ReentrantLogProcessor<T> {
     fn emit(&self, record: &mut SdkLogRecord, scope: &InstrumentationScope) {
         let log_tuple = &[(record as &SdkLogRecord, scope)];
+        // TODO: Using futures_executor::block_on can make the code non reentrant safe
```
cijothomas, line 29: same risk for EventHeader library, but that library is meant for low level interaction with Linux tracepoints, so unlikely to introduce logging of its own.

lalitb (approved):

PR #213, opened 2025-03-24
refactor: part b and c

psandana:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-etw-logs/CHANGELOG.md
+++ b/opentelemetry-etw-logs/CHANGELOG.md
@@ -18,6 +18,8 @@
 
 - `default` feature does not enable `spec_unstable_logs_enabled` anymore.
 
+- Added support for TraceId,SpanId
+
 ## v0.7.0
 
 - Bump msrv to 1.75.0
--- a/opentelemetry-etw-logs/CHANGELOG.md
+++ b/opentelemetry-etw-logs/CHANGELOG.md
@@ -22,6 +22,8 @@
 
 - Added support for TraceId,SpanId
 
+- Added support for populating cloud `role` and `roleInstance` from Resource's `service.name` and `service.instance.id` attributes respectively.
+
 ## v0.7.0
 
 - Bump msrv to 1.75.0
--- a/opentelemetry-etw-logs/CHANGELOG.md
+++ b/opentelemetry-etw-logs/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Renamed `logs_level_enabled` feature to `spec_unstable_logs_enabled` to match `opentelemetry` features.
 
 - `default` feature does not enable `spec_unstable_logs_enabled` anymore.
+- Bump opentelemetry and opentelemetry_sdk versions to 0.29
 
 - Added support for TraceId,SpanId
 
--- a/opentelemetry-etw-logs/CHANGELOG.md
+++ b/opentelemetry-etw-logs/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Renamed `logs_level_enabled` feature to `spec_unstable_logs_enabled` to match `opentelemetry` features.
 
 - `default` feature does not enable `spec_unstable_logs_enabled` anymore.
+
 - Bump opentelemetry and opentelemetry_sdk versions to 0.29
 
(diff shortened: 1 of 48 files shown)


opentelemetry-etw-logs/src/logs/exporter/part_b.rs
```diff
@@ -0,0 +1,65 @@
...
+    log_record: &opentelemetry_sdk::logs::SdkLogRecord,
+    level: Severity,
+    event_id: Option<i64>,
+    event_name: Option<&str>,
+) {
+    // Count fields in PartB
+    const COUNT_TYPE_NAME: u8 = 1u8;
+    const COUNT_SEVERITY_NUMBER: u8 = 1u8;
+
+    let field_count = COUNT_TYPE_NAME
```
cijothomas, line 15: minor: we can bookmark the location and update count later. It might be slightly faster, as it avoids check is_some() twice. user-events does this. same with .NET Geneva exporters.

opentelemetry-etw-logs/src/logs/exporter/part_c.rs
```diff
@@ -0,0 +1,127 @@
...
+                cs_c_count += 1;
+            }
+        }
+    }
+
+    // If there are additional PartC attributes, add them to the event
+    if cs_c_count > 0 {
+        event.add_struct("PartC", cs_c_count, field_tag);
+
+        for (key, value) in log_record.attributes_iter() {
```
cijothomas, line 45: Lets use the bookmarking strategy and avoid the dual iteration over attributes.

opentelemetry-etw-logs/src/logs/exporter/part_b.rs
psandana: Confirmed. There is a method to set the struct size in .NET implementation, but not in Rust version. Created an issue requesting it: https://github.com/microsoft/tracelogging/issues/73.

I suggest to leave this at it is, and compare performance when the tracelogging solution is available. Then we stick with the optimal solution.

psandana: As a side note, I would rather we measure, as `is_some()` method should be really cheap, while setting the field count afterwards has some indirect memory accesses.

opentelemetry-etw-logs/src/logs/exporter/part_c.rs
psandana: As explained before, we don't have bookmarking yet in ETW for Rust. Created issue for its implementation.

opentelemetry-etw-logs/src/logs/exporter/part_b.rs
cijothomas: https://github.com/microsoft/LinuxTracepoints-Rust/issues/5 This was also missing in the EventHeader, but was added for this reason. Good point about measuring before/after!

opentelemetry-etw-logs/src/logs/exporter/part_c.rs
```diff
@@ -0,0 +1,127 @@
...
+        match (key.as_str(), &value) {
+            (EVENT_ID, AnyValue::Int(value)) => {
+                event_id = Some(*value);
+                continue;
+            }
+            (EVENT_NAME_PRIMARY, AnyValue::String(value)) => {
+                event_name = Some(value.as_str());
+                continue;
+            }
+            (EVENT_NAME_SECONDARY, AnyValue::String(value)) => {
```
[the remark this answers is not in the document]
psandana: Done

opentelemetry-etw-logs/CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 
 - Added support for TraceId,SpanId
 
+- Added support for populating cloud `role` and `roleInstance` from Resource's `service.name` and `service.instance.id` attributes respectively.
```
cijothomas, line 25: I suggest to do refactoring on own PR, separate from adding features. It is a lot easier to review PR that only does one thing.

cijothomas (approved):
LGTM

PR #221, opened 2025-03-25
fix: Make usereventlogexporter re-entrant safe and handle shutdown

cijothomas:
1. Handle shutdown to exporter
2. No more internal logs - This limits self observability, but can add some metric, or eprintln() in the future, if this turns out to be an issue.

### DIFF
--- a/opentelemetry-user-events-logs/CHANGELOG.md
+++ b/opentelemetry-user-events-logs/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Make exporter reentrant-safe by removing logs that could be bridged back
   to itself.
 - Exporter now unregisters the `Provider` on `shutdown()`.
+  [221](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/221)
 
 ## v0.11.0
 
--- a/opentelemetry-user-events-logs/CHANGELOG.md
+++ b/opentelemetry-user-events-logs/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Added support for Populating Cloud RoleName, RoleInstance from Resource's
   "service.name" and "service.instance.id" attributes respectively.
+- Make exporter reentrant-safe by removing logs that could be bridged back
+  to itself.
 
 ## v0.11.0
 
--- a/opentelemetry-user-events-logs/CHANGELOG.md
+++ b/opentelemetry-user-events-logs/CHANGELOG.md
@@ -6,6 +6,7 @@
   "service.name" and "service.instance.id" attributes respectively.
 - Make exporter reentrant-safe by removing logs that could be bridged back
   to itself.
+- Exporter now unregisters the Tracepoint on `shutdown()`.
 
 ## v0.11.0
 
--- a/opentelemetry-user-events-logs/CHANGELOG.md
+++ b/opentelemetry-user-events-logs/CHANGELOG.md
@@ -6,7 +6,7 @@
   "service.name" and "service.instance.id" attributes respectively.
 - Make exporter reentrant-safe by removing logs that could be bridged back
   to itself.
-- Exporter now unregisters the Tracepoint on `shutdown()`.
+- Exporter now unregisters the `Provider` on `shutdown()`.
 
 ## v0.11.0
 
--- a/opentelemetry-user-events-logs/CHANGELOG.md
+++ b/opentelemetry-user-events-logs/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Make exporter reentrant-safe by removing logs that could be bridged back
   to itself.
 - Exporter now unregisters the `Provider` on `shutdown()`.
-  [221](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/221)
+  [#221](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/221)
 
 ## v0.11.0
 
--- a/opentelemetry-user-events-logs/src/logs/exporter.rs
+++ b/opentelemetry-user-events-logs/src/logs/exporter.rs
@@ -1,6 +1,6 @@
 use eventheader::{FieldFormat, Level, Opcode};
 use eventheader_dynamic::{EventBuilder, EventSet, Provider};
-use opentelemetry::{otel_debug, otel_info};
+use opentelemetry::otel_debug;
 use opentelemetry_sdk::Resource;
 use std::sync::Arc;
 use std::{fmt::Debug, sync::Mutex};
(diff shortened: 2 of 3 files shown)


opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -327,20 +327,18 @@ impl UserEventsExporter {
 
                 let result = eb.write(event_set, None, None);
                 if result > 0 {
-                    // Specially log the case where there is no listener and size exceeding.
+                    // Specially treat the case where there is no listener and size exceeding.
                     if result == 9 {
-                        otel_debug!(name: "UserEvents.EventWriteFailed", result = result, reason = "No listener. This can occur when there was a listener but it was removed before the event was written");
+                        Err(OTelSdkError::InternalFailure("Failed to write event to user_events tracepoint as there is no listener. This can occur when there was a listener when we started serializing event, but it was removed before the event was written".to_string()))
```
utpilla, line 332:
```suggestion
                        Err(OTelSdkError::InternalFailure("Failed to write event to user_events tracepoint as there is no listener. This can occur if there was a listener when we started serializing the event, but it was removed before the event was written".to_string()))
```

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -362,10 +360,12 @@ impl Debug for UserEventsExporter {
 
 impl opentelemetry_sdk::logs::LogExporter for UserEventsExporter {
     async fn export(&self, batch: opentelemetry_sdk::logs::LogBatch<'_>) -> OTelSdkResult {
+        let mut res = Ok(());
         for (record, instrumentation) in batch.iter() {
-            let _ = self.export_log_data(record, instrumentation);
+            // no, res is not overwritten as the batch will only have one item
```
utpilla, line 365: I think we can remove the no here 😁

```suggestion
            // res is not overwritten as the batch will only have one item
```

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -327,20 +327,18 @@ impl UserEventsExporter {
 
                 let result = eb.write(event_set, None, None);
                 if result > 0 {
-                    // Specially log the case where there is no listener and size exceeding.
+                    // Specially treat the case where there is no listener and size exceeding.
```
utpilla, line 330:
```suggestion
                    // Specially treat the case where there is no listener or payload size exceeds the limit.
```

opentelemetry-user-events-logs/src/logs/exporter.rs
lalitb, line 365: If batch is always once, loop can be avoided.
```rust
let (record, instrumentation) = batch.iter().next().expect("....");
self.export_log_data(record,instruction)
```

opentelemetry-user-events-logs/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 - Added support for Populating Cloud RoleName, RoleInstance from Resource's
   "service.name" and "service.instance.id" attributes respectively.
+- Make exporter reentrant-safe by removing logs that could be bridged back
+  to itself.
+- Exporter now unregisters the Tracepoint on `shutdown()`.
```
utpilla, line 9: Not fully aware of the implementation details. Is it `Provider` or `Tracepoint`?

```suggestion
- Exporter now unregisters the `Provider` on `shutdown()`.
```

utpilla (approved):

opentelemetry-user-events-logs/CHANGELOG.md
utpilla, line 9: Add the PR link as well?

lalitb (approved):

opentelemetry-user-events-logs/src/logs/exporter.rs
lalitb: I liked the no actually  jk :)

cijothomas: Nice! just pushed that change.

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -362,10 +360,11 @@ impl Debug for UserEventsExporter {
...
 impl opentelemetry_sdk::logs::LogExporter for UserEventsExporter {
     async fn export(&self, batch: opentelemetry_sdk::logs::LogBatch<'_>) -> OTelSdkResult {
-        for (record, instrumentation) in batch.iter() {
-            let _ = self.export_log_data(record, instrumentation);
-        }
-        Ok(())
+        let (record, instrumentation) = batch
+            .iter()
+            .next()
+            .expect("batch is expected to have one and only one record");
```
utpilla, line 366: Should we use `expect()` here or be more defensive and use `if let Some(item)` construct? We know that we would never send an empty batch, but it might be sensible to still add this check in the exporter?

utpilla: I would be in favor of doing error handling here. It also aligns with our design principle:

https://github.com/open-telemetry/opentelemetry-rust/blob/main/docs/design/logs.md#key-design-principles

> Robust error handling, returning Result where possible instead of panicking.

opentelemetry-user-events-logs/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 - Added support for Populating Cloud RoleName, RoleInstance from Resource's
   "service.name" and "service.instance.id" attributes respectively.
+- Make exporter reentrant-safe by removing logs that could be bridged back
+  to itself.
+- Exporter now unregisters the `Provider` on `shutdown()`.
+  [221](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/221)
```
utpilla, line 10:
```suggestion
  [#221](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/221)
```

opentelemetry-user-events-logs/src/logs/exporter.rs
cijothomas: agree. made some change in latest commit to reflect this, PTAL.

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -362,10 +360,19 @@ impl Debug for UserEventsExporter {
 
 impl opentelemetry_sdk::logs::LogExporter for UserEventsExporter {
     async fn export(&self, batch: opentelemetry_sdk::logs::LogBatch<'_>) -> OTelSdkResult {
-        for (record, instrumentation) in batch.iter() {
-            let _ = self.export_log_data(record, instrumentation);
+        let mut iter = batch.iter();
+        if let Some((record, instrumentation)) = iter.next() {
+            if iter.next().is_some() {
```
utpilla, line 365: Let's export the first element and then check this.

In fact, I think it's okay to even get rid of this check. We could have a stance that the exporter would only export the first element of the batch. If the batch contained more than one items, those items would be discarded.

cijothomas: let me revisit this

cijothomas: addressed in https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/222/files

PR #226, opened 2025-03-28
fix: user_events log exporter minor fixes and test improvements, doc additions

cijothomas:


opentelemetry-user-events-logs/src/lib.rs
```diff
@@ -1,5 +1,72 @@
...
+//! ## Getting Started
+//!
+//! To use the `user_events` exporter, you can set up a logger provider as follows:
+//!
+//! ```rust
+//! use opentelemetry_sdk::logs::LoggerProviderBuilder;
+//! use opentelemetry_user_events_logs::UserEventsLoggerProviderBuilderExt;
+//!
+//! let logger_provider = LoggerProviderBuilder::default()
+//!     .with_user_event_exporter("myprovider")
```
lalitb, line 64: should we rename this method name to `with_user_events_exporter`, as separate PR?

opentelemetry-user-events-logs/src/lib.rs
```diff
@@ -1,5 +1,72 @@
...
+//! Event Tracing for Windows (ETW) on Windows and leverages Linux Tracepoints
+//! to enable user processes to create traceable events and data. These events
+//! can be analyzed using existing tools like `ftrace` and `perf`.
+//!
+//! ## Key Features of `user_events`
+//!
+//! - **Efficient Tracing Path**: Provides a faster path for tracing from
+//!   user-mode applications by utilizing kernel-mode memory address space.
+//! - **Selective Event Export**: User processes can export telemetry events
+//!   only when they are useful, i.e., when the registered set of tracepoint
```
utpilla, line 16: I think we can be more specific about what we mean by "useful".

> User processes can export telemetry events only when they are useful

User processes have the ability to selectively export events that have an interested subscriber.

opentelemetry-user-events-logs/src/lib.rs
```diff
@@ -1,5 +1,72 @@
...
+//!   events are enabled.
+//!
+//! ## Purpose of This Exporter
```
utpilla, line 19:
```suggestion
//! ## Purpose of this Exporter
```

opentelemetry-user-events-logs/src/lib.rs
```diff
@@ -1,5 +1,72 @@
...
+//! - **Captured by Local Agents**: Agents running locally can listen for
+//!   specific events within the `user_events` subsystem.
+//! - **Monitored in Real-Time**: Events can be monitored in real-time using
+//!   Linux tools like `perf` or `ftrace`.
+//!
+//! ## Prerequisites
+//!
+//! - **Linux Kernel Version**: Requires Linux kernel 6.4 or later with
+//!   `user_events` support enabled.
+//! - **Tools**: Tools like `perf` or `ftrace` must be installed for decoding
```
utpilla, line 34: These tools are not a pre-requisite for using the exporter, right? They seem to be only related to troubleshooting.

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -279,30 +285,31 @@ impl UserEventsExporter {
...
-                            AnyValue::Map(_value) => "".to_string(),
-                            &_ => "".to_string(),
-                        },
-                        FieldFormat::Default,
-                        0,
-                    );
-                    cs_b_count += 1;
-                }
-                if level != Level::Invalid {
-                    eb.add_value("severityNumber", otel_severity as i16, FieldFormat::SignedInt, 0);
```
lalitb, line 303: seems we were already exporting `otel_severity`. What exactly is the change?

opentelemetry-user-events-logs/src/lib.rs
```diff
@@ -1,5 +1,72 @@
...
+//!   and analyzing the exported events.
+//!
+//! ## Synchronous Export
+//!
+//! This exporter writes telemetry events to the `user_events` subsystem
+//! synchronously, without any buffering or batching. Each event is immediately
```
utpilla, line 40: I think we can also mention that the export is non-blocking.

```suggestion
//! synchronously, without any buffering or batching. The exporter is non-blocking and each event is immediately
```

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -103,23 +103,31 @@ impl UserEventsExporter {
...
                 eb.add_value(field_name, *f, FieldFormat::Float, 0);
+                true
             }
             AnyValue::String(s) => {
                 eb.add_str(field_name, s.as_str(), FieldFormat::Default, 0);
+                true
             }
             // TODO: Handle other types. Arrays are required in Trace for storing Links.
-            _ => (),
+            _ => false,
```
utpilla, line 130: How about we add an empty string as value? That way we would at least send information about the key.

opentelemetry-user-events-logs/src/lib.rs
cijothomas: good point, will do this PR itself.

opentelemetry-user-events-logs/src/lib.rs
cijothomas: The wording states they are required for decoding/analyzing only, but tried to make it more clear in latest commit

opentelemetry-user-events-logs/src/logs/exporter.rs
cijothomas: removed the if() check that was unnecessary.

opentelemetry-user-events-logs/src/logs/exporter.rs
cijothomas: yea that'll make less silent drop!. Added

opentelemetry-user-events-logs/src/lib.rs
utpilla: The commit update is definitely better than before, but my point is that if these tools are not required to use the exporter, then they shouldn't be placed under "Prerequisites".  Perquisites should only be a list of things that are absolutely necessary and without whom you cannot use the product.

Do we want the user to worry about the presence of `perf` or `ftrace` before using this exporter? In this case, it seems like all that a user needs to ensure before using the exporter is that they have the right kernel version. These tools can probably be mentioned in their own section, like "Troubleshooting/Testing".

lalitb (approved):

PR #227, opened 2025-03-29
feat: Geneva Uploader : Add config service client

lalitb:
## Changes

Implements the config service 
  - Support Certificate based authentication
  - Getting Upload credentials from config service
  
The sequence diagram (also added in README.md)

```mermaid
sequenceDiagram
    participant App as User
    participant Client as GenevaConfigClient
    participant TLS as native_tls
    participant GCS as Geneva Config Service
    App->>Client: new(GenevaConfigClientConfig)
    Client->>TLS: Load PKCS#12 cert
    TLS-->>Client: native_tls::TlsConnector
    Client->>Client: Build reqwest::Client with mTLS
    App->>Client: get_ingestion_info()
    alt Token in cache and not expired
        Client->>App: Return cached (IngestionGatewayInfo, MonikerInfo)
    else Cache miss or token expired
        Client->>Client: Build HTTP GET URL
        Client->>GCS: Send HTTPS request with mTLS\n+ Query Params & Headers
        GCS-->>Client: JSON response (200 OK or error)
        alt Response contains valid moniker
            Client->>Client: Parse IngestionGatewayInfo and MonikerInfo
            Client->>Client: Cache new token + expiry
            Client->>App: Return new (IngestionGatewayInfo, MonikerInfo)
        else No valid moniker
            Client->>App: Error (MonikerNotFound)
        end
    end
```


opentelemetry-exporter-geneva/geneva-uploader/Cargo.toml
```diff
@@ -4,7 +4,26 @@ version = "0.1.0"
...
 
 [dependencies]
-opentelemetry-proto = {workspace = true, default-features = false, features = ["logs"]}
+opentelemetry-proto = {workspace = true, default-features = false, features = ["logs", "gen-tonic-messages"]}
+base64 = "0.22"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = { version = "1.0", features = ["raw_value"] }
+uuid = { version = "1.0", features = ["v4"] }
+# TODO - support both native-tls and rustls
+reqwest = { version = "0.12", features = ["native-tls", "native-tls-alpn"]}
```
lalitb, line 13: TODO - this would be abstracted with http-client, to remove the indirect dependency on tokio runtime.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+/// ```
+#[allow(dead_code)]
+#[derive(Clone)]
+pub(crate) enum AuthMethod {
+    /// Certificate-based authentication
+    ///
+    /// # Arguments
+    /// * `path` - Path to the PKCS#12 (.p12) certificate file
+    /// * `password` - Password to decrypt the PKCS#12 file
+    Certificate { path: String, password: String },
```
utpilla, line 49: Use [`Path`](https://doc.rust-lang.org/std/path/struct.Path.html) instead of `String`?
```suggestion
    Certificate { path: Path, password: String },
```

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+///     auth_method: AuthMethod::Certificate {
+///         path: "/path/to/cert.p12".to_string(),
+///         password: "password".to_string(),
+///     },
+/// };
+/// ```
+#[allow(dead_code)]
+#[derive(Clone)]
+pub(crate) struct GenevaConfigClientConfig {
+    pub endpoint: String,
```
utpilla, line 112: nit: Mark the access modifiers for each of these fields as `pub(crate)` similar to `IngestionGatewayInfo` struct.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+    #[serde(rename = "IsPrimaryMoniker")]
+    is_primary_moniker: bool,
+}
+
+#[allow(dead_code)]
+#[derive(Debug, Deserialize)]
+struct GenevaResponse {
+    #[serde(rename = "IngestionGatewayInfo")]
+    ingestion_gateway_info: IngestionGatewayInfo,
+    // Make storage_account_keys optional since it might not be present in all responses
```
utpilla, line 155: Is this a TODO?

```suggestion
    // TODO: Make storage_account_keys optional since it might not be present in all responses
```

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+    #[serde(rename = "StorageAccountKeys", default)]
+    storage_account_keys: Vec<StorageAccountKey>,
+    // Keep tag_id as it might be used for validation
+    #[serde(rename = "TagId", default)]
+    tag_id: Option<String>,
```
utpilla, line 160: Is tag_id also not always expected to be a part of the response?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+}
+
+#[allow(dead_code)]
+struct CachedAuthData {
+    // Store the complete token and moniker info
+    auth_info: Option<(IngestionGatewayInfo, MonikerInfo)>,
```
utpilla, line 166: We don't expect this struct to be partially populated, right? If yes, instead of making each field optional, could we just `Option<CachedAuthData>` in the config client?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/README.md
```diff
@@ -0,0 +1,33 @@
+### GenevaConfigClient Flow (Certificate-Based Authentication)
+
+The diagram below illustrates how the `GenevaConfigClient` initializes with a client certificate (in PKCS#12 format) and fetches ingestion information from the Geneva Config Service using mutual TLS (mTLS). It includes the flow for loading the certificate, handling cached tokens, making authenticated requests, and parsing the response (including primary diagnostic monikers).
```
utpilla, line 3:
```suggestion
The diagram below illustrates how the `GenevaConfigClient` is initialized with a client certificate (in PKCS#12 format) and then used to fetch ingestion information from the Geneva Config Service using mutual TLS (mTLS). It includes the flow for loading the certificate, handling cached tokens, making authenticated requests, and parsing the response (including primary diagnostic monikers).
```

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+    /// * `GenevaConfigClientError::Certificate` - If certificate authentication fails
+    /// * `GenevaConfigClientError::Io` - If reading certificate files fails
+    /// * `GenevaConfigClientError::Tls` - If TLS configuration fails
+    #[allow(dead_code)]
+    pub(crate) async fn new(config: GenevaConfigClientConfig) -> Result<Self> {
+        let mut client_builder = Client::builder()
+            .http1_only()
+            .timeout(Duration::from_secs(30))
+            .pool_idle_timeout(Duration::from_secs(90))
+            .pool_max_idle_per_host(16);
```
utpilla, line 203: Why are we setting this configuration?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+
+        match &config.auth_method {
+            // TODO: Certificate auth would be removed in favor of managed identity.,
+            // This is for testing, so we can use self-signed certs, and password in plain text.
```
utpilla, line 207: If this is just for testing, could use PEM based certificates? I'm assuming it's simpler as we wouldn't have to deal with passwords at all in that case.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+            AuthMethod::Certificate { path, password } => {
+                // Read the PKCS#12 file
+                let p12_bytes = fs::read(path)?;
+                let identity = Identity::from_pkcs12(&p12_bytes, password)?;
+                let tls_connector =
+                    configure_tls_connector(native_tls::TlsConnector::builder(), identity)
+                        .build()?;
+                client_builder = client_builder.use_preconfigured_tls(tls_connector);
```
utpilla, line 215: Could we have used [`use_native_tls`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.use_native_tls) method here instead of creating the `tls_connector` ourselves?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+            }
+            AuthMethod::ManagedIdentity => {
+                return Err(GenevaConfigClientError::Certificate(
```
utpilla, line 218: This is not a certificate error. Either use `InternalError` or use the `todo!` macro.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+    ///
+    /// # Returns
+    /// * `Result<Self>` - A new client instance or an error
+    ///
+    /// # Errors
+    /// * `GenevaConfigClientError::Certificate` - If certificate authentication fails
+    /// * `GenevaConfigClientError::Io` - If reading certificate files fails
+    /// * `GenevaConfigClientError::Tls` - If TLS configuration fails
+    #[allow(dead_code)]
+    pub(crate) async fn new(config: GenevaConfigClientConfig) -> Result<Self> {
```
utpilla, line 198:
```suggestion
    pub(crate) fn new(config: GenevaConfigClientConfig) -> Result<Self> {
```

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+    ///
+    /// # Arguments
+    /// * `config` - The client configuration
+    ///
+    /// # Returns
+    /// * `Result<Self>` - A new client instance or an error
+    ///
+    /// # Errors
+    /// * `GenevaConfigClientError::Certificate` - If certificate authentication fails
+    /// * `GenevaConfigClientError::Io` - If reading certificate files fails
```
utpilla, line 195: It doesn't look like these methods could return these errors.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+        url.push_str(&tag_id);
+
+        let req_id = Uuid::new_v4().to_string();
+
+        let response = self
+            .http_client
+            .get(&url)
+            .header(
+                "User-Agent",
+                format!("{}-{}", agent_identity, agent_version),
```
utpilla, line 417: nit: This header value doesn't change. We could avoid creating it for each request.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+            .map_err(GenevaConfigClientError::Http)?;
+
+        let status = response.status();
+        let body = response.text().await?;
+
+        if status.is_success() {
+            let parsed = match serde_json::from_str::<GenevaResponse>(&body) {
+                Ok(response) => response,
+                Err(_) => {
+                    return Err(GenevaConfigClientError::AuthInfoNotFound(
```
utpilla, line 432: Why are we using `GenevaConfigClientError::AuthInfoNotFound` here? This would be some deserialization error, right?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+                        "No primary diag moniker found in storage accounts".to_string(),
+                    ));
+                }
+            };
+
+            for account in &parsed.storage_account_keys {
+                if account.is_primary_moniker && account.account_moniker_name.contains("diag") {
+                    let moniker_info = MonikerInfo {
+                        name: account.account_moniker_name.clone(),
```
utpilla, line 441: Could we use [`take`](https://doc.rust-lang.org/std/mem/fn.take.html) instead of cloning here?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+    /// Internal method that actually fetches data from Geneva Config Service
+    async fn fetch_ingestion_info(&self) -> Result<(IngestionGatewayInfo, MonikerInfo)> {
+        let agent_identity = "GenevaUploader"; // TODO make this configurable
+        let agent_version = "0.1"; // TODO make this configurable
+        let identity = format!(
+            "Tenant=Default/Role=GcsClient/RoleInstance={}",
+            agent_identity
+        );
+
+        let encoded_identity = general_purpose::STANDARD.encode(&identity);
```
utpilla, line 378: This also looks like something that doesn't need to be computed for each call.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+                    }
+                }
+            }
+
+            // Update cache with fresh data
+            cached_data.auth_info = Some((fresh_token_info, fresh_moniker_info));
+            cached_data.token_endpoint = Some(token_endpoint);
+            cached_data.token_expiry = token_expiry;
+
+            if let (Some(info), Some(endpoint)) =
```
utpilla, line 357: Do we need this if else check? At this point, the cache has just been updated under a write lock. Why use the `if let()` check.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+        "windows" => "Windows",
+        "linux" => "Linux",
+        other => other,
+    }
+}
+
+fn extract_endpoint_from_token(token: &str) -> Result<String> {
+    let parts: Vec<&str> = token.split('.').collect();
+    if parts.len() != 3 {
+        return Err(GenevaConfigClientError::Certificate(
```
utpilla, line 474: We should use some other error type here not `Certificate`.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+        0 => payload.to_string(),
+        2 => format!("{}==", payload),
+        3 => format!("{}=", payload),
+        _ => payload.to_string(),
+    };
+
+    let decoded = general_purpose::URL_SAFE_NO_PAD
+        .decode(payload)
+        .map_err(|e| {
+            GenevaConfigClientError::Certificate(format!("Failed to decode JWT: {}", e))
```
utpilla, line 491: Maybe we should add a JWT related variant to the error enum.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+
+fn extract_endpoint_from_token(token: &str) -> Result<String> {
+    let parts: Vec<&str> = token.split('.').collect();
+    if parts.len() != 3 {
+        return Err(GenevaConfigClientError::Certificate(
+            "Invalid JWT token format".into(),
+        ));
+    }
+
+    // Decode the payload
```
utpilla, line 479:
```suggestion
    // Decode the payload
    
    //Convert to Base64 string
```

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,536 @@
...
+
+    fn parse_token_expiry(&self, expiry_str: &str) -> Option<DateTime<Utc>> {
+        // Attempt to parse the ISO 8601 datetime string
+        DateTime::parse_from_rfc3339(expiry_str)
+            .ok()
+            .map(|dt| dt.with_timezone(&Utc))
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn get_token_expiry(&self) -> Option<DateTime<Utc>> {
```
utpilla, line 245: Do we expect this method to be called?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Done

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: yes, made it non-optional parameter.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Make sense. Updated.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: The problem is - reqwest has no built-in PEM loader when using native-tls. And pkcs12 was easy, being self-contained and password protected. Also, plan is to remove the certificate based auth in future, so didn't want to spend too much time on troubleshooting with PEM :)

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Good catch, this is not required. Have removed it now.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Good catch - have precomputed them within GenevaConfigClient::new()

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Yes correct. Have precomputed and stored it with GenevaConfigClient now.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Yes, it is deserialization error, if the response can't be converted into `GenevaResponse` object, which actually means that auth info couldn't be obtained from the server.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Yes, have done that now.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: On second thought, we probably don’t need all of these settings. Since the request is only made once every 3 days (based on token expiry), there’s no need to keep idle connections around or configure max connections per host — the defaults should suffice. However, since the server doesn’t support HTTP/2, we do need to enforce HTTP/1. Also, the timeout should ideally be configurable — I’ve added a TODO for that.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Actually, if was much easier to take the ownership of parsed data. Have modified accordingly. Thanks for the suggestion :)

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: using `GenevaConfigClientError::JwtTokenError` now.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Have added `GenevaConfigClientError::AuthMethodNotImplemented`.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: I think `async` would be idiomatic, and probably future proof here - specifically it performs file I/O for certificates,  and potentially would need network i/o for Managed Identity too.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: If I recall correctly, I had to change from `use_native_tls` to `use_preconfigured_tls` - to support self-signed certificates and enforce TLS1.2. Once those constraints are no longer needed, we can revert to using the simpler `use_native_tls` option.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Have updated the error as per the implementation. Let me know if there is something still missing.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: I anticipated this could be required when Geneva config fetch is added. But have removed it, and will bring it back at that time.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
utpilla: It would be good to add a TODO for that.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
utpilla: I think a `todo!` macro would have been simpler as you would now also have to remember to remove this newly added variant once you implement the logic for Manager Identity. Whatever works for you. 🙂

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
utpilla: This method is not performing any async operation. It would be against the idiomatic way to mark a method async when it does not have any await call in its implementation. If in future, this method is updated to perform asynchronous I/O, we can add the `async` keyword then.

lalitb: fair enough. Have made it sync.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: yes done.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,552 @@
...
+
+        let token_expiry = self
+            .parse_token_expiry(&fresh_ingestion_gateway_info.auth_token_expiry_time)
+            .ok_or_else(|| {
+                GenevaConfigClientError::InternalError("Failed to parse token expiry".into())
+            })?;
+
+        let token_endpoint = extract_endpoint_from_token(&fresh_ingestion_gateway_info.auth_token)
+            .map_err(|e| {
+                GenevaConfigClientError::Certificate(format!(
```
utpilla, line 374: This is not valid for `CertificateError`. In fact, there is no need to map the error. `extract_endpoint_from_token` method would only return `JwtTokenError` which is already a `GenevaCVonfigClientError` variant.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,552 @@
...
+                client_builder = client_builder.use_preconfigured_tls(tls_connector);
+            }
+            AuthMethod::ManagedIdentity => {
+                return Err(GenevaConfigClientError::AuthMethodNotImplemented(
+                    "Managed Identity authentication is not implemented yet".into(),
+                ));
+            }
+        }
+
+        let agent_identity = "GenevaUploader".to_string();
```
utpilla, line 243: We don't need `to_string()` for `agent_identity` and `agent_version`.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,552 @@
...
+            http_client,
+            cached_data: RwLock::new(None),
+            precomputed_url_prefix: pre_url,
+            agent_identity: agent_identity.to_string(), // TODO make this configurable
+            agent_version: "1.0".to_string(),           // TODO make this configurable
+            static_headers,
+        })
+    }
+
+    fn parse_token_expiry(&self, expiry_str: &str) -> Option<DateTime<Utc>> {
```
utpilla, line 285: This method does not `&self` parameter.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,552 @@
...
+        let token_endpoint = extract_endpoint_from_token(&fresh_ingestion_gateway_info.auth_token)
+            .map_err(|e| {
+                GenevaConfigClientError::Certificate(format!(
+                    "Failed to extract endpoint from token: {}",
+                    e
+                ))
+            })?;
+
+        // Now update the cache with exclusive write access
+        {
```
utpilla, line 381: We could get rid of the extra pair of parenthesis.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+            auth_method: AuthMethod::Certificate {
+                path: PathBuf::from(temp_p12_file.path().to_string_lossy().to_string()),
+                password,
+            },
+        };
+
+        let client = GenevaConfigClient::new(config).unwrap();
+        let result = client.get_ingestion_info().await;
+
+        assert!(result.is_err());
```
utpilla, line 159: You could simplify the tests that check for errors by using `unwrap_err`.

```rust
let err = result.unwrap_err();
match err {
    GenevaConfigClientError::RequestFailed { status, .. } => {
        assert_eq!(status, 403);
    }
    _ => panic!("Expected RequestFailed error, got: {:?}", err),
}
```

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+                password: "test".to_string(),
+            },
+        };
+
+        let result = GenevaConfigClient::new(config);
+
+        assert!(result.is_err());
+        match result {
+            Err(err) => match err {
+                crate::config_service::client::GenevaConfigClientError::Io(_) => {
```
utpilla, line 243: I think returning a `CertificateError` instead of `Io` would be more intuitive. In fact, we could only keep `CertificateError` variant and remove the `Io` variant. I believe reading the certificate was the only place where we could have returned the `Io` error.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
utpilla: If we are following this convention, then let's be consistent. Let's remove the `SerdeJson` variant altogether. We have another case of deserializing the JWT payload where we return a `SerdeJson` error. Let's return a `JwtTokenError` with the message "Failed to parse {error}".

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -0,0 +1,552 @@
...
+    token_endpoint: String,
+    // Store expiry separately for quick access
+    token_expiry: DateTime<Utc>,
+}
+
+#[allow(dead_code)]
+pub(crate) struct GenevaConfigClient {
+    config: GenevaConfigClientConfig,
+    http_client: Client,
+    cached_data: RwLock<Option<CachedAuthData>>,
```
utpilla, line 192: nit: Add a TODO here to revisit this and check if we can get rid of the `RwLock`.

utpilla (approved):
Left some suggestions.

PR #232, opened 2025-04-01
fix: user_events log: Handle empty event_name

lalitb:
## Changes

Handle scenario where event_name is empty string.

```rust
error!(name: "", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io", message = "This is an example message");)
```

### DIFF
--- a/opentelemetry-user-events-logs/src/logs/exporter.rs
+++ b/opentelemetry-user-events-logs/src/logs/exporter.rs
@@ -190,7 +190,8 @@ impl UserEventsExporter {
                 // for each event.
                 // TODO: What if the event name is not provided? "Log" is used as default.
                 // TODO: Should event_tag be non-zero?
-                eb.reset(log_record.event_name().unwrap_or("Log"), 0);
+                let event_name = log_record.event_name().filter(|s| !s.is_empty()).unwrap_or("Log");
+                eb.reset(event_name, 0);
 
                 eb.add_value("__csver__", 1024, FieldFormat::UnsignedInt, 0); // 0x400 in hex
 
--- a/opentelemetry-user-events-logs/src/logs/exporter.rs
+++ b/opentelemetry-user-events-logs/src/logs/exporter.rs
@@ -190,7 +190,10 @@ impl UserEventsExporter {
                 // for each event.
                 // TODO: What if the event name is not provided? "Log" is used as default.
                 // TODO: Should event_tag be non-zero?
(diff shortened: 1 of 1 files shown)


opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -190,7 +190,8 @@ impl UserEventsExporter {
                 // for each event.
                 // TODO: What if the event name is not provided? "Log" is used as default.
                 // TODO: Should event_tag be non-zero?
-                eb.reset(log_record.event_name().unwrap_or("Log"), 0);
+                let event_name = log_record.event_name().filter(|s| !s.is_empty()).unwrap_or("Log");
```
utpilla, line 193: What about whitespace only string?

utpilla, line 193: Also, I think it would be better to validate this upstream at SDK level itself in `SdkLogRecord::set_event_name` method. Why allow users to set empty/whitespace only strings at all?

lalitb: we can control it at the appender level - don't set event_name if it's empty string. I am fine to move the fix there we all agree.,

cijothomas: event_name() is used later in this file too, can you fix that too.

cijothomas: in an upcoming release, we can consider validating event_name in the sdk itself.

cijothomas:
@psandana Would need similar validation in ETW-logs too

opentelemetry-user-events-logs/src/logs/exporter.rs
lalitb: It's better to handle it at the appender level, if we decide to do it at all, since the event is already parsed there. No need to check again in the SDK.

utpilla: @lalitb There are two cases where we would benefit from doing the validation at the SDK level:
- We can only do this for the appenders we host in the main repo. We can't enforce this validation for custom appender implementations
-  Since the `Logger` and `LogRecord` APIs are public, technically a user can emit logs using these APIs directly without having to use any kind of appender.

utpilla (approved):

PR #233, opened 2025-04-02
chore: Prepare release to pickup 0.29

cijothomas:
Some crates had version already updated, some didn't. This updates the missing ones.

### DIFF
--- a/opentelemetry-user-events-trace/Cargo.toml
+++ b/opentelemetry-user-events-trace/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "opentelemetry-user-events-trace"
 description = "OpenTelemetry-Rust exporter to user_events"
-version = "0.10.0"
+version = "0.1.0"
 edition = "2021"
 homepage = "https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-user-events-traces"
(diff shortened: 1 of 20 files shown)


opentelemetry-user-events-trace/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "opentelemetry-user-events-trace"
 description = "OpenTelemetry-Rust exporter to user_events"
-version = "0.10.0"
```
lalitb, line 4: Oh, that was not copy-paste, but I thought of directly publish the much stable 0.10.0 version :)

lalitb (approved):

gruebel (approved):
🍻

opentelemetry-user-events-trace/Cargo.toml
cijothomas: since tracing signal itself is far from stable upstream, this can't go stable anyway. I think 0.1.0 is good first release, and we can totally skip few to match the versions across etw/user-event exporters too.

lalitb: My comment was in lighter note, it was copy paste error :)

cijothomas: I need ☕ before reviewing Github 🤣

PR #234, opened 2025-04-03
feat: stabilize etw exporters

psandana:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-etw-logs/Cargo.toml
+++ b/opentelemetry-etw-logs/Cargo.toml
@@ -11,7 +11,6 @@ keywords = ["opentelemetry", "log", "trace", "etw"]
 license = "Apache-2.0"
 
 [dependencies]
-tracelogging = "1.2.1"
 tracelogging_dynamic = "1.2.1"
 opentelemetry = { workspace = true, features = ["logs"] }
 opentelemetry_sdk = { workspace = true, features = ["logs"] }
--- a/opentelemetry-etw-logs/Cargo.toml
+++ b/opentelemetry-etw-logs/Cargo.toml
@@ -17,7 +17,7 @@ opentelemetry_sdk = { workspace = true, features = ["logs"] }
 tracing = { version = "0.1", optional = true }
 serde_json = { version = "1.0.113", optional = true }
 futures-executor = "0.3.31"
-chrono = { version = "0.4.40", features = ["alloc"] }
+chrono = { version = "0.4.40", default-features = false, features = ["alloc", "std"] }
 
 [dev-dependencies]
 opentelemetry-appender-tracing = { workspace = true }
--- a/opentelemetry-etw-logs/Cargo.toml
+++ b/opentelemetry-etw-logs/Cargo.toml
@@ -15,6 +15,7 @@ tracelogging = "1.2.1"
 tracelogging_dynamic = "1.2.1"
 opentelemetry = { workspace = true, features = ["logs"] }
 opentelemetry_sdk = { workspace = true, features = ["logs"] }
+tracing = { version = "0.1", optional = true }
 # TODO: Make json optional
 serde_json = "1.0.113"
 futures-executor = "0.3.31"
@@ -36,7 +37,8 @@ spec_unstable_logs_enabled = [
     "opentelemetry/spec_unstable_logs_enabled",
     "opentelemetry_sdk/spec_unstable_logs_enabled",
 ]
-default = []
+internal-logs = ["tracing", "opentelemetry/internal-logs", "opentelemetry_sdk/internal-logs"]
+default = ["internal-logs"]
 
 [[example]]
 name = "basic"
--- a/opentelemetry-etw-logs/Cargo.toml
+++ b/opentelemetry-etw-logs/Cargo.toml
@@ -17,7 +17,7 @@ opentelemetry_sdk = { workspace = true, features = ["logs"] }
 tracing = { version = "0.1", optional = true }
 serde_json = { version = "1.0.113", optional = true }
 futures-executor = "0.3.31"
-chrono = "0.4.40"
+chrono = { version = "0.4.40", features = ["alloc"] }
 
 [dev-dependencies]
 opentelemetry-appender-tracing = { workspace = true }
--- a/opentelemetry-etw-logs/Cargo.toml
+++ b/opentelemetry-etw-logs/Cargo.toml
@@ -19,6 +19,7 @@ tracing = { version = "0.1", optional = true }
 # TODO: Make json optional
 serde_json = "1.0.113"
 futures-executor = "0.3.31"
+chrono = "0.4.40"
 
 [dev-dependencies]
 opentelemetry-appender-tracing = { workspace = true }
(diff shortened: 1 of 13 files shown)


opentelemetry-etw-logs/src/logs/exporter/common.rs
```diff
@@ -22,6 +23,7 @@ pub fn add_attribute_to_event(event: &mut tld::EventBuilder, key: &Key, value: &
         AnyValue::Bytes(b) => {
             event.add_binaryc(key.as_str(), b.as_slice(), tld::OutType::Default, 0);
         }
+        #[cfg(feature = "serde_json")]
```
cijothomas, line 26: when the feature is not enabled, can we modify the catch_all arm to write the field with an empty string as field value.

opentelemetry-etw-logs/src/logs/exporter/part_a.rs
```diff
@@ -86,12 +85,19 @@ fn populate_time(
...
+    use chrono::{DateTime, Utc};
+
     let event_time: SystemTime = log_record
         .timestamp()
         .or(log_record.observed_timestamp())
         .unwrap_or_else(SystemTime::now);
 
-    let timestamp = win_filetime_from_systemtime!(event_time);
+    let timestamp: DateTime<Utc> = event_time.into();
     // TODO: time should be string
```
cijothomas, line 96: the TODO can be now removed.

opentelemetry-etw-logs/Cargo.toml
```diff
@@ -11,13 +11,13 @@ keywords = ["opentelemetry", "log", "trace", "etw"]
...
-tracelogging = "1.2.1"
 tracelogging_dynamic = "1.2.1"
 opentelemetry = { workspace = true, features = ["logs"] }
 opentelemetry_sdk = { workspace = true, features = ["logs"] }
-# TODO: Make json optional
-serde_json = "1.0.113"
+tracing = { version = "0.1", optional = true }
+serde_json = { version = "1.0.113", optional = true }
 futures-executor = "0.3.31"
+chrono = "0.4.40"
```
cijothomas, line 20: lets hand-pick the needed feature only from chrono crate.

opentelemetry-etw-logs/src/logs/reentrant_logprocessor.rs
```diff
@@ -7,6 +7,19 @@ use opentelemetry_sdk::Resource;
 
 use crate::logs::exporter::*;
 
+pub fn validate_provider_name(provider_name: &str) -> Result<(), String> {
+    if provider_name.len() >= 234 {
+        return Err("Provider name must be less than 234 characters.".to_string());
+    }
+    if !provider_name
```
utpilla, line 14: Update the check to not allow an empty string as well?

opentelemetry-etw-logs/src/logs/exporter/common.rs
lalitb: Good point. As a future enhancement - we can also consider a simple recursive serializer method to preserve `ListAny` and `Map` structures when serde is disabled.

psandana: done

opentelemetry-etw-logs/Cargo.toml
psandana: done

opentelemetry-etw-logs/src/logs/reentrant_logprocessor.rs
psandana: done

opentelemetry-etw-logs/src/logs/exporter/part_a.rs
psandana: done

opentelemetry-etw-logs/Cargo.toml
psandana: done

opentelemetry-etw-logs/Cargo.toml
```diff
@@ -36,7 +36,13 @@ spec_unstable_logs_enabled = [
     "opentelemetry/spec_unstable_logs_enabled",
     "opentelemetry_sdk/spec_unstable_logs_enabled",
 ]
-default = []
+internal-logs = [
+    "tracing",
+    "opentelemetry/internal-logs",
+    "opentelemetry_sdk/internal-logs",
+]
+serde_json = ["dep:serde_json"]
```
cijothomas, line 44: looks like we use "-" instead of "_" for feature flags everywhere. Lets stick with that for consistency?

opentelemetry-etw-logs/Cargo.toml
```diff
@@ -11,13 +11,13 @@ keywords = ["opentelemetry", "log", "trace", "etw"]
...
-tracelogging = "1.2.1"
 tracelogging_dynamic = "1.2.1"
 opentelemetry = { workspace = true, features = ["logs"] }
 opentelemetry_sdk = { workspace = true, features = ["logs"] }
-# TODO: Make json optional
-serde_json = "1.0.113"
+tracing = { version = "0.1", optional = true }
+serde_json = { version = "1.0.113", optional = true }
 futures-executor = "0.3.31"
+chrono = { version = "0.4.40", features = ["alloc"] }
```
cijothomas, line 20: disable default features too.

opentelemetry-etw-logs/Cargo.toml
cijothomas: After checking more, "_" is used everywhere! "internal-logs" is the exception by using "-" !

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -26,6 +26,9 @@ impl UserEventsExporter {
     /// Create instance of the exporter
     pub(crate) fn new(provider_name: &str) -> Result<Self, String> {
         // Validate provider_name
+        if provider_name.is_empty() {
```
cijothomas, line 29: please add this to changelog too. I also suggest to do separate PRs to keep scope of PR short and focused.

cijothomas (approved):
LGTM, left couple of comments to be addressed before merge.

cijothomas:
Lint failures are unrelated, will address it separately.

opentelemetry-etw-logs/src/logs/exporter/common.rs
utpilla, line 26: @psandana @cijothomas I reviewed the current JSON serialization logic and noticed that we don't restrict the recursion depth when serializing complex value types like ListAny and Map. This could result in performance issues and potentially cause a StackOverflow if a user-provided log with a deeply nested structure is processed by the exporter.

I think we should our custom serialization logic to enforce a maximum depth limit. I don't think `serde` offers any such configuration for serialization.

cijothomas: We'd be having same issue in OTLP Exporters too, so need to address it there too.

utpilla: One way to fix would be #238

PR #236, opened 2025-04-03
chore: Exclude autogenerated file from lint checks

cijothomas:
Excluding Stackdriver autogenerated file, as new rustc will fail CI


gruebel (approved):

opentelemetry-stackdriver/src/proto/api.rs
```diff
@@ -1,4 +1,5 @@
 // This file is @generated by prost-build.
+#![allow(clippy::all)]
```
lalitb, line 2: This should get added using generation script right?

gruebel: ok, I figured it out 😅 you have to add the clippy allow annotation here `#[allow(clippy::doc_overindented_list_items)]` (you could also add `#[allow(clippy::all)]`, but let's try not to do that for now)
https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/712daebd2ce3f2ac156dbb231ee98e6dd9a26270/opentelemetry-stackdriver/src/lib.rs#L42

gruebel: @lalitb I'm curious how would we fix it in the generation process? is there a repo or script for it somewhere?

lalitb: @gruebel - Not exactly. Your suggestion above it the right approach :)

PR #238, opened 2025-04-03
fix: [exporter-etw] Enforce depth limit when JSON serializing complex types

utpilla:
Addresses https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/234#discussion_r2027551351

## Changes
- Enforce a maximum depth limit of `1` when serializing `AnyValue::ListAny` and `AnyValue::HashMap`
- This means that we would correctly serialize a `vec` or a `HashMap` containing primitive types
- If a `vec` or a `HashMap` contains yet another `vec` or a `HashMap` we would serialize an error message string instead of parsing it futher


cijothomas:
@psandana Could you review? This seems safer

opentelemetry-etw-logs/src/logs/converters.rs
```diff
@@ -9,33 +9,53 @@ pub(super) trait IntoJson {
...
 }
 
-impl IntoJson for HashMap<Key, AnyValue> {
-    fn as_json_value(&self) -> Value {
-        Value::Object(
-            self.iter()
-                .map(|(k, v)| (k.to_string(), v.as_json_value()))
-                .collect::<Map<String, Value>>(),
-        )
+const ERROR_MSG: &str = "Maximum allowed nesting depth of `1` exceeded";
```
lalitb, line 16: nit - should we return more structured error message. Eg
```rust
 json!({ "truncated": true, "type": "list/map", reason: " max depth reached"})
```

psandana: First. I must mention that I'm not a fan of silent failures. I would rather the user gets a panic right away, than unexpectedly finding truncated content in production logs. But I recognize that a proper fix may not come easily, thus, let's move forward.

I `+1` @lalitb comment, and we should actually mention this is not an error, is a truncation. I would add the truncation depth as another value. In the future, we could allow users to modify this number under their own risk.

```rust
 json!({ "truncated": true, "type": "list/map", reason: "max depth reached", reason_depth: "1"})
 ```

Comment (utpilla) [reply opentelemetry-etw-logs/src/logs/converters.rs]: > First. I must mention that I'm not a fan of silent failures. I would rather the user gets a panic right away, than unexpectedly finding truncated content in production logs.

I would have to disagree here. Do we really want a log statement to be capable of bringing down the entire application and affect the user's SLA and business? In my opinion, it's better that we give the user a chance to fix their telemetry without crashing their application.

> we should actually mention this is not an error, is a truncation. I would add the truncation depth as another value. In the future, we could allow users to modify this number under their own risk.

This is what the message looks like now: `Message truncated as nested lists/maps are not supported. ` I think this is much simpler foe the user to understand than mentioning `depth = 1` etc. We can always include depth information later, if in future we decide to support nested structures. But for now, I favor the simplicity of the current message.

Comment (cijothomas) [reply opentelemetry-etw-logs/src/logs/converters.rs]: +1, we can't panic. OTel spec has been very clear on this from day 1:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md

Comment (psandana) [reply opentelemetry-etw-logs/src/logs/converters.rs]: >  Do we really want a log statement to be capable of bringing down the entire application and affect the user's SLA and business?

Personally, yes, I would like to fail early. Any piece of code should be covered already by tests, so I would get the panic early anyway. But I just agree is not the way that OTel specification is, so granted :)

Comment (psandana) [reply opentelemetry-etw-logs/src/logs/converters.rs]: >to be secure by default
I've seen two meanings for the statement above: fail early, or fallback safely. I rather to fail than to produce a wrong and unexpected side effect.

With this in mind, I was thinking that the correct approach may be to:
1. Issue a warning that `depth > n` has been reached, and application may crash, with `n` default to `1`.
2. Enable a feature to truncate the depth logged if a user-defined depth is trespassed
2. Just let the app crash if SO happen. Warnings have been given, and user must configure the library correctly as in previous point.

I think logs nesting dynamically just won't happen. Mostly logs structure will be statically defined. Thus, unexpected SO should be really infrequent.

File: opentelemetry-etw-logs/src/logs/converters.rs [hunk 2]
```diff
@@ -9,33 +9,55 @@ pub(super) trait IntoJson {
...
 }
 
-impl IntoJson for HashMap<Key, AnyValue> {
-    fn as_json_value(&self) -> Value {
-        Value::Object(
-            self.iter()
-                .map(|(k, v)| (k.to_string(), v.as_json_value()))
-                .collect::<Map<String, Value>>(),
-        )
+const ERROR_MSG: &str = "Message truncated as nested lists/maps are not supported.";
```
cijothomas, line 16:
```suggestion
const ERROR_MSG: &str = "Value truncated as nested lists/maps are not supported.";
```


"Message" feels a bit specific as truncation could be in Body/Message or Attributes. is "Value" better? 
Or just "truncated as nested ..."

opentelemetry-etw-logs/src/logs/converters.rs
cijothomas: @psandana Can you give more specifics? its not clear to me how this would be applied.
For example, can you clarify where to "issue a warning" ?

> Issue a warning that depth > n has been reached, and application may crash, with n default to 1.


Please clarify this too - are you referring to some feature-flags to control depth?

> Enable a feature to truncate the depth logged if a user-defined depth is trespassed

cijothomas: > Just let the app crash if SO happen. Warnings have been given, and user must configure the library correctly as in previous point.

We don't want the app to crash due to telemetry issues. We could make a future feature-request to enable such capability if a user explicitly opts-in. (I think this would be more like a general Otel wide setting, not just in an exporter)

opentelemetry-etw-logs/src/logs/converters.rs
utpilla: Update the message to use "Value".

opentelemetry-etw-logs/src/logs/converters.rs
psandana: >For example, can you clarify where to "issue a warning" ?

Throw/print a warning.

> Please clarify this too - are you referring to some feature-flags to control depth?
> > Enable a feature to truncate the depth logged if a user-defined depth is trespassed

Yes, making it configurable the truncation via feature-flags (features defined in cargo.toml).

cijothomas: > Please clarify this too - are you referring to some feature-flags to control depth?
> 
> Enable a feature to truncate the depth logged if a user-defined depth is trespassed
> Yes, making it configurable the truncation via feature-flags (features defined in cargo.toml).

Yes makes sense. Though we need to decide between feature flags vs something in the Options struct. Please open a separate issue and we can discuss specifics there.

cijothomas: > For example, can you clarify where to "issue a warning" ?

> Throw/print a warning.

We can't throw (panic). No good way to print warning in hot path either -
https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-user-events-logs/src/logs/reentrant_logprocessor.rs#L29-L35 

We can discuss this in a separate issue. There could be other places where we want internal logs also.

PR #246, opened 2025-04-10
refactor: remove Prometheus related dependencies from actix instrumentation

gruebel:
## Changes

- removed Prometheus middleware in favour of leveraging OTLP with Prometheus
- I didn't plan on it, but now the crate only depends on `opentelemetry` + `opentelemetry-semantic-conventions` from the OTel family

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ debug = 1
 opentelemetry = "0.29"
 opentelemetry-appender-tracing = "0.29"
 opentelemetry-http = "0.29"
+opentelemetry-otlp = "0.29"
 opentelemetry-proto = { version = "0.29", default-features = false }
 opentelemetry_sdk = { version = "0.29", default-features = false }
 opentelemetry-stdout = "0.29"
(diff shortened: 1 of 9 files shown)


cijothomas:
> I didn't plan on it, but now the crate only depends on opentelemetry + opentelemetry-semantic-conventions from the OTel family

💯 those are the only expected dependencies!

Cargo.toml
```diff
@@ -31,6 +31,7 @@ debug = 1
 opentelemetry = "0.29"
 opentelemetry-appender-tracing = "0.29"
 opentelemetry-http = "0.29"
+opentelemetry-otlp = "0.29"
```
cijothomas, line 34: do we have to show OTLP exporter in the examples? I think stdout is good enough. Changing from stdout to otlp would be straightforward, and anyway not connected to the instrumentation itself.

cijothomas (approved):
Awesome! I left a comment about using stdout instead of otlp in the examples, as instrumentations are not concerned with exporters anyway, so we can keep examples the simplest.
(OTLP Exporter requires some more thinking, and we can avoid duplicating those information in every example by sticking to stdout exporter in every example)

Cargo.toml
gruebel: no problem, switched the examples to stdout and removed the dependency

PR #247, opened 2025-04-11
feat: Use "Log" as default event name

psandana:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-etw-logs/src/logs/exporter/common.rs
+++ b/opentelemetry-etw-logs/src/logs/exporter/common.rs
@@ -76,7 +76,7 @@ pub const fn convert_severity_to_level(severity: Severity) -> tld::Level {
 }
 
 pub fn get_event_name(log_record: &opentelemetry_sdk::logs::SdkLogRecord) -> &str {
-    log_record.event_name().unwrap_or("Log")
+    "Log"
 }
 
 #[cfg(test)]
--- a/opentelemetry-user-events-logs/src/logs/exporter.rs
+++ b/opentelemetry-user-events-logs/src/logs/exporter.rs
@@ -191,13 +191,9 @@ impl UserEventsExporter {
                 // reused for events with different schema. 
                 // In well-behaved application, event-name should be unique
                 // for each event.
-                // TODO: What if the event name is not provided? "Log" is used as default.
                 // TODO: Should event_tag be non-zero?
-                let event_name = log_record
-                    .event_name()
-                    .filter(|s| !s.trim().is_empty())
-                    .unwrap_or("Log");                
-                eb.reset(event_name, 0);
+                // "Log" is used as default.
+                eb.reset("Log", 0);
 
                 eb.add_value("__csver__", 1024, FieldFormat::UnsignedInt, 0); // 0x400 in hex
 
@@ -326,6 +322,10 @@ impl UserEventsExporter {
                 // TODO: eventname is already added to header.
                 // Should we add it again?
                 // Or should we use Target?
+                let event_name = log_record
+                    .event_name()
+                    .filter(|s| !s.trim().is_empty())
+                    .unwrap_or("Log");
                 eb.add_str("name", event_name, FieldFormat::Default, 0);
                 cs_b_count += 1;
                 eb.set_struct_field_count(cs_b_bookmark, cs_b_count);


opentelemetry-etw-logs/src/logs/exporter_options.rs
```diff
@@ -0,0 +1,262 @@
...
+pub struct ExporterOptions {
+    provider_name: Cow<'static, str>,
+    event_mapping: Option<EventMapping>,
+    on_missing_key_use_value: bool,
+    default_event_name: Cow<'static, str>,
+}
+
+impl ExporterOptions {
+    /// TODO: Add documentation
+    pub fn builder(provider_name: Cow<'static, str>) -> ExporterOptionsBuilder {
```
martintmk, line 24:
```suggestion
    pub fn builder(provider_name: impl Into<Cow<'static, str>>) -> ExporterOptionsBuilder {
```

^^ allows passing strings directly

opentelemetry-etw-logs/src/logs/exporter_options.rs
```diff
@@ -0,0 +1,265 @@
...
+}
+
+impl ExporterOptions {
+    /// TODO: Add documentation
+    pub fn builder(provider_name: impl Into<Cow<'static, str>>) -> ExporterOptionsBuilder {
+        ExporterOptionsBuilder::new(provider_name)
+    }
+
+    /// TODO: Add documentation
+    pub fn provider_name(&self) -> &str {
```
psandana, line 29:
```suggestion
    pub(crate) fn provider_name(&self) -> &str {
```

opentelemetry-etw-logs/src/logs/exporter_options.rs
```diff
@@ -0,0 +1,265 @@
...
+        &self.provider_name
+    }
+
+    /// TODO: Add documentation
+    pub fn event_mapping(&self) -> Option<&EventMapping> {
+        self.event_mapping.as_ref()
+    }
+
+    /// TODO: Add documentation
+    pub fn default_event_name(&self) -> &str {
```
psandana, line 39:
```suggestion
    pub(crate) fn default_event_name(&self) -> &str {
```

opentelemetry-etw-logs/src/logs/exporter_options.rs
```diff
@@ -0,0 +1,265 @@
...
+        self.default_event_name.as_ref()
+    }
+
+    /// TODO: Add documentation
+    pub fn get_etw_event_name<'a>(
```
psandana, line 44:
```suggestion
    pub(crate) fn get_etw_event_name<'a>(
```

opentelemetry-etw-logs/src/logs/exporter_options.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /// TODO: Add documentation
+    pub(crate) fn get_etw_event_name<'a>(
+        &'a self,
+        log_record: &'a opentelemetry_sdk::logs::SdkLogRecord,
+    ) -> &'a str {
+        // Using target for now. This is the default behavior.
+        // Future versions of this library may add mechanisms to chose which attribute to use for the mapping key
+        if let Some(target) = log_record.target() {
```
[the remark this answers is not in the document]
psandana: I propose a default behavior of using `target` if exists, and default to `Log` or any other default set by user. `log` and `tracing` crates uses `target` with defaults to contains the `crate` (and possibly the module) originating a log record. This can be used to write to the desired ETW Event Name.

While hardcoding `Logs` can work for basic scenarios, most users of ETW will expect to configure their own Event Names.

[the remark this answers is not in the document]
psandana: After discussing this further, we are providing:
- `Log` as default value for ETW event name
- the `Log` default value can be overridden by calling: `with_default_event_name()`
- if user wants to use the `SdkLogRecord` name, it can select: `use_etw_event_name_from_name()`
- if user wants to use the `SdkLogRecord` target, it can select: `use_etw_event_name_from_target()`

opentelemetry-etw-logs/src/logs/reentrant_logprocessor.rs
```diff
@@ -6,40 +6,29 @@ use opentelemetry_sdk::logs::{LogBatch, LogExporter, SdkLogRecord};
...
+    /// Creates a new instance of the ReentrantLogProcessor using the given options.
+    pub(crate) fn new(options: ExporterOptions) -> Self {
+        let exporter: ETWExporter = ETWExporter::new(options);
         ReentrantLogProcessor {
             event_exporter: exporter,
         }
     }
 }
 
+/// Creates an opaque LogProcessor that can be used with the OpenTelemetry SDK.
```
cijothomas, line 27: given this returns a *new* instance of LogRecordProcessor, *not* the one actually used if user did `with_etw_exporter` option to add etw_exporter, this can make it easy for users to make mistakes!
(it won't be obvious that the processor returned in this method is not the same as what is used by the trait method!)

opentelemetry-etw-logs/src/logs/exporter_options.rs
```diff
@@ -0,0 +1,356 @@
...
+        ExporterOptionsBuilder {
+            inner: ExporterOptions {
+                provider_name: provider_name.into(),
+                default_event_name: "Log".into(),
+                event_name_from: ETWEventNameFrom::Default,
+            },
+        }
+    }
+
+    /// Sets the default event name to use by `ExporterOptions::get_etw_event_name` if:
```
cijothomas, line 76: `ExporterOptions::get_etw_event_name` is an internal method right? Lets not mention about it public docs.

opentelemetry-etw-logs/src/logs/exporter_options.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// - it cannot extract name or target from the `SdkLogRecord`.
+    pub fn with_default_event_name(
+        mut self,
+        default_event_name: impl Into<Cow<'static, str>>,
+    ) -> Self {
+        self.inner.default_event_name = default_event_name.into();
+        self
+    }
+
+    /// Sets the provider name to be the default value ("Log"). The default value may be overridden by `with_default_event_name()`.
```
cijothomas, line 87: likely a typo - you meant etw_event name, not the provider name right?

opentelemetry-etw-logs/src/logs/exporter_options.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    /// Sets the provider name to be the `name` from `SdkLogRecord`. If `name` is `None`, it uses the default value ("Log"). The default value may be overridden by `with_default_event_name()`.
+    pub fn use_etw_event_name_from_name(mut self) -> Self {
+        self.inner.event_name_from = ETWEventNameFrom::Name;
+        self
+    }
+
+    /// Validates the options given by consuming itself and returning the `ExporterOptions` or an error.
+    pub fn build(self) -> Result<ExporterOptions, String> {
```
cijothomas, line 106: Is it better to use a dedicated Error enum, instead of String, which the user has to parse out to understand what went wrong? as of today, only provider name can cause failure, but in the future there could be other build errors we want to surface back to users?

OTLP Example:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/mod.rs#L107

opentelemetry-etw-logs/src/logs/exporter_options.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub fn with_default_event_name(
+        mut self,
+        default_event_name: impl Into<Cow<'static, str>>,
+    ) -> Self {
+        self.inner.default_event_name = default_event_name.into();
+        self
+    }
+
+    /// Sets the provider name to be the default value ("Log"). The default value may be overridden by `with_default_event_name()`.
+    pub fn use_etw_event_name_from_default(mut self) -> Self {
```
cijothomas, line 88: isn't this the default anyway, so does the user has to set default as default explicitly?

opentelemetry-etw-logs/Cargo.toml
```diff
@@ -48,5 +50,9 @@ default = ["internal-logs"]
 name = "basic"
 path = "examples/basic.rs"
 
+[[example]]
```
cijothomas, line 53: not so sure if its worth adding another example, just for exporter_options, which is anyway basically covered in the basic.rs too?

opentelemetry-etw-logs/src/logs/with_etw_exporter.rs
```diff
@@ -60,4 +58,28 @@ mod tests {
             );
         });
     }
+
+    #[test]
+    fn log_with_etw_exporter_trait() {
+        use log::{error, Level};
```
cijothomas, line 64: why test `log` here specifically? Exporters should have no connection to the logging library anyway right?

opentelemetry-etw-logs/src/logs/exporter_options.rs
```diff
@@ -0,0 +1,356 @@
...
+#[derive(Debug)]
+enum ETWEventNameFrom {
+    Default,
+    Target,
+    Name,
+}
+
+/// Options used by the Exporter.
+#[derive(Debug)]
+pub struct ExporterOptions {
```
cijothomas, line 13: @lalitb @utpilla 
Would love to get more feedback on this approach. 
One downside with this approach is that, users will most likely need to write LogRecordProcessor to customize Name,Target as their defaults may not work well. (i.e the default name, target provided by `tracing` is not likely useful)

Other option is to expose a callback for end-users, in the exporter.
`get_etw_event_name(&log_record) -> &str`
This has the upside that users can return the name without having to replace the original name (and target), so PartB.Log.name will always contain the original EventName, and users can influence PartA-name (and Table names in backend). This does have a risk of letting arbitrary user code run in the hot path. Given the alternative is to write LogRecordProcessor, which is also arbitrary user-code, we could reconsider (@psandana yes, I known I was not very supportive of providing arbitrary callbacks, but considering the full picture, I can change my position) this approach.

opentelemetry-etw-logs/src/logs/reentrant_logprocessor.rs
cijothomas: > It should definitely be clarified better in the rustdoc, but I'd hope that that's enough. Or do you have something else in mind as for how to address this?

Doc is the minimum we should do. Or better is to avoid 2 APIs, so there is no scope of misconfiguring things.

opentelemetry-etw-logs/src/logs/with_etw_exporter.rs
psandana: I agree, and for that we should use the `LogRecord` interface directly. But as we decided to not do so, and use `tracing`, I also feel we should explicitly test the integration of `log` crate for parity.

opentelemetry-etw-logs/src/logs/exporter_options.rs
psandana: Yes, it is the default already. No, use does not have to do it. It is left to document better the 3 possibilities. If we remove it, users may have the wrong assumption they need to choose from target or name.

I will improve docs around this

opentelemetry-etw-logs/Cargo.toml
psandana: basic is just the default. I will add more options to make it worth to have it.

opentelemetry-etw-logs/src/logs/with_etw_exporter.rs
cijothomas: please make those changes in separate PR to keep the scope in control. This is already touching too many aspects in one PR.

opentelemetry-etw-logs/src/logs/exporter_options.rs
psandana: Done

opentelemetry-etw-logs/Cargo.toml
psandana: Removed

opentelemetry-etw-logs/src/logs/processor.rs
```diff
@@ -0,0 +1,295 @@
...
+#[derive(Debug)]
+pub struct ProcessorBuilder {
+    options_builder: OptionsBuilder,
+}
+
+impl ProcessorBuilder {
+    /// Creates a new instance of `ProcessorBuilder` with the given provider name.
+    ///
+    /// By default, all events will be exported to the "Log" ETW event.
+    pub fn new(provider_name: impl Into<Cow<'static, str>>) -> Self {
```
cijothomas, line 85: can you check if this can be consistent with user_events exporter, which accepts `&str` for provider_name.

opentelemetry-etw-logs/src/logs/processor.rs
```diff
@@ -0,0 +1,295 @@
...
+        write!(f, "{}", self.kind)?;
+        if self.backtrace.status() == std::backtrace::BacktraceStatus::Captured {
+            write!(f, "\nBacktrace: {:?}", self.backtrace)?;
+        }
+        Ok(())
+    }
+}
+
+impl std::error::Error for ProcessorBuildError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
```
cijothomas, line 70: is this required to be implemented?

opentelemetry-etw-logs/src/logs/processor.rs
cijothomas, line 85: for consistency with user_events, lets not keep this public. Only way for users to get to this is via Processor::builder(..).

opentelemetry-etw-logs/src/logs/processor.rs
```diff
@@ -0,0 +1,295 @@
...
+use opentelemetry::InstrumentationScope;
+use opentelemetry_sdk::error::OTelSdkResult;
+use opentelemetry_sdk::logs::{LogBatch, LogExporter, SdkLogRecord};
+use opentelemetry_sdk::Resource;
+
+use crate::logs::exporter::*;
+
+/// Errors that can occur while building a `Processor` instance.
+#[derive(Debug)]
+pub struct ProcessorBuildError {
```
cijothomas, line 17: I think we should stick with existing patterns in the Otel repo. 
We can either do an Enum with all possible errors listed (and mark non_exhaustive).
Or
return `Box<dyn Error>`

user-events exporter followed the pattern of returning `Box<dyn Error>`

opentelemetry-etw-logs/src/logs/processor.rs
```diff
@@ -0,0 +1,295 @@
...
+#[derive(Error, Debug, PartialEq)]
+pub(crate) enum ProcessorBuildErrorKind {
+    #[error("Provider name cannot be empty.")]
+    ProviderNameEmpty,
+    #[error("Provider name must be less than 234 characters.")]
+    ProviderNameTooLong,
+    #[error("Provider name must contain only ASCII alphanumeric characters, '_' or '-'.")]
+    ProviderNameInvalid,
+    #[error("Event name cannot be empty.")]
+    EventNameEmpty,
```
cijothomas, line 56: this looks incorrect....Since event_name is not an input to the builder, builder cannot fail to build due to empty event name...

Also, this pub (crate) enum vs the pub struct is making bit hard to follow the overall flow... Unless there is a very strong reason, lets keep it simpler

opentelemetry-etw-logs/examples/basic.rs
```diff
@@ -39,14 +39,20 @@
...
-use opentelemetry_etw_logs::ETWLoggerProviderBuilderExt;
+use opentelemetry_etw_logs::Processor;
 use opentelemetry_sdk::logs::SdkLoggerProvider;
 use tracing::error;
 use tracing_subscriber::prelude::*;
 
 fn init_logger() -> SdkLoggerProvider {
+    let processor = Processor::builder("provider-name")
+        // Use a constant custom name for all events instead of "Log"
+        .etw_event_name_from_callback(|_| "CustomEvent")
```
cijothomas, line 50: nit: basic/simple example need not show customization.  suggest to use the simplest.

opentelemetry-etw-logs/src/logs/exporter/options.rs
```diff
@@ -0,0 +1,256 @@
...
+impl<F> EventNameCallback for F where F: Fn(&SdkLogRecord) -> &str + Send + Sync + 'static {}
+
+impl std::fmt::Debug for dyn EventNameCallback {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("ETW event name callback")
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct OptionsBuilder {
```
cijothomas, line 56: its easier to review, if builder is handled completed inside the ProcessorBuilder itself.

Cargo.toml
```diff
@@ -29,6 +29,7 @@ debug = 1
 
 [workspace.dependencies]
 opentelemetry = "0.29"
+opentelemetry-appender-log = "0.29"
```
cijothomas, line 32: please make these changes in separate PRs. Big PRs are extremely time consuming to review. Address one issue/concern at a time.

cijothomas (changes requested):
Thanks! The direction looks good, here are few things we need to reconsider 
1. Please remove all changes unrelated to changing from "with_etw" to the processor builder pattern. Large PRs gets harder to review, especially when new commits are pushed.
2. Unsure if we should use a totally different approach of returning Errors here - I suggest to stick with returning Box<dyn Err> for now.
3. user-events did recent refactor to support the new processor builder, see if we can make it consistent with that. I think we should simplify code bit more to make it easier to follow, left some comments about this in couple of places.

opentelemetry-etw-logs/src/logs/processor.rs
psandana: Done

opentelemetry-etw-logs/src/logs/processor.rs
```diff
@@ -0,0 +1,295 @@
...
+        self.options_builder = self.options_builder.etw_event_name_from_callback(callback);
+        self
+    }
+
+    /// Validates the options given by consuming itself and returning the `Processor` or an error.
+    pub fn build(self) -> Result<Processor, ProcessorBuildError> {
+        match self.options_builder.build() {
+            Ok(options) => Ok(Processor::new(options)),
+            Err(error) => {
+                otel_warn!(name: "ETW.Processor.CreationFailed", reason = &error.to_string());
```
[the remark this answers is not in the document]
psandana: removed

opentelemetry-etw-logs/examples/basic.rs
psandana: removed

opentelemetry-etw-logs/src/logs/exporter/mod.rs
```diff
@@ -233,6 +240,21 @@ mod tests {
...
+        use opentelemetry_sdk::logs::LogBatch;
+        use opentelemetry_sdk::logs::LogExporter;
+
+        let records = [];
+        let batch = LogBatch::new(&records);
+
+        let exporter = common::test_utils::new_etw_exporter();
+        let result = exporter.export(batch);
+
+        assert!(result.await.is_ok());
```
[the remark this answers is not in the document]
psandana: Good call. Fixed

Cargo.toml
psandana: This was dangling dep. Removing

opentelemetry-etw-logs/CHANGELOG.md
```diff
@@ -2,8 +2,18 @@
 
 ## vNext
 
-- Added validation to provider name
+- Added validation to provider name.
 - Added optional feature `serde_json` to serialize List and Maps.
+- Added `etw_log_processor()` method that returns an `impl LogProcessor`.
```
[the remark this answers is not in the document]
psandana: Done

psandana:
> Thanks! The direction looks good, here are few things we need to reconsider
> 
> 1. Please remove all changes unrelated to changing from "with_etw" to the processor builder pattern. Large PRs gets harder to review, especially when new commits are pushed.
> 2. Unsure if we should use a totally different approach of returning Errors here - I suggest to stick with returning Box for now.
> 3. user-events did recent refactor to support the new processor builder, see if we can make it consistent with that. I think we should simplify code bit more to make it easier to follow, left some comments about this in couple of places.

Hi @cijothomas . I reworked the PR using the feedback provided. Errors are Boxed and simplified things. Please review.

opentelemetry-etw-logs/Cargo.toml
```diff
@@ -11,15 +11,19 @@ keywords = ["opentelemetry", "log", "trace", "etw"]
...
 opentelemetry_sdk = { workspace = true, features = ["logs"] }
-tracing = { version = "0.1", optional = true }
 serde_json = { version = "1.0.113", optional = true }
-futures-executor = "0.3.31"
-chrono = { version = "0.4.40", default-features = false, features = ["alloc", "std"] }
+tracelogging_dynamic = "1.2.4"
+tracing = { version = "0.1", optional = true }
 
 [dev-dependencies]
+log = { version = "0.4.21", features = ["kv_serde"] }
```
cijothomas, line 26: is this required?

opentelemetry-etw-logs/Cargo.toml
```diff
@@ -11,15 +11,19 @@ keywords = ["opentelemetry", "log", "trace", "etw"]
 license = "Apache-2.0"
 
 [dependencies]
-tracelogging_dynamic = "1.2.4"
+chrono = { version = "0.4.40", default-features = false, features = [
```
cijothomas, line 14: its hard to tell if this file actually has changes or not..looks like just re-arranging ? please remove unrelated changes from this PR.

opentelemetry-etw-logs/Cargo.toml
psandana: Not required anymore, along with microbench. Fixed

opentelemetry-etw-logs/Cargo.toml
psandana: I won't trim this PR further. Please, accept this PR as is. Subsequent ones should be small, incremental changes.

opentelemetry-etw-logs/src/logs/processor.rs
```diff
@@ -0,0 +1,302 @@
...
+        ProcessorBuilder {
+            options_builder: Options::new(provider_name.to_string()),
+        }
+    }
+
+    /// Sets a user-defined callback that returns the ETW event name, using the the [`SdkLogRecord`] as input.
+    ///
+    /// The resulting name must be a valid CommonSchema 4.0 TraceLoggingDynamic event name. Otherwise,
+    /// the default "Log" ETW event name will be used.
+    pub fn etw_event_name_from_callback(
```
cijothomas, line 104: Please see the discussion here : https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/275#discussion_r2103713620

We need to decide how best to handle this.

opentelemetry-etw-logs/Cargo.toml
cijothomas: Sorry I don't follow. Does this file actually have any changes or just re-arranging the order?
Its putting unnecessary burden on reviewers to closely check and confirm if this is actually changed or not.

opentelemetry-etw-logs/src/logs/processor.rs
```diff
@@ -0,0 +1,302 @@
...
+        mut self,
+        callback: impl Fn(&SdkLogRecord) -> &str + Send + Sync + 'static,
+    ) -> Self {
+        self.options_builder = self.options_builder.etw_event_name_from_callback(callback);
+        self
+    }
+
+    /// Validates the options given by consuming itself and returning the `Processor` or an error.
```
cijothomas, line 112: nit: The comment should be something like "Builds the processor with given options, returning Err when build fails (eg: due to invalid options)".

opentelemetry-etw-logs/src/logs/processor.rs
```diff
@@ -0,0 +1,302 @@
...
+    pub fn build(self) -> Result<Processor, Box<dyn Error>> {
+        self.validate()?;
+
+        Ok(Processor::new(self.options_builder))
```
cijothomas, line 116: looks a bit odd to see that a builder is still passed to Processor...is the processor expected to call build() on the options_builder()?

opentelemetry-etw-logs/src/logs/exporter/options.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn test_get_event_name_from_callback_returning_name() {
+        use opentelemetry::logs::LogRecord;
+
+        let mut log_record = test_utils::new_sdk_log_record();
+
+        let options = Options::new("test-provider-name")
+            .etw_event_name_from_callback(|log_record| log_record.event_name().unwrap_or_default());
+
+        let result = options.get_etw_event_name(&log_record);
+        assert_eq!(result, "Log");
```
cijothomas, line 128: interesting that "Log" is still the name, though the callback is always returning something... Is this trying to test that `invalid_event_name_from_callback_will_use_default` ?

opentelemetry-etw-logs/src/logs/exporter/options.rs
```diff
@@ -0,0 +1,169 @@
...
+            });
+
+        let result = options.get_etw_event_name(&log_record);
+        assert_eq!(result, "Log");
+
+        log_record.set_event_name("event-name");
+        let result = options.get_etw_event_name(&log_record);
+        assert_eq!(result, "Log");
+
+        log_record.set_target("target-name");
```
cijothomas, line 161: what happens if we pass empty string to target?

opentelemetry-etw-logs/src/logs/processor.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+impl Processor {
+    /// Creates a new instance of [`ProcessorBuilder`] with the given provider name.
+    ///
+    /// The provider name must be:
+    ///
+    /// - non-empty,
+    /// - less than 234 characters long, and
+    /// - contain only ASCII alphanumeric characters, underscores (`_`), or hyphens (`-`).
```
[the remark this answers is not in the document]
psandana: If we follow the stricter user-events validation, then we will not be able to match all the possible valid providers in ETW. There are system ETW providers that uses hyphens already (unofficial list: https://gist.github.com/guitarrapc/35a94b908bad677a7310)

opentelemetry-etw-logs/src/logs/processor.rs
psandana: `options_builder` should be just `options`. The change was missing from last refactor. Fixed.

opentelemetry-etw-logs/src/logs/processor.rs
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    #[test]
+    fn tracing_test() {
+        use opentelemetry_appender_tracing::layer;
+        use tracing::error;
+        use tracing_subscriber::prelude::*;
+
+        let processor = Processor::builder("provider-name")
+            .etw_event_name_from_callback(|_| "CustomEvent")
```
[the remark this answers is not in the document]
psandana: It is an integration test at this moment. Renamed to `tracing_integration_test`. We can move it to the `tests` folder (we should create it) if you prefer, either on this PR or a new one.

opentelemetry-etw-logs/src/logs/exporter/options.rs
psandana: Depends on the callback given. Assuming the callback is returning the target, the current validation will discard it because it is empty and return the default "Log".

opentelemetry-etw-logs/src/logs/exporter/part_c.rs
```diff
@@ -1,9 +1,9 @@
 use opentelemetry::logs::AnyValue;
 use tracelogging_dynamic as tld;
 
-pub const EVENT_ID: &str = "event_id";
+pub(crate) const EVENT_ID: &str = "event_id";
 
-pub fn populate_part_c(
+pub(crate) fn populate_part_c(
```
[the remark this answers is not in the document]
psandana: This is just more intentional and resilient to refactorings.

opentelemetry-etw-logs/README.md
```diff
@@ -28,7 +28,51 @@ captured by agents running locally and listening for specific ETW events.
...
+
+Instructions using Powershell:
+
+1. Get the ETW Session Guid for the given provider (on this example `provider-name`):
+
+    ```ps
+    $EtwSessionGuid = (new-object System.Diagnostics.Tracing.EventSource("provider-name")).Guid.ToString()`
+    ```
+
+1. Start Logman session:
```
utpilla, line 47: Correct the numbering used for these steps. All the steps are numbered as "1".

psandana: Yes. This is correct. MD solves them at rendering time. https://www.markdownguide.org/basic-syntax/#ordered-lists

opentelemetry-etw-logs/Cargo.toml
utpilla: > Its putting unnecessary burden on reviewers to closely check and confirm if this is actually changed or not.

I’m with @cijothomas on this. Simplifying the PR for review not only helps reviewers but also benefits the author by increasing the chances of a quicker merge. In this case, the only intended change appears to be the removal of `microbench` from the dev dependencies. Ideally, that should have been the sole modification, without reordering the other dependencies.

> The project does not have a toml formatting, causing this kind of inconsistencies.

It’s usually a good idea to stick to the existing style, formatting, or conventions used in the project when sending in PRs. If there doesn’t seem to be a clear standard yet, it might be better to open a separate PR to introduce one, rather than mixing style changes into an unrelated change.

opentelemetry-etw-logs/README.md
utpilla: Ah okay

opentelemetry-etw-logs/src/logs/processor.rs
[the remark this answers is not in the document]
cijothomas: My comment was more about - "There are not asserts in the test, so what is the purpose of the test"?

opentelemetry-etw-logs/src/logs/exporter/options.rs
cijothomas: Got it. It's better to have a targeted test for the same. `invalid_event_name_from_callback_will_use_default` would be my suggestion.

Not a blocker.

PR #254, opened 2025-05-07
feat: etw_log_processor()

psandana:
Fixes #

## Changes

Adding `etw_log_processor()` method that returns an `impl LogProcessor`. Useful for advanced scenarios that requires access to the underlaying `LogProcessor`.

### DIFF
--- a/opentelemetry-etw-logs/src/logs/reentrant_logprocessor.rs
+++ b/opentelemetry-etw-logs/src/logs/reentrant_logprocessor.rs
@@ -40,6 +40,13 @@ impl ReentrantLogProcessor {
     }
 }
 
+/// Creates an opaque LogProcessor that can be used with the OpenTelemetry SDK.
+pub fn etw_log_processor(
+    provider_name: &str,
+) -> impl opentelemetry_sdk::logs::LogProcessor + use<> {
+    ReentrantLogProcessor::new(provider_name)
+}
+
 impl opentelemetry_sdk::logs::LogProcessor for ReentrantLogProcessor {
     fn emit(&self, data: &mut SdkLogRecord, instrumentation: &InstrumentationScope) {
         let log_tuple = &[(data as &SdkLogRecord, instrumentation)];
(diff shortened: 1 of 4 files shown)


opentelemetry-etw-logs/src/logs/reentrant_logprocessor.rs
```diff
@@ -40,6 +40,13 @@ impl ReentrantLogProcessor {
     }
 }
 
+/// Creates an opaque LogProcessor that can be used with the OpenTelemetry SDK.
+pub fn etw_log_processor(
```
cijothomas, line 44: We need to remove the `with_etw_exporter, if we have this method to get the processor - else users might end up using `with_etw_exporter` *and* `etw_log_processor` together, which will not work.

Suggestion:

```rust
let etw_options = EtwOptions::builder()
       .with_provider_name("provider")
       .build();
let etw_processor = EtwExporter::build_processor(etw_options);
let logger_provider = LoggerProvider::builder()
       .with_processor(etw_processor)
       .build();
 ```
This also allows decorating/wrapping processor

psandana: I can do that change, but in the other PR #247 . To make work this PR on 1.75 would be extra thow-away work.

psandana: @cijothomas where is the `.with_processor()`? I canot find any references to it.

cijothomas: > I can do that change, but in the other PR #247 . To make work this PR on 1.75 would be extra throw-away work.

I am bit confused with multiple open PRs.. 247 looks quite big - please do shorter, focused PR to make reviews easier.

psandana: I know 247 is big (although I've seen bigger ones very often), but is hard to decouple at this moment.

PR #257, opened 2025-05-10
feat: [Geneva uploader] Add LZ4 chunked compression implementation 

lalitb:
## Changes
This PR introduces LZ4 chunked compression logic within the new `payload_encoder` directory of the `geneva-uploader` crate.

- The `payload_encoder` directory is created to house all payload encoding and compression logic.
- Currently, only the compression code (LZ4 chunked compression) is added and is kept internal to the crate.
- Encoding functionality will be added in a separate PR.


opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/mod.rs
```diff
@@ -0,0 +1,76 @@
+mod lz4_chunked_compression;
+
+#[cfg(test)]
+mod tests {
```
gruebel, line 4: code looks good, I'm just curious why you put the tests here and not in the `lz4_chunked_compression.rs` file?

lalitb: Thanks @gruebel for reviewing. You are correct, the tests logically belong to `lz4_chunked_compression.rs`. Have moved them there.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -0,0 +1,149 @@
...
+///   If single-buffer "in-place" compression is possible (e.g., with unsafe or buffer aliasing), document or implement it here.
+#[allow(dead_code)]
+pub(crate) fn lz4_chunked_compression(input: &[u8]) -> Vec<u8> {
+    const CHUNK_SIZE: usize = 64 * 1024;
+
+    // Pre-allocate an output buffer large enough for the worst-case total output size:
+    // Each chunk may require up to get_maximum_output_size(CHUNK_SIZE) bytes for compressed data,
+    // plus 4 bytes for the length header per chunk.
+    let mut output = Vec::with_capacity(
+        (input.len() / CHUNK_SIZE + 1) * (4 + get_maximum_output_size(CHUNK_SIZE)),
```
utpilla, line 45:
```suggestion
        input.len().div_ceil(CHUNK_SIZE) * (4 + get_maximum_output_size(CHUNK_SIZE)),
```

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -0,0 +1,149 @@
...
+/// ```text
+///  |<--Chunk 1-->|<--Chunk 2-->| ... |<--Chunk N-->|
+///  +----+--------+----+-------+----+-----+
+///  | 04 |....    | 03 |...    | 02 |..   |
+///  +----+--------+----+-------+----+-----+
+///
+///  Where:
+///    04, 03, 02  = 4-byte little-endian u32 chunk length headers
+///    ...., ..., .. = compressed LZ4 data for each chunk (length matches header)
+///    N     = number of chunks = (input.len() / CHUNK_SIZE) + (if input.len() % CHUNK_SIZE > 0 { 1 } else { 0 })
```
utpilla, line 19:
```suggestion
///    N     = number of chunks = input.len().div_ceil(CHUNK_SIZE)
```

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -0,0 +1,149 @@
...
+///   but care must be taken: LZ4 does not natively support in-place compression, and the compressed size may be larger than the input.
+///   If single-buffer "in-place" compression is possible (e.g., with unsafe or buffer aliasing), document or implement it here.
+#[allow(dead_code)]
+pub(crate) fn lz4_chunked_compression(input: &[u8]) -> Vec<u8> {
+    const CHUNK_SIZE: usize = 64 * 1024;
+
+    // Pre-allocate an output buffer large enough for the worst-case total output size:
+    // Each chunk may require up to get_maximum_output_size(CHUNK_SIZE) bytes for compressed data,
+    // plus 4 bytes for the length header per chunk.
+    let mut output = Vec::with_capacity(
```
utpilla, line 44: It might be possible to avoid allocating a vec on each operation on the request. If we enforce the creation/ reuse of vec in the upstream callers. I believe this would be the code flow:

export() -> encode() -> compress() -> upload()

If we push the responsibility of creating the vec at the `export()` level, we can avoid intermediate vec allocations in the subsequent methods such as `encode()` and `compress()`. These subsequent methods can simply accept a `&mut [u8]`  parameter in their method signature to which they can write data.

This might be something to check once we have these methods integrated.

utpilla, line 44: Also, we should set a limit for the maximum size that we would allocate a vec with. It shouldn't be unbounded.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -0,0 +1,149 @@
...
+        (input.len() / CHUNK_SIZE + 1) * (4 + get_maximum_output_size(CHUNK_SIZE)),
+    );
+
+    // Temporary buffer for compressing each chunk (reused for all chunks).
+    let mut temp = vec![0u8; get_maximum_output_size(CHUNK_SIZE)];
```
utpilla, line 49: We should be able to avoid this extra vec allocation. Since, we would always use 4 bytes for storing the length of any chunk, we could do this using just the one large vec. You just have to provide the right slice of output buffer when calling `compress_into()` and backfill the length of the chunk before you proceed to compress the next chunk.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -0,0 +1,149 @@
...
+    let mut offset = 0;
+    // Process the input in 64 KiB chunks.
+    while offset < input.len() {
+        // Determine the end index for the current chunk.
+        let end = usize::min(offset + CHUNK_SIZE, input.len());
+        // Get the current chunk from input.
+        let chunk = &input[offset..end];
+
+        // Compress the chunk into the temporary buffer.
+        let compressed_size = compress_into(chunk, &mut temp).expect("Compression failed");
```
utpilla, line 60: Use `Result` instead of calling `expect()`.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -0,0 +1,149 @@
...
+        output.extend_from_slice(&len_bytes);
+
+        // Append the compressed data itself.
+        output.extend_from_slice(&temp[..compressed_size]);
+
+        // Move to the next chunk.
+        offset = end;
+    }
+
+    output
```
utpilla, line 73: The output buffer might contain significantly more items than necessary. If this method's result is passed to `upload()`, it could lead to transmitting excess data over the network, which would be inefficient and unnecessary. 

We should either truncate the vec before returning or provide the appropriate `&mut [u8]` if we follow the approach mentioned in https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/257#discussion_r2085927799.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
lalitb: Agree, this is something in plan once we have more complete integration in place.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn test_roundtrip_empty_input() {
+        // Empty input
+        let input: Vec<u8> = Vec::new();
+        let compressed = lz4_chunked_compression(&input);
+        let decompressed = decompress_chunked_lz4(&compressed, input.len());
+        assert_eq!(decompressed, input);
+    }
+
+    #[test]
+    fn test_roundtrip_unicode_input() {
```
utpilla, line 100: We don't need a test for unicode string. The compression method is operating on bytes as input. It's agnostic of the source of the bytes. It doesn't matter if it came from an ASCII-only string, a UTF-8 string, or an invalid UTF-8 string.

What this test does check though is whether the compression works for an input size which is less than the `CHUNK_SIZE`. I would say you could rename the test to `test_roundtrip_small_input()` and use something like `let input = vec![0; 100];` instead of using some UTF-8 string.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -0,0 +1,149 @@
...
+
+#[cfg(test)]
+mod tests {
+    use crate::payload_encoder::lz4_chunked_compression::lz4_chunked_compression;
+    use lz4_flex::block::decompress;
+
+    #[test]
+    fn test_roundtrip_large_input() {
+        // Very large input (10 MB of repeating pattern)
+        let input = vec![0xAB; 10 * 1024 * 1024];
```
utpilla, line 84: Let's also add a test for an input size which is greater than `CHUNK_SIZE` but not divisible by `64 * 1024`.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
lalitb: On thinking more - Ideally, we should limit the input size during encoding, since the compressed output will typically be smaller due to padding (bond encoding has large number of zeros due to padding which get eliminated during compression). By enforcing a maximum input size before compression, we avoid needing to set an output size limit here.

lalitb: If the input size is bounded by upstream, we can avoid adding maximum size for output here?

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
lalitb: Good point. Have fixed accordingly.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
lalitb: Please see comment above.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
lalitb: have added a separate test for less than chunk size -`test_roundtrip_less_than_chunk_size`, and removed this.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
lalitb: Good point. Done.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
utpilla: This is another reason why the downstream methods should accept a `&mut [u8]` instead of creating vec in their own implementation. If they only accept `&mut [u8]`, then we enforce that these methods cannot add or allocate more items otherwise we would have to rely on convention that the upstream would always send us a reasonable size.

lalitb: This can be done, but not sure if it would be ergonomic, as the callee would now need to know the size requirements of the compression algorithm and accordingly pass the buffer as below. Or else allocate the size to be large enough based on _some_ approximation
```rust
    use lz4_flex::block::get_maximum_output_size;
    const CHUNK_SIZE: usize = 64 * 1024;
    let num_chunks = (input.len() + CHUNK_SIZE - 1) / CHUNK_SIZE;
    let max_compressed_per_chunk = get_maximum_output_size(CHUNK_SIZE);
    let required_output_size = num_chunks * (4 + max_compressed_per_chunk);

    let mut output_buffer= vec![0u8; required_output_size];
    let compressed_len = lz4_chunked_compression(&input, &mut output_buffer)
```

The benefit of sending the output buffer by caller I can see if it now has full control of allocation - and can either pass heap or stack allocated memory. We can provide both the APIs for now, or else add TODO to revisit, and keep existing. I am even fine if you want to change it to `&mut v[]`, with TODO to revisit if any change is need later :)

utpilla: > Or else allocate the size to be large enough based on some approximation

It doesn't have to be an approximation. The downstream components such as compression can offer an API such as `fn get_max_buffer_size_for_compression() -> usize` which the upstream components can call and use it to allocate the vec. That way the logic for the size calculation would still remain within the individual component.

> The benefit of sending the output buffer by caller I can see is it now has full control of allocation - and can either pass heap or stack allocated memory. We can provide both the APIs for now, or else add TODO to revisit, and keep existing. I am even fine if you want to change it to &mut v[], with TODO to revisit if any change is need later :)

You don't have to change it to use `&mut [u8]` in this PR. You can add a TODO to revisit this later.

lalitb: Thanks, added TODO for passing output buffer as mutable slice.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -0,0 +1,177 @@
...
+        let compressed_size = compress_into(
+            chunk,
+            &mut output[data_offset..data_offset + max_chunk_compressed],
+        )?;
+
+        // Write the actual compressed length as little-endian u32
+        let compressed_size_le = (compressed_size as u32).to_le_bytes();
+        output[header_offset..header_offset + 4].copy_from_slice(&compressed_size_le);
+        // Truncate output to actual size (header + compressed)
+        output.truncate(data_offset + compressed_size);
```
utpilla, line 77: This can be optimized further without needing to resize and truncate during each iteration.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
utpilla: I don't see any comments related to this 😄 But I noticed that you're truncating the vec before returning now.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
utpilla: Add a TODO for this if you want to address this later.

opentelemetry-exporter-geneva/geneva-uploader/Cargo.toml
```diff
@@ -15,7 +15,7 @@ native-tls = "0.2"
 thiserror = "2.0"
 chrono = "0.4"
 url = "2.2"
-
+lz4_flex = { version = "0.11", features = ["safe-encode"], default-features = false }
```
bantonsson, line 18: Should this be behind a feature or should it always be included?

lalitb: It needs to be always included, as it is required for the actual compression.

bantonsson: Yes, if you use the compression it needs to be included, but right now the compression is dead code. So my question is if the compression will always be used so you always need to include this.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
lalitb: Added TODO for now, will optimize it in subsequent PR for encoder.

opentelemetry-exporter-geneva/geneva-uploader/Cargo.toml
lalitb: Understood - compression will definitely be used as part of #261.

PR #263, opened 2025-05-15
chore: leverage fallback resolver for MSRV check

gruebel:
## Changes

Rust 1.84 introduced a new resolver type which is MSRV aware


hdost (changes requested):
We have typically tried to reserve bumping of dependencies for when it's necessary by our dependencies. By changing the msrv check this negates that.

gruebel:
> We have typically tried to reserve bumping of dependencies for when it's necessary by our dependencies. By changing the msrv check this negates that.

not sure I follow, but there is no dependency upgrade or something like that. Rust 1.84 added a new resolver, which is MSRV aware so it chooses automatically the latest version, which fits the given MSRV. This means you don't need to maintain a separate patch mechanism for making sure you don't update to a version with a higher MSRV need. Here it is nicely explained https://doc.rust-lang.org/cargo/reference/resolver.html#rust-version

The MSRV check still works the same and this is proof it works as intended otherwise it would have failed with complaining about the dependencies needing a higher MSRV.

bantonsson (approved):

lalitb:
@hdost these changes look good to me. The resolver will now automatically pick compatible crate versions based on the MSRV, so we can drop the manual patching.

cijothomas:
@hdost merging now. We can follow up if there are any concerns.

hdost:
The resolver doesn't prevent us from merging code that is no longer compatible even if the resolver does checking.

lalitb:
> 
The resolver doesn't prevent us from merging code that is no longer compatible even if the resolver does checking.

hdost:
> > The resolver doesn't prevent us from merging code that is no longer compatible even if the resolver does checking.
> 
> 
> 
>  I believe it will. The Cargo resolver in **1.84 onwards is MSRV-aware** - which mean if there is no compatible dependency version available for our MSRV, the build will fail - it will not silently pick an incompatible version.
> 

I would agree with that part but are we certain that i couldn't try to add a new feature to the that's only available in version MSRV+N and it won't prevent it?

hdost:
> > 
> 
> The resolver doesn't prevent us from merging code that is no longer compatible even if the resolver does checking.

Exactly so you're proving my point. The resolver fallback doesn't prevent us from using things available in newer versions only. Unless the compiler cares which version is set in the cargo file when compiling then there's still additional protection needed.

hdost:
@gruebel thank you for confirming it works, just saw the LazyLock PR. That's even better of tool than I had even imagined. Thanks also for this initial PR.

PR #272, opened 2025-05-22
chore: switch to LazyLock

gruebel:
just for testing the new MSRV flow


cijothomas:
Thanks for confirming that our msrv checks are working as desired! Lets close this?

PR #277, opened 2025-05-27
feat: optimize user-events metrics serialization and add basic integr…

cijothomas:


opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -46,411 +89,514 @@ impl Debug for MetricsExporter {
     }
 }
 
+pub(crate) fn to_nanos(time: SystemTime) -> u64 {
```
utpilla, line 92: This method need not be `pub(crate)`.

```suggestion
fn to_nanos(time: SystemTime) -> u64 {
```

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -46,411 +89,514 @@ impl Debug for MetricsExporter {
...
+        failed_count
+    }
+
+    fn encode_and_emit_metric(
+        &self,
+        export_metric_service_request_common: &ExportMetricsServiceRequest,
+        byte_array: &mut Vec<u8>,
+        metric: &opentelemetry_sdk::metrics::data::Metric,
+    ) -> Result<(), String> {
+        match export_metric_service_request_common.encode(byte_array) {
```
utpilla, line 426: You don't have to actually encode the message to find out if the length exceeds 64KB. Use `Message:encoded_len()` to calculate the size of the serialized bytes upfront. Refer to ETW exporter for this:

https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/9741f10aaafe7e96a51e33cab0fa2030cb73d603/opentelemetry-etw-metrics/src/exporter/mod.rs#L48-L50

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -46,411 +89,514 @@ impl Debug for MetricsExporter {
...
+        //
+        // LOOP 3: Metric → Data Points
+        //   - For each metric, iterate through individual data points
+        //   - Each data point represents a unique combination of metric +
+        //     attributes + timestamp
+        //   - SERIALIZATION HAPPENS HERE: Each data point is encoded and
+        //     emitted individually
+        //
+        // PERFORMANCE OPTIMIZATIONS:
+        // - Reuse protobuf structure templates at each level to avoid repeated
```
utpilla, line 496: This can be improved further by reusing the same templates throughout the lifetime of the exporter instead of limiting the scope of reuse to each export method call.

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -46,411 +89,514 @@ impl Debug for MetricsExporter {
...
+
+    fn encode_and_emit_metric(
+        &self,
+        export_metric_service_request_common: &ExportMetricsServiceRequest,
+        byte_array: &mut Vec<u8>,
+        metric: &opentelemetry_sdk::metrics::data::Metric,
+    ) -> Result<(), String> {
+        match export_metric_service_request_common.encode(byte_array) {
             Ok(_) => {
                 otel_debug!(name: "SerializeSuccess",
```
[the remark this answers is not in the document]
cijothomas: agree.

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -46,411 +89,514 @@ impl Debug for MetricsExporter {
...
+                    bucket_counts: dp.bucket_counts().collect(),
+                    explicit_bounds: dp.bounds().collect(),
+                    exemplars: Vec::new(), // No support for exemplars
+                    flags: default_flags,
+                    min: dp.min().map(|v| v.into_f64()),
+                    max: dp.max().map(|v| v.into_f64()),
+                };
+
+            let histogram_point_proto = opentelemetry_proto::tonic::metrics::v1::Histogram {
+                aggregation_temporality: hist.temporality() as i32,
```
utpilla, line 315: Hardcode temporality for `Histogram` and `ExponentialHistogram` as `1`. It would be nice to add a comment as well saying "Only delta temporality is supported for Histgorams" and "Only delta temporality is supported for Counters but only cumulative temporality is supported for UpDownCounters".

utpilla (approved):
Left some suggestions for improvement.

I haven't looked at the integration tests. Adding byte-level serialization tests would help ensure that future modifications to the serialization logic are straightforward to verify and accept.

opentelemetry-user-events-metrics/src/exporter/mod.rs
cijothomas: nice. Don't see this making a big difference in current implementation, as possibility of hitting the limit with just one datapoint is very low. But this will come in handy when we try to put multiple datapoints into one tracepoint::write!

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -46,411 +89,514 @@ impl Debug for MetricsExporter {
...
+                }
+            }
         }
 
-        Ok(())
+        // Even a single failure in the export process is considered a failure of overall export
+        // The debug level logs will show exactly which metrics failed
+        if has_failures {
+            Err(OTelSdkError::InternalFailure(
+                "Failed to export some metrics due to serialization or tracepoint write errors"
```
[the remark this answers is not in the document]
cijothomas: adding improvement for a future milestone.

opentelemetry-user-events-metrics/src/exporter/mod.rs
utpilla: Yeah, it might also be useful with exporting exemplars.

PR #280, opened 2025-05-27
chore: Temp fix CI runs

cijothomas:
CI is broken on main. Looks like related to `requests` version update.
Temporarily removing Geneva-exporter from workspace to unblock CI.
(I didn't investigate details on why this crate alone causes issue though there are other crates using reqwest.)


TommyCpp (approved):
Approve to unblock

utpilla (approved):

PR #284, opened 2025-06-04
fix: [Geneva Exporter] Add  LZ4 compression benchmark

lalitb:
## Changes

- Added benchmark test


utpilla:
@lalitb We want the tests related to the changes in the PR to be run in the CI. Could we update the workspace to include these projects first? https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/4bf46670bda776d38d064c046197330f79b0f7d1/Cargo.toml#L14-L16

opentelemetry-exporter-geneva/geneva-uploader/src/bench.rs
```diff
@@ -0,0 +1,46 @@
...
+
+            | Input size  | Median time        | Throughput         |
+            |-------------|--------------------|--------------------|
+            | 1 byte      | ~711 ns            | ~1.3 MiB/s         |
+            | 1 KiB       | ~31.8 us           | ~30.6 MiB/s        |
+            | 1 MiB       | ~3.44 ms           | ~291 MiB/s         |
+        - No significant regressions or improvements detected.
+        - Machine: WSL2, Linux 6.6.36, AMD EPYC 7763 64-core (8 cores visible), 62GiB RAM, 16GiB swap.
+    */
+    #[test]
```
utpilla, line 26: This setup looks a bit unusual. Does this mean that we would be running this benchmark in the CI test? I think we only want to run this with `cargo bench`. Could we follow something similar to how we have added benchmarks in other methods?

lalitb: This is testing the crate private code, which won't be possible under `benches/*`. Let me see if this is something which can be re-exported as public specifically for benchmark, and then add it there. Else, will make it to be ignored with cargo test.

lalitb:
> Could we update the workspace to include these projects first?

Sure.

opentelemetry-exporter-geneva/geneva-uploader/src/bench.rs
utpilla: Got it. I would suggest using `[[bench]]` target in that case and keep the benchmarks under `src/`. That way you can access private methods of the crate.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -35,48 +35,49 @@ use lz4_flex::block::{compress_into, get_maximum_output_size};
...
+        let mut output = Vec::with_capacity(4 + compressed_size);
+        output.extend_from_slice(&(compressed_size as u32).to_le_bytes());
+        output.extend_from_slice(&temp_buf[..compressed_size]);
+        return Ok(output);
+    }
+
+    let max_chunk_compressed = get_maximum_output_size(CHUNK_SIZE);
+    let num_chunks = input.len().div_ceil(CHUNK_SIZE);
+    let mut output = Vec::with_capacity(num_chunks * (4 + max_chunk_compressed));
+    let mut temp_buf = vec![0u8; max_chunk_compressed];
```
utpilla, line 72: I don't follow. We had removed all temporary buffer allocation. Why are we introducing it again?

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
```diff
@@ -35,48 +35,49 @@ use lz4_flex::block::{compress_into, get_maximum_output_size};
...
+/// Compresses input data in user-defined chunk size using LZ4.
+/// Each chunk is prefixed with a 4-byte (little-endian) length header.
+pub(crate) fn lz4_chunked_compression_custom<const CHUNK_SIZE: usize>(
+    input: &[u8],
+) -> Result<Vec<u8>, lz4_flex::block::CompressError> {
+    if input.len() == 0 {
+        return Ok(Vec::new());
+    }
+    // Special case: very small data (less than one chunk)
+    if input.len() <= CHUNK_SIZE {
```
utpilla, line 59: From the PR description:

> Added a fast path for small inputs (≤ one chunk): compresses the whole input in a single step, avoiding the main loop and extra allocations.

It doesn't look like there is anything special happening for this case. It would have been same (cost-wise) even if you didn't have a special case for this. The `while` loop would have only run once.

opentelemetry-exporter-geneva/geneva-uploader/src/bench.rs
lalitb: As discussed offline, have made to be ignored by cargo test.

opentelemetry-exporter-geneva/geneva-uploader/src/bench.rs
```diff
@@ -0,0 +1,46 @@
...
+    fn lz4_benchmark() {
+        let mut criterion = Criterion::default()
+            .sample_size(100)
```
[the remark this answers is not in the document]
lalitb: The benchmark was not completing in default time, so the sample size was changed. But not getting the issue anymore, so removed this.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
lalitb: Good point. This is not correct. I see improvements in the small size with this, but not with large size. Have reverted the optimization efforts - the change is now limited to using the benchmark! Thanks for catching this.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/lz4_chunked_compression.rs
lalitb: Agree.

opentelemetry-exporter-geneva/geneva-uploader/src/bench.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    /*
+        - Criterion benchmarks (lz4_chunked_compression, lz4_flex backend):
+
+            | Input size  | Median time        | Throughput         |
+            |-------------|--------------------|--------------------|
+            | 1 byte      | ~533 ns            | ~1.6 MiB/s         |
+            | 1 KiB       | ~597 ns           | ~1.5 GiB/s        |
+            | 1 MiB       | ~42 us           | ~22.1 GiB/s         |
+        - No significant regressions or improvements detected.
```
utpilla, line 23: nit: I think you could remove this line.

opentelemetry-exporter-geneva/geneva-uploader/src/bench.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /*
+        - Criterion benchmarks (lz4_chunked_compression, lz4_flex backend):
+
+            | Input size  | Median time        | Throughput         |
+            |-------------|--------------------|--------------------|
+            | 1 byte      | ~533 ns            | ~1.6 MiB/s         |
+            | 1 KiB       | ~597 ns           | ~1.5 GiB/s        |
+            | 1 MiB       | ~42 us           | ~22.1 GiB/s         |
```
utpilla, line 22: It would be good to identify what size would better represent the actual scenario.

opentelemetry-exporter-geneva/geneva-uploader/src/bench.rs
```diff
@@ -0,0 +1,46 @@
...
+    #[test]
+    #[ignore = "benchmark on crate private, ignored by default during normal test runs"]
+    /// To run: $cargo test --release lz4_benchmark -- --nocapture --ignored
+    fn lz4_benchmark() {
+        let mut criterion = Criterion::default();
+
+        let mut group = criterion.benchmark_group("lz4_compression");
+
+        for size in [1, 1024, 1024 * 1024] {
+            let data = setup_test_data(size);
```
utpilla, line 35: We should consider adding a benchmark for OTLP data as well.

PR #288, opened 2025-06-05
feat: user-events logs exporter - add support for Resource attributes

cijothomas:
Adding opt-in support for Resource attributes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
     continue-on-error: ${{ matrix.rust == 'beta' }}
     steps:
     - name: Harden the runner (Audit all outbound calls)
-      uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+      uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
       with:
         egress-policy: audit
 
@@ -72,7 +72,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - name: Harden the runner (Audit all outbound calls)
-      uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+      uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
       with:
         egress-policy: audit
 
@@ -103,7 +103,7 @@ jobs:
     continue-on-error: true
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+        uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
         with:
           egress-policy: audit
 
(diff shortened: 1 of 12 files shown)


lalitb (approved):

opentelemetry-user-events-logs/src/logs/processor.rs
```diff
@@ -66,6 +66,7 @@ impl opentelemetry_sdk::logs::LogProcessor for Processor {
 #[derive(Debug)]
 pub struct ProcessorBuilder<'a> {
     provider_name: &'a str,
+    resource_attribute_keys: Vec<String>,
```
utpilla, line 69: Why not use a `HashSet` to begin with instead of collecting strings into a vec and then collecting the vec into `Hashset` in the exporter new method.

opentelemetry-user-events-logs/src/logs/processor.rs
```diff
@@ -93,7 +94,18 @@ impl<'a> ProcessorBuilder<'a> {
...
+        Self {
+            provider_name,
+            resource_attribute_keys: vec![],
+        }
+    }
+
+    /// Sets the resource attributes for the processor
+    /// This is the list of attribute keys, which, if present in Resource attributes,
+    /// will be exported with each log.
+    pub fn with_resource_attributes<S: Into<String>>(mut self, attributes: Vec<S>) -> Self {
```
utpilla, line 106: We could make the API more ergonomic for the user by accepting an iterator instead of a vec. You could also use `Cow<'static,str>` if we want to avoid allocation for static strings.

```suggestion
    pub fn with_resource_attributes<I, S>(mut self, attributes: I) -> Self
    where
        I: IntoIterator<Item = S>,
        S: Into<Cow<'static, str>>,
```

opentelemetry-user-events-logs/src/logs/processor.rs
```diff
@@ -93,7 +96,25 @@ impl<'a> ProcessorBuilder<'a> {
...
+        }
+    }
+
+    /// Sets the resource attributes for the processor This is the list of
+    /// attribute keys, which, if present in Resource attributes, will be
+    /// exported with each log. Without this, Resource attributes (except
+    /// service.name and service.instance.id, which are exported as
+    /// cloud.roleName, cloud.roleInstance respectively) are ignored and are not
+    /// exported.
+    pub fn with_resource_attributes<I, S>(mut self, attributes: I) -> Self
```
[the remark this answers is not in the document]
cijothomas: On second thought, since resources attributes are sent inside PartC envelope, it won't clash with anything in partB, so its okay to leave this as-is, and let the listener deal with conflict.

(I think we still need to validate when service.name,service.instance.id is contained or when it is empty as they indicate user mistakes and better to fail fast)

opentelemetry-user-events-logs/src/logs/processor.rs
```diff
@@ -93,7 +96,47 @@ impl<'a> ProcessorBuilder<'a> {
...
+    /// Focus on attributes that are truly specific to your application instance
+    /// and cannot be easily determined by the local agent.
+    pub fn with_resource_attributes<I, S>(mut self, attributes: I) -> Self
+    where
+        I: IntoIterator<Item = S>,
+        S: Into<Cow<'static, str>>,
+    {
+        self.resource_attribute_keys = attributes
+            .into_iter()
+            .map(|s| s.into().into_owned())
```
utpilla, line 137: This negates the value of using `Cow<'static, str>`. If we want to avoid allocations for static strings, we need to use `Cow<static, str>` throughout the processor builder and exporter. If not, we could have simply used `S: Into<String>`.

opentelemetry-user-events-logs/src/logs/processor.rs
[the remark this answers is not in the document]
cijothomas: Since exporter put user-provided attributes in PartC, which is separate struct that PartA/B, we can let listener's deal with conflict.

For reserved strings, we can document this. given we are not losing info in this scenario, this is acceptable.

utpilla (approved):

PR #261, opened 2025-05-14
feat: [Geneva Exporter] - Add Serializer

lalitb:
Fixes #
Design discussion issue (if applicable) #

## Changes

Please provide a brief description of the changes here.

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #297, opened 2025-06-16
perf: [Geneva Exporter] Add throughput test framework for async tasks, and use it for geneva_exporter

lalitb:
## Change
Added stress test framework, for measuring the throughput and error rate of asynchronous operations under configurable concurrency. The structure and reporting style are similar to the existing synchronous throughput tester, but adapted for async workloads using Tokio.

### Highlights

- Allows users to specify:
   - Number of async worker tasks
   - Level of concurrency per worker
   - A user-provided async function to benchmark

### Reports:
- Live per-second throughput and error counts during the test
   - Total successes, failures, and success rate at the end
   - Graceful shutdown via Ctrl+C (in-flight tasks are completed)
- Designed for use with network clients, exporters, or any async Rust API
- Reporting and implementation structure are modeled closely after the existing sync version for a familiar experience

### workers:
- Controls the number of independent async tasks (Tokio tasks) that will each maintain their own concurrent pipeline.
- Typical recommendation:
   - Set to the number of logical CPU cores (num_cpus::get()) for a balanced workload.
   - Increasing beyond core count may improve throughput if your operations are highly I/O bound, but can add scheduling overhead for CPU-bound tasks.

### concurrency_per_worker:
- Controls how many async operations each worker maintains in flight.
   - Higher values mean more total concurrent operations (workers × concurrency_per_worker).
   - If your async function is I/O bound (e.g., making network calls), you can increase concurrency to better saturate the system and external resource.
   - If your async function is CPU bound, too much concurrency can lead to diminishing returns due to contention.

### General Guidance

- For network- or I/O-heavy tasks:
   - Use a higher total concurrency to keep the pipeline full and maximize throughput.
   - Start with workers = num_cpus::get() and increase concurrency_per_worker as needed.

- For CPU-heavy tasks: 
  - Matching workers to the number of CPU cores and keeping concurrency_per_worker moderate is usually optimal.

-Monitor live stats while running; adjust parameters if throughput is lower than expected or if you see resource exhaustion (CPU at 100%, high error rate, etc).

### Result for Geneva Exporter:

```
➜  stress git:(geneva-exporter-stress-test) ✗ cargo run --bin geneva_exporter --release
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/lalitb/obs/otel/rust/lalitb/opentelemetry-rust-contrib/stress/Cargo.toml
workspace: /Users/lalitb/obs/otel/rust/lalitb/opentelemetry-rust-contrib/Cargo.toml
   Compiling geneva-uploader v0.1.0 (/Users/lalitb/obs/otel/rust/lalitb/opentelemetry-rust-contrib/opentelemetry-exporter-geneva/geneva-uploader)
   Compiling stress v0.1.0 (/Users/lalitb/obs/otel/rust/lalitb/opentelemetry-rust-contrib/stress)
    Finished `release` profile [optimized] target(s) in 3.28s
     Running `/Users/lalitb/obs/otel/rust/lalitb/opentelemetry-rust-contrib/target/release/geneva_exporter`
Using mocked Geneva endpoints
Embedding ingestion_endpoint in JWT: http://127.0.0.1:53816/ingestion
MOCK JWT: eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJFbmRwb2ludCI6Imh0dHA6Ly8xMjcuMC4wLjE6NTM4MTYvaW5nZXN0aW9uIiwiZXhwIjoxNzUwODEwMTQ1fQ.dummy
WARNING: Using MockAuth for GenevaConfigClient. This should only be used in tests!
Async Throughput Test: Geneva Uploader
Workers: 14 | Concurrency/worker: 100
Total max concurrent: 1400
Press Ctrl+C to stop

Throughput:  19,783/s | Total:   19,783 | Errors: 0
Throughput:  40,492/s | Total:   60,275 | Errors: 0
Throughput:  43,452/s | Total:  103,727 | Errors: 0
Throughput:  51,372/s | Total:  155,099 | Errors: 0
Throughput:  41,008/s | Total:  196,107 | Errors: 0
Throughput:  41,318/s | Total:  237,425 | Errors: 0
Throughput:  43,023/s | Total:  280,448 | Errors: 0
Throughput:  40,052/s | Total:  320,500 | Errors: 0
Throughput:  45,197/s | Total:  365,697 | Errors: 0
Throughput:  45,105/s | Total:  410,802 | Errors: 0
Throughput:  40,959/s | Total:  451,761 | Errors: 0
Throughput:  40,480/s | Total:  492,241 | Errors: 0
Throughput:  39,205/s | Total:  531,446 | Errors: 0
Throughput:  39,803/s | Total:  571,249 | Errors: 0
Throughput:  47,391/s | Total:  618,640 | Errors: 0
Throughput:  41,252/s | Total:  659,892 | Errors: 0
Throughput:  38,921/s | Total:  698,813 | Errors: 0
Throughput:  38,332/s | Total:  737,145 | Errors: 0
Throughput:  38,831/s | Total:  775,976 | Errors: 0
Throughput:  46,672/s | Total:  822,648 | Errors: 0
Throughput:  39,286/s | Total:  861,934 | Errors: 0
Throughput:  37,787/s | Total:  899,721 | Errors: 0
^C

Graceful shutdown initiated...
Worker 4 draining 99 requests...
Worker 5 draining 99 requests...
Worker 8 draining 99 requests...
Worker 9 draining 99 requests...
Worker 2 draining 99 requests...
Worker 10 draining 99 requests...
Worker 0 draining 99 requests...
Worker 1 draining 99 requests...
Worker 7 draining 99 requests...
Worker 11 draining 99 requests...
Worker 13 draining 99 requests...
Worker 12 draining 99 requests...
Worker 3 draining 99 requests...
Worker 6 draining 99 requests...

╔══════════════════════════════════════════════════════╗
║                 FINAL RESULTS                        ║
╠══════════════════════════════════════════════════════╣
║ Duration:        22.25s                              ║
║ Total Ops:       911,684                             ║
║ Successful:      911,684                             ║
║ Failed:          0                                   ║
║ Success Rate:    100.00%                             ║
║ Throughput:      40965 ops/s                         ║
╚══════════════════════════════════════════════════════╝
```
These results can serve as a baseline for future performance improvements. Note that this isn’t a pure measurement of SDK throughput—the actual throughput could be higher, since the locally running wiremock server may introduce its own limitations when handling high levels of concurrent requests.


stress/src/geneva_exporter.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    // Use the provided stress test framework
+    println!("Starting stress test for Geneva Uploader...");
+    throughput::test_throughput(move || {
+        let client = client.clone();
+        let logs = test_logs.clone();
+        let runtime = runtime.clone();
+
+        // Execute one upload operation
+        runtime.block_on(async {
```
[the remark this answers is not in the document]
utpilla: This made me realize that the current code uploads blobs sequentially. We could improve the throughput by fixing that. Check https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/261#discussion_r2160566340.

I think this offers us an opportunity to write a throughput test which shows us the improvement once we fix the sequential upload of blobs.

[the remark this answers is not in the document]
lalitb: > This is effectively making blocking I/O calls within each worker thread.

You're correct - I'm using `block_on` here because the existing throughput framework expects a synchronous function. The framework was originally designed for CPU-bound operations, not async I/O. To properly measure async throughput, we'd need to either:
1. Modify the framework to support async functions, or
2. Replace the measurement approach with an async-native solution using tasks/futures

Let me think more on this. In fact, we may have better performance if use correct approach.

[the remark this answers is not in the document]
lalitb: Have opted for approach 1 above, and added the test framework for asynchronous tasks, and used that for Geneva Exporter.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -498,21 +498,24 @@ fn extract_endpoint_from_token(token: &str) -> Result<String> {
 
     // Base64-decode the JWT payload (2nd segment of the token).
     // Some JWTs may omit padding ('='), so we restore it manually to ensure valid Base64.
-    // This is necessary because the decoder (URL_SAFE_NO_PAD) expects properly padded input.
+    // This is necessary because the decoder (URL_SAFE) expects properly padded input.
```
utpilla, line 501: What are these changes? Are these needed for correctness?

lalitb: Was getting error in decoding token on MacOS with `URL_SAFE_NO_PAD` + locally generated token. `URL_SAFE` worked both in MacOS and WSL, for remote and locally generated token. But have reverted this change for now, will raise separate PR for this, if necessary.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -504,6 +502,11 @@ fn extract_endpoint_from_token(token: &str) -> Result<String> {
         0 => payload.to_string(),
         2 => format!("{}==", payload),
         3 => format!("{}=", payload),
+        1 => {
```
utpilla, line 505: Did you intentionally add this change for this PR?

stress/config/nginx.conf
```diff
@@ -0,0 +1,49 @@
...
+    access_log off;
+
+    server {
+        listen 8080;
+        
+        # Config endpoint - matches any GET request
+        location / {
+            if ($request_method = GET) {
+                add_header Content-Type application/json;
+                return 200 '{"IngestionGatewayInfo":{"Endpoint":"http://localhost:8080/ingestion","AuthToken":"eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJFbmRwb2ludCI6Imh0dHA6Ly9sb2NhbGhvc3Q6ODA4MC9pbmdlc3Rpb24iLCJleHAiOjE3NjQwODk2MDB9.dummy","AuthTokenExpiryTime":"2025-11-25T16:00:00+00:00"},"StorageAccountKeys":[{"AccountMonikerName":"testdiagaccount","AccountGroupName":"testgroup","IsPrimaryMoniker":true}],"TagId":"test"}';
```
utpilla, line 37: Can we skip authentication altogether considering it's only for local testing?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: This isn't directly related to the PR, but for JWT, the payload length modulo 4 can never be 1 - even if padding is missing. So, the check for 1 is correct, but the _ match arm isn't needed and can be removed.

stress/src/async_throughput.rs
```diff
@@ -0,0 +1,257 @@
...
+            stats: Arc::new(SimpleStats::new()),
+            stop: Arc::new(AtomicBool::new(false)),
+        }
+    }
+
+    /// Run the throughput test with the given async function
+    ///
+    /// # Arguments
+    /// * `name` - Test name for display
+    /// * `workers` - Number of worker tasks
```
utpilla, line 61: This worker approach seems to be adding unnecessary overhead. The total concurrency (number of operations running simultaneously) is the only thing that matters for stress testing. The number of "worker" threads working on these operations should be decided by the tokio runtime flavor:
- One for `current_thread` runtime
- Multiple for `multi_thread` runtime

stress/src/async_throughput.rs
```diff
@@ -0,0 +1,257 @@
...
+
+        tokio::spawn(async move {
+            let mut futures = FuturesUnordered::new();
+            let mut local_success = 0u64;
+            let mut local_errors = 0u64;
+
+            // Main loop
+            while !stop.load(Ordering::SeqCst) {
+                // Keep pipeline full
+                while futures.len() < concurrency && !stop.load(Ordering::SeqCst) {
```
utpilla, line 176: I think using [`buffer_unordered`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffer_unordered) with the required concurrency is much better than manually maintaining concurrency.

stress/src/async_throughput.rs
utpilla: Yes.

stress/src/geneva_exporter.rs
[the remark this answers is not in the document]
utpilla: Please feel free to keep it in your PR. Thanks for testing this! I’m happy to have contributed!

stress/src/async_throughput.rs
lalitb: The code is updated accordingly.

stress/src/async_throughput.rs
lalitb: Good point. done.

stress/src/async_throughput.rs
```diff
@@ -0,0 +1,259 @@
...
+                ops,
+                success,
+                failed,
+                elapsed.as_secs_f64(),
+                throughput
+            );
+        });
+
+        // Create operation factory Arc for sharing
+        let operation_factory = Arc::new(operation_factory);
```
utpilla, line 105: This is not required.

stress/src/async_throughput.rs
```diff
@@ -0,0 +1,259 @@
...
+
+        // Create infinite stream of operations
+        let mut operation_stream = stream::iter(0..)
```
utpilla, line 108: We need to spawn tasks for these operations to utilize the worker threads:

```rust
let mut operation_stream = stream::iter(0..)
    .map(|_| tokio::task::spawn(operation_factory()))
    .buffer_unordered(config.concurrency);
```

stress/src/geneva_exporter.rs
```diff
@@ -0,0 +1,339 @@
...
+            let logs_clone = logs.clone();
+
+            let stats = ThroughputTest::run_continuous("Geneva Upload", config, move || {
+                let client = client_clone.clone();
+                let logs = logs_clone.clone();
+                async move {
+                    client
+                        .upload_logs(&logs)
+                        .await
+                        .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))
```
utpilla, line 269: Remove these `map_err` statements?

stress/src/geneva_exporter.rs
```diff
@@ -0,0 +1,339 @@
...
+    match mode {
+        "continuous" => {
+            // Run continuous test
+            let config = ThroughputConfig {
+                concurrency,
+                report_interval: std::time::Duration::from_secs(5),
+                target_ops: None,
+            };
+
+            let client_clone = client.clone();
```
utpilla, line 259: You don't need these additional clones.

stress/src/async_throughput.rs
```diff
@@ -0,0 +1,259 @@
...
+
+        // Process stream until interrupted
+        while let Some(result) = operation_stream.next().await {
+            completed_ops.fetch_add(1, Ordering::Relaxed);
+            match result {
+                Ok(_) => {
+                    successful_ops.fetch_add(1, Ordering::Relaxed);
+                }
+                Err(e) => {
+                    failed_ops.fetch_add(1, Ordering::Relaxed);
```
utpilla, line 123: I think we can avoid this repeated computation for failed operations. The reporting could simply subtract successful operations from completed operations.

stress/src/geneva_exporter.rs
```diff
@@ -0,0 +1,339 @@
...
+            role_instance: "test".to_string(),
+        };
+
+        let client = GenevaClient::new(config).await?;
+        Ok((client, Some(mock_server.uri())))
+    }
+}
+
+#[tokio::main(flavor = "current_thread")]
+//#[tokio::main(flavor = "multi_thread")]
```
utpilla, line 227: It would be good to also make the runtime type a command line argument.

stress/src/async_throughput.rs
lalitb: Thanks for fixing this, done.

stress/src/geneva_exporter.rs
lalitb: Good point. done.

stress/src/async_throughput.rs
```diff
@@ -0,0 +1,238 @@
...
+        // Spawn reporting thread
+        let _reporting_handle = std::thread::spawn(move || loop {
+            std::thread::sleep(report_interval);
+            let ops = completed_clone.load(Ordering::Relaxed);
+            let success = successful_clone.load(Ordering::Relaxed);
+            let elapsed = start_time.elapsed();
+            let throughput = ops as f64 / elapsed.as_secs_f64();
+
+            println!(
+                "Progress: {} ops completed ({} successful) in {:.2}s = {:.2} ops/sec",
```
utpilla, line 90: Same as https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/297#discussion_r2176491549

stress/src/async_throughput.rs
```diff
@@ -0,0 +1,249 @@
...
+                success,
+                success_percentage,
+                elapsed.as_secs_f64(),
+                throughput
+            );
+        });
+
+        // Create infinite stream of operations
+        let mut operation_stream = stream::iter(0..)
+            .map(|_| tokio::task::spawn(operation_factory()))
```
utpilla, line 114: I believe we only need to spawn tasks for multi-threaded runtime. We should avoid this overhead for current threaded runtime so that we get more accurate measurements and have a fair comparison with multi-threaded runtime.

lalitb: Actually thinking more, I believe the avoiding `task::spawn` should be enough - this way we keep the same overhead (Arc) across both runtimes, giving consistent comparison :)

utpilla: I'm okay with removing `task::spawn` as I learnt that `buffer_unordered` would anyway poll all of its items.

Keeping an `Arc` for both is not a fair comparison. If a certain approach doesn't need thread synchronization to work, then why force it to pay that cost when measuring it?

stress/src/async_throughput.rs
```diff
@@ -0,0 +1,295 @@
...
+                    }
+                    Err(e) => {
+                        eprintln!("Operation failed: {}", e);
+                    }
+                }
+            }
+        }
+
+        // Return final stats
+        let duration = start_time.elapsed();
```
utpilla, line 158: nit: Since this function would never return and relies on the user to Ctrl +C, it doesn't really need a return value.

stress/src/geneva_exporter.rs
```diff
@@ -0,0 +1,352 @@
...
+     - OS: Ubuntu 6.6 (WSL2 on Windows 11), x86_64
+
+     Using Tokio multi-thread runtime
+    Using standalone mock server at: http://localhost:8080
+    WARNING: Using MockAuth for GenevaConfigClient. This should only be used in tests!
+    Warming up token cache...
+
+    $ cargo run --bin geneva_exporter --release -- multi 500  continuous
+    Testing Geneva Upload with concurrency level: 500
+        Progress: 288092 ops completed (288092 successful, 100.0%) in 5.00s = 57580.46 ops/sec
```
utpilla, line 20: It looks like the upload task is more CPU intensive than the I/O. It would be good to check what's the optimal concurrency level for which we get the highest throughput. I believe we would see the throughput plateau after a certain concurrency level. Anything after that would be adding more overhead in terms of memory used etc.

lalitb:
> Left a few suggestions for follow up.

Thanks for reviewing this! Your feedback, as always, really helped get the stress framework right :)

PR #298, opened 2025-06-16
feat: using otel 0.30; disabling bench

### LINKED ISSUE #296: Update opentelemetry-etw-metrics to latest opentelemetry_sdk
Need to rewrite transforms too, similar to https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/277

psandana:
Fixes #296 

## Changes


opentelemetry-etw-metrics/src/exporter/mod.rs
```diff
@@ -69,30 +71,34 @@ fn emit_export_metric_service_request(
...
+                    AggregatedMetrics::F64(data) => data.into(),
+                    AggregatedMetrics::I64(data) => data.into(),
+                    AggregatedMetrics::U64(data) => data.into(),
+                };
 
                 // This ExportMetricsServiceRequest is created for each metric and will hold a single data point.
                 let mut export_metrics_service_request = ExportMetricsServiceRequest {
                     resource_metrics: vec![TonicResourceMetrics {
-                        resource: Some((&metrics.resource).into()),
+                        resource: Some((metrics.resource()).into()),
```
cijothomas, line 95: I think we can create the Resource part once per export, instead of once per Metric, as it'll be same for entire batch for metric. (Ideally this can be even re-used across exports, as Resource is fixed!). It may be easier to do this when consolidating user-events and etw encoding into a common helper.

opentelemetry-etw-metrics/src/exporter/mod.rs
```diff
@@ -102,83 +108,76 @@ impl PushMetricExporter for MetricsExporter {
...
-                                )?;
-                            }
+                                .data = Some(TonicMetricData::Histogram(TonicHistogram {
+                                aggregation_temporality: hist.aggregation_temporality,
+                                data_points: vec![data_point],
+                            }));
+                            emit_export_metric_service_request(
+                                &export_metrics_service_request,
+                                &mut encoding_buffer,
+                            )?;
```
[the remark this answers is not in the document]
psandana: Done. I chose to add a stack buffer. If desired to own it at the MetricsExporter, we would require interior mutability.

opentelemetry-etw-metrics/src/exporter/mod.rs
```diff
@@ -69,30 +72,34 @@ fn emit_export_metric_service_request(
...
-            .resource
+            .resource()
             .schema_url()
             .map(Into::into)
             .unwrap_or_default();
 
-        // TODO: Reuse this vec across exports by storing it in `MetricsExporter`
-        let mut encoding_buffer = Vec::new();
+        let resource: Resource = metrics.resource().into();
+        let mut encoding_buffer = Vec::<u8>::with_capacity(etw::MAX_EVENT_SIZE);
```
cijothomas, line 83: given we serialize one datapoint at a time, pre-allocating the etw_max would be quite excess. Something we can revisit and improve in the future. (Since vec is dropped after the export, not a big issue)

opentelemetry-etw-metrics/Cargo.toml
```diff
@@ -29,9 +30,10 @@ default = ["internal-logs"]
 [package.metadata.cargo-machete]
 ignored = ["tracing"]
 
-[[bench]]
-name = "exporter"
-harness = false
+# TODO: Enabled this once we can use data without aggregation (git mv disabled_benches benches).
```
cijothomas, line 33: unsure why rename/move is done?

psandana: Benches are automatically compiled if located in the `benches` folder. This declaration here is to switch the `harness = false` flag to it, but removing the declaration does not prevent the run of the bench.

Thus, we either delete the bench, or move it to a location where cargo does not capture it automatically.

PR #359, opened 2025-07-09
feat: opentelemetry-etw-logs: ability to disable cross compat provide…

psandana:
…r names

Fixes #

## Changes

Adding `ProcessorBuilder::disable_cross_compat_provider_name()` method that allows the processor to use a provider name that is not compatible with UserEvents, but conforming with ETW requirements which allows hyphens (`-`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,7 +113,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@b3b07ba8b418998c39fb20f53e8b695cdcc8de1b
         with:
           toolchain: stable
-      - uses: taiki-e/install-action@2a3397031cc3ea3b5341eab31f073a602ab6d852 # v2.56.10
+      - uses: taiki-e/install-action@c07504cae06f832dc8de08911c9a9c5cddb0d2d3 # v2.56.13
         with:
           tool: cargo-msrv
       - name: Check MSRV for all crates
@@ -172,7 +172,7 @@ jobs:
         with:
             repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: cargo install cargo-llvm-cov
-        uses: taiki-e/install-action@2a3397031cc3ea3b5341eab31f073a602ab6d852 # v2.56.10
+        uses: taiki-e/install-action@c07504cae06f832dc8de08911c9a9c5cddb0d2d3 # v2.56.13
         with:
           tool: cargo-llvm-cov
       - name: cargo generate-lockfile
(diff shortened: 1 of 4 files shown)


opentelemetry-etw-logs/src/processor.rs
```diff
@@ -85,18 +85,30 @@ impl opentelemetry_sdk::logs::LogProcessor for Processor {
...
         ProcessorBuilder {
             options: Options::new(provider_name.to_string()),
+            cross_compat_provider_name: true,
         }
     }
 
+    /// Disables the cross-compatibility provider name with UserEvents (Linux).
+    ///
+    /// Allows the processor to use a provider name that is not compatible with UserEvents, but conforming with ETW requirements which allows hyphens (`-`).
+    pub fn disable_cross_compat_provider_name(mut self) -> Self {
```
cijothomas, line 107: `allow_etw_only_provider_name` - suggested alternate name. "disable>>" feels bit indirect.

berkus: there are only two options - we're either compat or not compat;

it makes sense to indicate from the start that you WANT the non-compat version.

You could rename new to something more sensible, but then it's a breaking change. 
`new_etw_compat_only`? dunno
`allow_noncompatible_provider_name` perhaps? dunno.

psandana: Let's go with the path of least resistance and use a new c'tor

opentelemetry-etw-logs/src/processor.rs
```diff
@@ -118,21 +153,32 @@ impl ProcessorBuilder {
...
     }
+
+    if cross_compat_provider_name && provider_name.contains('-') {
+        return Err("Provider name must not contain hyphens ('-') when cross-compatibility with UserEvents is enabled. Use ProviderBuilder::builder_etw_compat_only() to enable ETW only compatibility instead.".into());
+    }
+
     if !provider_name
         .chars()
-        .all(|c| c.is_ascii_alphanumeric() || c == '_')
+        .all(|c| c.is_ascii_alphanumeric() || c == '_' || (!cross_compat_provider_name && c == '-'))
```
berkus, line 181: it makes sense to simplify this expression by turning "cross_compat_provider_name" into "etw_only_provider_name" - then it will look like `|| (etw_only_provider_name && c == '-')` and also makes constructor assignments more readable

opentelemetry-etw-logs/src/processor.rs
```diff
@@ -118,21 +153,32 @@ impl ProcessorBuilder {
...
     {
         return Err("Provider name must contain only ASCII alphanumeric characters or '_'.".into());
```
berkus, line 183: This message is invalid in the case of enabled "-" support, and why do we have a separate message for the same thing above?

opentelemetry-etw-logs/src/processor.rs
psandana: Fixed

cijothomas (approved):

Issue #306, opened 2025-06-20
[Geneva Exporter] Optimize log batching in OtlpEncoder


lalitb:
# Optimize log batching in OtlpEncoder

## Description
The current implementation of log batching in the `OtlpEncoder` can be optimized in a couple of ways:

1. **Avoid storing `event_name` in the batch value:**
   - The `event_name` is already part of the batch key, so storing it again in the value is redundant.
   - By removing `event_name` from the value, we can avoid cloning it for HashMap lookups.
   - Update the `BatchValue` type to remove the `event_name` field:
     ```rust
     type BatchValue = (CentralSchemaEntry, Vec<(u8, Vec<u8>)>); // (level, row_buffer)
     ```
   - Modify the code that inserts into batches and creates `CentralEventEntry` to use the `event_name` from the batch key instead of the value.

2. **Perform equality check on field names in addition to hash check:**
   - The current implementation relies solely on the hash of the field names to determine schema uniqueness.
   - However, hash collisions can occur, leading to incorrect schema matching.
   - To ensure schema uniqueness, we should perform an additional equality check on the field names.
   - Update the `get_or_create_schema` method to compare the field names of the existing schema in the cache with the field names of the new schema.
   - If the field names match, consider it the same schema; otherwise, create a new schema entry.

3. **Consider removing schema tracking and simplify grouping:**
   - Instead of tracking different schemas, we could simplify the grouping by only tracking `event_name`.
   - Serialize whatever fields the corresponding log records have, without the need for complex grouping based on schemas.
   - This approach would reduce the complexity of the code and eliminate the need for schema tracking and matching.
   - Investigate the impact of this simplification on the overall functionality and performance of the `OtlpEncoder`.

4. **The upload for batches is sequential as mentioned [here](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/261#discussion_r2160566340). This needs to be made concurrent.
## Implementation
To implement these optimizations, the following changes are needed:

1. Update the `BatchValue` type to remove the `event_name` field.
2. Modify the code that inserts into batches to use the `event_name` from the batch key.
3. Update the `get_or_create_schema` method to perform an equality check on field names in addition to the hash check.
   - Retrieve the existing schema from the cache using the schema ID.
   - Compare the field names of the existing schema with the field names of the new schema.
   - If the field names match, return the existing schema; otherwise, create a new schema entry.
[...]

PR #372, opened 2025-07-14
feat: [Geneva Exporter] Add concurrent uploads to GenevaClient with configurable parallelism

lalitb:
Towards 4th point of #306.

## Changes

Enhanced GenevaClient to support concurrent uploads with configurable parallelism, improving throughput for bulk log ingestion.

## Changes
- Added `max_concurrent_uploads` configuration option to `GenevaClientConfig`
- Modified `upload_logs()` to use `buffer_unordered` for concurrent upload processing
- Moved LZ4 compression into async tasks to enable concurrent compression
- Defaults to number of CPU cores if `max_concurrent_uploads` is not specified
- Enhanced error messages to include event names for better debugging

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
@@ -4,6 +4,7 @@ use crate::config_service::client::{AuthMethod, GenevaConfigClient, GenevaConfig
 use crate::ingestion_service::uploader::{GenevaUploader, GenevaUploaderConfig};
 use crate::payload_encoder::lz4_chunked_compression::lz4_chunked_compression;
 use crate::payload_encoder::otlp_encoder::OtlpEncoder;
+use futures::stream::{self, StreamExt};
 use opentelemetry_proto::tonic::logs::v1::ResourceLogs;
 use std::sync::Arc;
 
@@ -20,6 +21,8 @@ pub struct GenevaClientConfig {
     pub tenant: String,
     pub role_name: String,
     pub role_instance: String,
+    /// Maximum number of concurrent uploads. If None, defaults to number of CPU cores.
+    pub max_concurrent_uploads: Option<usize>,
     // Add event name/version here if constant, or per-upload if you want them per call.
 }
 
@@ -29,6 +32,7 @@ pub struct GenevaClient {
     uploader: Arc<GenevaUploader>,
     encoder: OtlpEncoder,
     metadata: String,
+    max_concurrent_uploads: usize,
 }
 
 impl GenevaClient {
@@ -76,10 +80,16 @@ impl GenevaClient {
             cfg.role_name,
             cfg.role_instance,
         );
+        let max_concurrent_uploads = cfg.max_concurrent_uploads.unwrap_or_else(|| {
+            std::thread::available_parallelism()
+                .map(|p| p.get())
+                .unwrap_or(4)
+        });
         Ok(Self {
             uploader: Arc::new(uploader),
             encoder: OtlpEncoder::new(),
             metadata,
+            max_concurrent_uploads,
         })
     }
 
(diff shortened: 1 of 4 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -76,10 +80,16 @@ impl GenevaClient {
             cfg.role_name,
             cfg.role_instance,
         );
+        let max_concurrent_uploads = cfg.max_concurrent_uploads.unwrap_or_else(|| {
+            std::thread::available_parallelism()
```
utpilla, line 84: We shouldn't use number of cores as the default considering our workload is I/O bound. I think the default concurrency value should match the number of connections in the `reqwest` client's connection pool because we can never execute more HTTP requests in parallel than the available connections.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -89,17 +99,49 @@ impl GenevaClient {
...
+        let upload_futures = blobs
+            .into_iter()
+            .map(|(event_name, encoded_blob, _row_count)| {
+                let uploader = Arc::clone(&self.uploader);
+                let event_version = "Ver2v0"; // TODO - find the actual value to be populated
+                async move {
+                    // TODO: Investigate using tokio::spawn_blocking for LZ4 compression to avoid blocking
+                    // the async executor thread for CPU-intensive work.
+                    let compressed_blob = match lz4_chunked_compression(&encoded_blob) {
+                        Ok(blob) => blob,
```
utpilla, line 116: nit: Switch back to using `map_err` to be more concise.

```rust
let compressed_blob = lz4_chunked_compression(&encoded_blob)
    .map_err(|e| format!("LZ4 compression failed: {e} Event: {event_name}"))?;

uploader
    .upload(compressed_blob, &event_name, event_version)
    .await
    .map(|_| ())
    .map_err(|e| format!("Geneva upload failed: {e} Event: {event_name}"))
```

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -89,17 +99,49 @@ impl GenevaClient {
...
+                        .upload(compressed_blob, &event_name, event_version)
+                        .await
+                    {
+                        Ok(_) => Ok(()),
+                        Err(e) => Err(format!("Geneva upload failed: {e} Event: {event_name}")),
+                    }
+                }
+            });
+        // Execute uploads concurrently with configurable concurrency
+        let results: Vec<Result<(), String>> = stream::iter(upload_futures)
```
utpilla, line 131: We could avoid this intermediate allocation of `Vec<Result<(), String>>` by doing all the operations in one iterator chain:

```rust
let errors: Vec<String> = stream::iter(upload_futures)
    .buffer_unordered(self.max_concurrent_uploads)
    .filter_map(|result| async move { result.err() })
    .collect()
    .await;
```

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -89,17 +99,49 @@ impl GenevaClient {
...
-            self.uploader
-                .upload(compressed_blob, &event_name, event_version)
-                .await
-                .map_err(|e| format!("Geneva upload failed: {e}"))?;
+
+        // create an iterator that yields futures for each upload
+        let upload_futures = blobs
+            .into_iter()
+            .map(|(event_name, encoded_blob, _row_count)| {
+                let uploader = Arc::clone(&self.uploader);
```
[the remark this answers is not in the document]
lalitb: Good point Done.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
lalitb: Yes, but I noticed now that there [default connection pool size](https://github.com/seanmonstar/reqwest/blob/5b705ccc6bed87f8bf1b6a980049c6764ffdf493/src/async_impl/client.rs#L289)in `reqwest` client is theoretically unlimited, constraint by the OS limits, or else the available resources. For now, I have added a TODO in the code to find the optimal value:
```rust
        let max_concurrent_uploads = cfg.max_concurrent_uploads.unwrap_or_else(|| {
            // TODO - Use a more sophisticated method to determine concurrency if needed
            // currently using number of CPU cores
            std::thread::available_parallelism()
                .map(|p| p.get())
                .unwrap_or(4)
        });
```

PR #379, opened 2025-07-18
fix: [Geneva Exporter] Fix schema ID query parameter to use actual payload schemas instead of hardcoded values

lalitb:
### Problem

The `schemaIds` query parameter was hardcoded to static values instead of reflecting the actual schemas present in the uploaded payload. While this doesn't affect ingestion functionality, it impacts backend operations including auditing, correlation, and pipeline routing that rely on accurate schema identification.

### Solution

- __Dynamic Schema ID Generation__: Extract actual schema IDs from encoded payload and pass them to the ingestion service
- __API Updates__: Enhanced upload method to accept dynamic schema IDs parameter

### Changes Made

1. __OTLP Encoder__ (`otlp_encoder.rs`):

   - Modified `encode_log_batch` to return schema IDs along with encoded data

2. __Geneva Client__ (`client.rs`):

   - Updated to extract schema IDs from encoder output

3. __Uploader__ (`uploader.rs`):

   - Enhanced `upload` method to accept dynamic `schema_ids` parameter
   - Updated query string generation to use actual schema IDs

4. __Tests__: Updated all test cases to use new method signature

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
@@ -44,13 +44,13 @@ impl OtlpEncoder {
         self.schema_cache.read().unwrap().len()
     }
 
-    /// Encode a batch of logs into a vector of (event_name, bytes, events_count)
+    /// Encode a batch of logs into a vector of (event_name, bytes, events_count, schema_ids)
     pub(crate) fn encode_log_batch<'a, I>(
         &self,
         logs: I,
         metadata: &str,
-    ) -> Vec<(String, Vec<u8>, usize)>
-    //(event_name, bytes, events_count)
+    ) -> Vec<(String, Vec<u8>, usize, Vec<u64>)>
+    //(event_name, bytes, events_count, schema_ids)
     where
         I: IntoIterator<Item = &'a opentelemetry_proto::tonic::logs::v1::LogRecord>,
     {
@@ -106,6 +106,9 @@ impl OtlpEncoder {
         for (batch_event_name, (schema_entries, events)) in batches {
             let events_len = events.len();
 
+            // Extract schema IDs from the schema entries
+            let schema_ids: Vec<u64> = schema_entries.iter().map(|s| s.id).collect();
+
             let blob = CentralBlob {
                 version: 1,
                 format: 2,
@@ -114,7 +117,7 @@ impl OtlpEncoder {
                 events,
             };
             let bytes = blob.to_bytes();
-            blobs.push((batch_event_name, bytes, events_len));
+            blobs.push((batch_event_name, bytes, events_len, schema_ids));
         }
         blobs
     }
@@ -424,6 +427,7 @@ mod tests {
         assert_eq!(result.len(), 1);
         assert_eq!(result[0].0, "test_event");
         assert_eq!(result[0].2, 1); // events_count
+        assert!(!result[0].3.is_empty()); // schema_ids should not be empty
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/central_blob.rs
```diff
@@ -1,8 +1,96 @@
...
+            .collect::<Vec<String>>()
+            .join(";")
+    }
+
+    /// Convert nanoseconds timestamp to DateTime<Utc>
+    fn nanos_to_chrono(nanos: u64) -> DateTime<Utc> {
+        if nanos > 0 {
+            let secs = (nanos / 1_000_000_000) as i64;
+            let nsec = (nanos % 1_000_000_000) as u32;
+            Utc.timestamp_opt(secs, nsec)
```
utpilla, line 36: Could we use `DateTime::from_timestamp_nanos()` instead of this?

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/central_blob.rs
```diff
@@ -1,8 +1,96 @@
...
+impl BatchMetadata {
+    /// Format schema IDs as MD5 hashes separated by semicolons
+    pub(crate) fn format_schema_ids(&self) -> String {
+        self.schema_ids
+            .iter()
+            .map(|id| {
+                let md5_hash = md5::compute(id.to_le_bytes());
+                format!("{md5_hash:x}")
+            })
+            .collect::<Vec<String>>()
```
utpilla, line 27: Let's avoid the intermediate vec allocation:

```rust
pub(crate) fn format_schema_ids(&self) -> String {
    if self.schema_ids.is_empty() {
        return String::new();
    }

    // MD5 hash is 32 hex chars + 1 semicolon per ID (except last)
    let estimated_capacity = self.schema_ids.len() * 33 - 1;
    let mut result = String::with_capacity(estimated_capacity);

    for (i, id) in self.schema_ids.iter().enumerate() {
        if i > 0 {
            result.push(';');
        }
        let md5_hash = md5::compute(id.to_le_bytes());
        write!(&mut result, "{md5_hash:x}").unwrap();
    }

    result
}
```

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/central_blob.rs
```diff
@@ -1,8 +1,96 @@
...
+                .single()
+                .unwrap_or_else(Utc::now)
+        } else {
+            Utc::now()
+        }
+    }
+
+    /// Format DateTime as ISO 8601 with 7-digit precision (.NET compatible)
+    fn format_datetime(datetime: DateTime<Utc>) -> String {
+        format!(
```
utpilla, line 46: I believe it would be more performant to use `format` method here than making size separate method calls to extract year, month etc. and using to `format!` to process them:

```rust
datetime.format("%Y-%m-%dT%H:%M:%S%.7fZ").to_string()
```

lalitb: yes, I tried it earlier as it's existing code - there were some issue. For now, I have added the TODO to explore this further

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/central_blob.rs
```diff
@@ -1,8 +1,92 @@
...
 
+/// Metadata for a batch of events
+#[derive(Debug, Clone)]
+pub struct BatchMetadata {
+    /// Start time of the earliest event in nanoseconds since Unix epoch
+    pub start_time: u64,
+    /// End time of the latest event in nanoseconds since Unix epoch
+    pub end_time: u64,
+    /// List of schema IDs present in this batch
+    pub schema_ids: Vec<u64>,
```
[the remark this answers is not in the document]
lalitb: Great catch, @utpilla! This is an excellent optimization. The string formatting approach is much cleaner and more efficient. Thanks for the detailed suggestion with the code example :)

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -44,77 +43,116 @@ impl OtlpEncoder {
...
+            metadata: BatchMetadata,
+        }
+
+        impl BatchData {
+            fn format_schema_ids(&self) -> String {
+                use std::fmt::Write;
+                self.schemas
+                    .iter()
+                    .enumerate()
+                    .fold(String::new(), |mut acc, (i, s)| {
```
utpilla, line 66: We could pre calculate the string capacity to optimize it further.

utpilla (approved):

PR #398, opened 2025-07-29
feat: [Geneva Exporter] Split Client API into batching and uploading

lalitb:
## Changes

This PR refactors the Geneva client API to separate the encoding/compression phase from the upload phase, providing users with more flexibility and control over the upload process.

### Before
```rust
pub async fn upload_logs(&self, logs: &[ResourceLogs]) -> Result<(), String>
```

The previous API combined encoding, compression, and uploading into a single method, limiting user control over the process.

### After

**1. Encode and Compress Logs (Synchronous)**
```rust
pub fn encode_and_compress_logs(
    &self,
    logs: &[ResourceLogs],
) -> Result<Vec<CompressedBatch>, String>
```

**2. Upload Individual Batches (Asynchronous)**
```rust
pub async fn upload_batch(&self, batch: &CompressedBatch) -> Result<(), String>
```

## Benefits

- Users can now store `CompressedBatch` objects to disk or memory before upload. 
- Each batch has a separate upload API request
- Failed batches can be retried with exponential backoff or other strategies
- Users control how async upload tasks are scheduled

### DIFF
--- a/opentelemetry-user-events-logs/src/logs/processor.rs
+++ b/opentelemetry-user-events-logs/src/logs/processor.rs
@@ -36,7 +36,7 @@ where
 
 impl Processor<DefaultEventNameCallback> {
     /// Creates a builder for configuring a user_events Processor
-    pub fn builder(provider_name: &str) -> ProcessorBuilder {
+    pub fn builder(provider_name: &str) -> ProcessorBuilder<'_> {
         ProcessorBuilder::new(provider_name)
     }
 }


opentelemetry-exporter-geneva/opentelemetry-exporter-geneva/src/logs/exporter.rs
```diff
@@ -1,25 +1,33 @@
...
     _is_shutdown: atomic::AtomicBool,
-    geneva_client: GenevaClient,
+    geneva_client: Arc<GenevaClient>,
+    max_concurrent_uploads: usize,
 }
 
 impl GenevaExporter {
     /// Create a new GenavaExporter
     pub fn new(geneva_client: GenevaClient) -> Self {
+        Self::with_concurrency(geneva_client, 10) // Default to 10 concurrent uploads
```
utpilla, line 21: Either use `new_with_concurrency` as the method name or consider using a builder pattern? We could consider using builder pattern if we think we would have more options required later.

```suggestion
        Self::new_with_concurrency(geneva_client, 10) // Default to 10 concurrent uploads
```

opentelemetry-exporter-geneva/opentelemetry-exporter-geneva/src/logs/exporter.rs
```diff
@@ -1,25 +1,33 @@
...
+    }
+
+    /// Create a new GenavaExporter with custom concurrency level
+    pub fn with_concurrency(geneva_client: GenevaClient, max_concurrent_uploads: usize) -> Self {
```
utpilla, line 25: Either use `new_with_concurrency` as the method name or consider using a builder pattern? We could consider using builder pattern if we think we would have more options required later.

```suggestion
    pub fn new_with_concurrency(geneva_client: GenevaClient, max_concurrent_uploads: usize) -> Self {
```

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -78,57 +83,64 @@ impl GenevaClient {
...
-            .collect()
-            .await;
-
-        // Return error if any uploads failed
-        if !errors.is_empty() {
-            return Err(format!("Upload failures: {}", errors.join("; ")));
+        let encoded_batches = self.encoder.encode_log_batch(log_iter, &self.metadata);
+
+        // Pre-allocate with exact capacity to avoid reallocations
+        let mut compressed_batches = Vec::with_capacity(encoded_batches.len());
```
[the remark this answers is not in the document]
lalitb: good point. done.

stress/src/geneva_exporter.rs
```diff
@@ -290,7 +295,33 @@ async fn async_main(
...
+                    // All batch uploads happen within the same async task that called `encode_and_compress_logs`
+                    // Note: The batches are uploaded concurrently 10 at a time, however the test setup currently doesn't have
+                    // multiple concurrent operations.
+                    use futures::stream::{self, StreamExt};
+                    let errors: Vec<String> = stream::iter(batches)
+                        .map(|batch| {
+                            let client = client.clone();
+                            async move { client.upload_batch(&batch).await }
+                        })
+                        .buffer_unordered(10) // Max 10 concurrent uploads per operation
```
utpilla, line 312: Why use only `10` here?

stress/src/geneva_exporter.rs
```diff
@@ -290,7 +295,33 @@ async fn async_main(
...
+                async move {
+                    let batches = client
+                        .encode_and_compress_logs(&logs)
+                        .map_err(std::io::Error::other)?;
+
+                    // All batch uploads happen within the same async task that called `encode_and_compress_logs`
+                    // Note: The batches are uploaded concurrently 10 at a time, however the test setup currently doesn't have
+                    // multiple concurrent operations.
+                    use futures::stream::{self, StreamExt};
+                    let errors: Vec<String> = stream::iter(batches)
```
utpilla, line 307: This seems odd. We now have a stream of futures which further get used to produce another stream of futures in the `run_continuous` method. I think we need to fix the test.

utpilla (approved):
The `src` related changes look good.

I have some questions about the stress test changes. I'm not sure we are making the right changes to stress test in this PR. If it's not a problem, I would suggest commenting out the stress test related changes in this PR so that it can be reviewed again later.  Approving to unblock progress on the FFI work.

lalitb:
> I have some questions about the stress test changes. I'm not sure we are making the right changes to stress test in this PR. If it's not a problem, I would suggest commenting out the stress test related changes in this PR so that it can be reviewed again later. Approving to unblock progress on the FFI work.

The motive behind streams within streams was to  parallelize batch uploads within each concurrent task. But yes, this may be confusing, or not the right approach. For now, I have reverted back the change  - the only way to do it was to make the upload for each batch sequential within the concurrent batch. Which should be actually fine in the current test setup, as there is only single batch created per task.
Have added TODO to handle this separately, and will create issue for that too. Let me know if case there are still concerns. Will handle them in separate PR. Thanks for reviewing this.

PR #420, opened 2025-08-06
feat: auto listener benchmark

### LINKED ISSUE #369: Benchmarks for user-events should automatically enabled and disable listener
This is an improvement to the benches in
https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-user-events-logs/benches/logs.rs

The behavior depends on whether a user-events listener is present or not. There is a code comment that shows how to enable/disable the listener.

Currently it's manually enabled/disabled and the tests are run and the results stored as comment based on that. 

Task: Improve the benchmark, so that it enables/disables the user-event using the commands (echo..), and run the benches with enabled/disabled.

Modify the results (comment) with placeholder results.

artemziborev:
Fixes #369

## Changes

This PR improves the user-events logs benchmark by adding automatic enable/disable management for the user events listener using RAII guards. The benchmark now automatically enables the listener when tests start and disables it when they complete, ensuring proper cleanup even if the benchmark panics.

### Key Changes:
- Added `UserEventsListenerGuard` RAII guard for automatic listener management
- Updated benchmark results in comments with current performance data
- Added graceful fallback for systems without user events support
- Enabled required features by default to fix code analysis warnings

### Technical Details:
- Uses `echo` commands to enable/disable listener: `/sys/kernel/debug/tracing/events/user_events/myprovider_L2K1/enable`
- Preserves listener state (doesn't disable if already enabled)
- Provides detailed logging for debugging
- Handles systems without user events support gracefully

* [x] Unit tests added/updated (if applicable) - All existing tests pass
* [x] Changes in public API reviewed (if applicable) - No public API changes


opentelemetry-user-events-logs/benches/logs.rs
```diff
@@ -97,6 +225,24 @@ where
...
 
 fn benchmark_4_attributes(c: &mut Criterion) {
+    // Check if user events are available
+    if let Err(e) = UserEventsListenerGuard::check_user_events_available() {
+        eprintln!("Warning: User events not available: {}", e);
+        eprintln!("Benchmarks will run without listener enabled");
+    }
+    
+    // Enable listener with RAII guard
+    let _guard = UserEventsListenerGuard::enable("myprovider")
```
cijothomas, line 235: I don't think this works as the tracepoint name will not exist at this point! This should be done after the provider is built, which will build the user_events exporter, which will create the tracepoint names.

opentelemetry-user-events-logs/Cargo.toml
```diff
@@ -33,7 +33,7 @@ serde_json = "1.0.140"
 spec_unstable_logs_enabled = ["opentelemetry/spec_unstable_logs_enabled", "opentelemetry_sdk/spec_unstable_logs_enabled", "opentelemetry-appender-tracing/spec_unstable_logs_enabled"]
 internal-logs = ["tracing", "opentelemetry/internal-logs", "opentelemetry_sdk/internal-logs"]
 experimental_eventname_callback = []
-default = ["internal-logs"]
+default = ["internal-logs", "spec_unstable_logs_enabled", "experimental_eventname_callback"]
```
cijothomas, line 36: these should not be enabled by default.

opentelemetry-user-events-logs/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 ## vNext
 
+### Changed
```
cijothomas, line 5: changelog is only for end user facing changes, so this can be reverted.

artemziborev: Removed non end-user-facing (benchmark/internal) entries from vNext to keep the changelog user-focused.

opentelemetry-user-events-logs/Cargo.toml
artemziborev: Removed experimental/spec flags from defaults. Now: default = ["internal-logs"].
Benches continue to opt-in via required-features.

opentelemetry-user-events-logs/benches/logs.rs
artemziborev: Moved UserEventsListenerGuard::enable(PROVIDER_NAME) to occur after building the provider in all benchmarks, ensuring tracepoints exist first.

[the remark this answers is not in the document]
artemziborev: Introduced PROVIDER_NAME constant and replaced all "myprovider" usages (builders and UserEventsListenerGuard::enable(...)).

opentelemetry-user-events-logs/benches/logs.rs
```diff
@@ -35,6 +40,129 @@ use opentelemetry_user_events_logs::Processor;
...
+        
+        // Only disable if it wasn't already enabled
+        if !self.was_enabled {
+            let disable_output = Command::new("sudo")
+                .arg("sh")
+                .arg("-c")
+                .arg(format!("echo 0 | sudo tee {}", disable_path))
+                .output();
+            
+            match disable_output {
```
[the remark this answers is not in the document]
artemziborev: Implemented safe construction: replaced sh -c "echo X | sudo tee …" with sudo tee and piped X\n via stdin.
Applied to both enable and disable paths; retained output handling; lints pass.

hdost (approved):

lalitb (approved):

cijothomas (changes requested):
Thanks for working on this. Requesting changes to address https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/420/files#r2273844603

opentelemetry-user-events-logs/benches/logs.rs
```diff
@@ -97,7 +270,20 @@ where
...
 fn benchmark_4_attributes(c: &mut Criterion) {
+    // Check if user events are available
+    if let Err(e) = UserEventsListenerGuard::check_user_events_available() {
+        eprintln!("Warning: User events not available: {}", e);
+        eprintln!("Benchmarks will run without listener enabled");
+    }
+
     let provider = setup_provider_default();
+    // Enable listener with RAII guard (after provider is built so tracepoints exist)
+    let _guard = UserEventsListenerGuard::enable(PROVIDER_NAME).unwrap_or_else(|e| {
```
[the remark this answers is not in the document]
artemziborev: Thank you for the excellent suggestion! You're absolutely right - the current approach always enables the listener, which doesn't provide meaningful performance comparison data.
I've restructured all benchmark functions to use criterion bench groups that run both with and without the listener enabled:
Each benchmark now creates a benchmark_group with "disabled" and "enabled" variants
The "disabled" case runs without any listener management (baseline performance)
The "enabled" case uses the RAII guard to enable/disable the listener
This provides proper comparison data showing the actual overhead of enabling user events
The changes maintain all the safety benefits (RAII guards, proper cleanup) while giving us the performance comparison we need. Users can now see the baseline performance vs. the overhead of having the listener active.
This addresses the core issue you identified - we now have proper benchmark data for both states rather than always running with the listener enabled.

artemziborev (commented):
All reviewer comments have been addressed:

✅ Fixed command injection risk by using `sudo tee` with piped stdin instead of shell commands
✅ Replaced hardcoded "myprovider" with `PROVIDER_NAME` constant  
✅ Extracted duplicated dummy guard creation into `dummy_guard()` helper method
✅ Added `USER_EVENTS_PROVIDER_SUFFIX` constant to eliminate magic string
✅ Moved listener enabling to occur after provider creation (tracepoints exist)
✅ Removed experimental features from default in Cargo.toml
✅ Restructured benchmarks to compare enabled/disabled listener performance using criterion groups
✅ Reverted non-user-facing changes from CHANGELOG

The benchmarks now properly compare performance with and without the listener enabled, providing meaningful data for both states.

cijothomas:
> Summary: reworked benchmark to use criterion group with two functions — disabled (baseline) and enabled (via RAII guard). The listener enable happens after provider init to ensure tracepoints exist. Added graceful skip when user_events unsupported. Why: provides apples-to-apples perf comparison of listener overhead, addressing prior review feedback. Local run: cargo bench -p opentelemetry-user-events-logs
> 
> @cijothomas could you please re-review/dismiss the outdated “changes requested”?

Sorry for the huge delay! I'll get back to this by Monday.

PR #431, opened 2025-08-25
feat: Add tower tracing

jan-xyz:
## Changes

This adds tracing support for the tower layer. It was addressed briefly with @cijothomas on the CNCF slack, but we didn't discuss any details of the implementation. I took the liberty for now to take 2 major design decisions:

1. General vs. Specialized tower layer: I chose to pick a general layer that does metrics and tracing together with a no-op tracer in case someone doesn't want to do tracing with this layer. I assume that long term people/teams will fully switch to OpenTelemetry for Metrics and Tracing and instead of specifying two layers which each inspect the request and response, it makes more sense to do this in one layer.

2. Backwards compatibility: I tried to leave the original types in place which have the "Metric" implementation in the name. This is maybe a more smooth migration? I am not sure how helpful that actually is, and maybe we should just break compatibility in case we want to go with a unified layer.

Note: It currently does not include context extraction.

As a follow up I would also like to provide a way to use this as a client layer, where the semantics change a bit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ opentelemetry-appender-tracing = "0.31"
 opentelemetry-http = "0.31"
 opentelemetry-proto = { version = "0.31", default-features = false }
 opentelemetry_sdk = { version = "0.31", default-features = false }
-opentelemetry-stdout = "0.31"
+opentelemetry-stdout = { version = "0.31", default-features = false }
 opentelemetry-semantic-conventions = { version = "0.31", features = [
     "semconv_experimental",
 ] }
--- a/opentelemetry-instrumentation-actix-web/CHANGELOG.md
+++ b/opentelemetry-instrumentation-actix-web/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## vNext
 
+### Changed
+
+* `http.server.duration` histogram now uses custom boundaries to comply with OpenTelemetry semantic conventions: `[0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1.0, 2.5, 5.0, 7.5, 10.0]` (in seconds)
+
 ## v0.24.0
 
 ### Changed
--- a/opentelemetry-instrumentation-actix-web/Cargo.toml
+++ b/opentelemetry-instrumentation-actix-web/Cargo.toml
@@ -42,7 +42,6 @@ opentelemetry-instrumentation-actix-web = { path = ".", features = [
   "awc",
 ] }
 opentelemetry_sdk = { workspace = true, features = [
-  "spec_unstable_metrics_views",
   "metrics",
 ] }
 opentelemetry-stdout = { workspace = true, features = ["trace", "metrics"] }
--- a/opentelemetry-instrumentation-actix-web/examples/client.rs
+++ b/opentelemetry-instrumentation-actix-web/examples/client.rs
@@ -34,17 +34,18 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
         .with_attribute(KeyValue::new("service.name", "actix_client"))
         .build();
 
-    let tracer = SdkTracerProvider::builder()
-        .with_batch_exporter(SpanExporter::default())
+    let tracer_provider = SdkTracerProvider::builder()
+        .with_simple_exporter(SpanExporter::default())
         .with_resource(service_name_resource)
         .build();
+    global::set_tracer_provider(tracer_provider.clone());
 
     let client = awc::Client::new();
     let response = execute_request(client).await?;
 
     println!("Response: {response}");
 
-    tracer.shutdown()?;
+    tracer_provider.shutdown()?;
 
     Ok(())
 }
--- a/opentelemetry-instrumentation-actix-web/examples/server.rs
+++ b/opentelemetry-instrumentation-actix-web/examples/server.rs
@@ -2,15 +2,17 @@ use actix_web::{web, App, HttpResponse, HttpServer, Responder};
 use opentelemetry::{global, KeyValue};
 use opentelemetry_instrumentation_actix_web::{RequestMetrics, RequestTracing};
 use opentelemetry_sdk::{
-    metrics::{Aggregation, Instrument, SdkMeterProvider, Stream},
-    propagation::TraceContextPropagator,
-    trace::SdkTracerProvider,
+    metrics::SdkMeterProvider, propagation::TraceContextPropagator, trace::SdkTracerProvider,
     Resource,
 };
 use opentelemetry_stdout::{MetricExporter, SpanExporter};
 
 async fn manual_hello() -> impl Responder {
     HttpResponse::Ok().body("Hey there!")
+}
+
+async fn get_user(user_id: web::Path<String>) -> impl Responder {
+    HttpResponse::Ok().body(format!("User ID: {}", user_id))
 }
 
 #[actix_web::main]
@@ -22,12 +24,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .with_attribute(KeyValue::new("service.name", "actix_server"))
         .build();
 
-    let tracer = SdkTracerProvider::builder()
+    let tracer_provider = SdkTracerProvider::builder()
         .with_simple_exporter(SpanExporter::default())
         .with_resource(service_name_resource)
         .build();
 
-    global::set_tracer_provider(tracer.clone());
+    global::set_tracer_provider(tracer_provider.clone());
 
     // Setup a OTLP metrics exporter if --features metrics is used
     #[cfg(feature = "metrics")]
@@ -39,24 +41,6 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                     .with_attribute(KeyValue::new("service.name", "my_app"))
                     .build(),
             )
-            .with_view(|i: &Instrument| {
-                if i.name() == "http.server.duration" {
-                    Some(
-                        Stream::builder()
-                            .with_aggregation(Aggregation::ExplicitBucketHistogram {
-                                boundaries: vec![
-                                    0.0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75,
-                                    1.0, 2.5, 5.0, 7.5, 10.0,
-                                ],
-                                record_min_max: true,
-                            })
-                            .build()
-                            .unwrap(),
-                    )
-                } else {
-                    None
-                }
-            })
             .build();
         global::set_meter_provider(provider.clone());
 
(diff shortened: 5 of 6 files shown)


opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -31,23 +35,23 @@ const _OTEL_DEFAULT_HTTP_SERVER_DURATION_BOUNDARIES: [f64; 14] = [
 const LIBRARY_DEFAULT_HTTP_SERVER_DURATION_BOUNDARIES: [f64; 14] = [
     0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0, 30.0, 60.0, 120.0, 300.0,
 ];
-const HTTP_SERVER_ACTIVE_REQUESTS_METRIC: &str = "http.server.active_requests";
+const HTTP_SERVER_ACTIVE_REQUESTS_METRIC: &str = semconv::metric::HTTP_SERVER_ACTIVE_REQUESTS;
```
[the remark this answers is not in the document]
cijothomas: change itself is good, just don't want it to get blocked by discussions around adding tracing support etc.!

opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -505,3 +590,172 @@ fn split_and_format_protocol_version(http_version: http::Version) -> (String, St
...
+    use opentelemetry_sdk::metrics::SdkMeterProvider;
+    use opentelemetry_sdk::trace::InMemorySpanExporterBuilder;
+    use opentelemetry_sdk::trace::SdkTracerProvider;
+    use std::result::Result;
+    use tower::Service;
+    use tower::ServiceBuilder;
+    use tower::ServiceExt;
+
+    #[tokio::test(flavor = "current_thread")]
+    async fn test_tracing_with_in_memory_tracer() {
```
cijothomas, line 610: Thanks for adding test coverage for both traces and metrics!
💯

opentelemetry-instrumentation-tower/src/lib.rs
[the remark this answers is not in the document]
jan-xyz: https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/435

opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -281,8 +302,13 @@ impl<ReqExt, ResExt> HTTPMetricsLayerBuilder<ReqExt, ResExt> {
         self
     }
 
-    fn make_state(meter: Meter, req_dur_bounds: Vec<f64>) -> HTTPMetricsLayerState {
-        HTTPMetricsLayerState {
+    pub fn with_tracer(mut self, tracer: BoxedTracer) -> Self {
```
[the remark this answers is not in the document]
cijothomas: libraries (and instrumentation libraries which does instrumentation) should always retrieve the global tracer/meter. It is upto application owner to ensure they set a valid meter/tracer using `global` - if they don't, NoOps get used.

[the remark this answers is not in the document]
jan-xyz: Ok, so that implies that I should set the global tracer in tests to the InMemory variants?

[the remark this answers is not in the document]
jan-xyz: I updated it to use global providers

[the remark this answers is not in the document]
jan-xyz: I re-added the functions to set the tracer and meter provider, because the tests are not well supported for the global setup. The default is now to use the global tracer and meter, but with the option to override it, which I also chose for testing. Otherwise, it's not possible to normally run the tests, and the main `opentelemetry-rust` repo even had all these tests disabled that used the global state. I am not sure if this will become better supported at some point, or you already have a plan for it? Otherwise, I think this is a good compromise.

jan-xyz:
while working on this, I started to wonder how the overall strategy for this crate could look like, and I am wondering if my PR is maybe not going far enough. For the future we have multiple use cases for a (collection of?) tower layer(s).

So far I see at least 4:
- HTTP Server
- HTTP Client
- gRPC Server
- gRPC Client

As it stands now, it provides HTTP server support, for that it might make sense to also rename the structs accordingly to `HTTPServerLayer` a.s.o. I won't push the other 3 concerns into this PR, but I at least would have the plan to add these in the next couple of days/weeks if it fits this project since I need this functionality for a project I am currently working on.

jan-xyz:
@cijothomas anything I can do to move this PR forward?

opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -249,41 +269,61 @@ impl<ReqExt, ResExt> HTTPMetricsLayerBuilder<ReqExt, ResExt> {
...
 
     pub fn with_request_duration_bounds(mut self, bounds: Vec<f64>) -> Self {
         self.req_dur_bounds = Some(bounds);
         self
     }
 
-    fn make_state(meter: Meter, req_dur_bounds: Vec<f64>) -> HTTPMetricsLayerState {
-        HTTPMetricsLayerState {
+    /// Set a meter provider to use for creating a meter.
+    /// If none is specified, the global provider is used.
```
cijothomas: lets just use the global provider always. If there is a concrete reason why global cannot work, then we can follow up and accept a provider. (If we ever need to do that, it just need to accept a tracer or meter. Not the provider itself.But thats for another time discussion)

opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -394,29 +437,65 @@ where
...
+            None => global::tracer("opentelemetry-instrumentation-tower"),
+        };
+
+        let span = tracer
+            .span_builder(span_name)
+            .with_kind(SpanKind::Server)
+            .with_attributes(span_attributes)
+            .start(&tracer);
+
+        let cx = OtelContext::current_with_span(span);
```
cijothomas: do we attach/activate the context? or just store it ourselves? Normally, this should be just activated, so any subsequent code can access it.

opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -447,30 +528,56 @@ where
...
-        label_superset.extend(this.metrics_state.custom_request_attributes.clone());
+        label_superset.extend(this.future_state.custom_request_attributes.clone());
 
         // Extract and add custom response attributes
         let custom_response_attributes = this.response_extractor.extract_attributes(&response);
-        label_superset.extend(custom_response_attributes);
+        label_superset.extend(custom_response_attributes.clone());
+
+        // Update span
+        let span = this.future_state.otel_context.span();
```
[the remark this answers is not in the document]
jan-xyz: That's not possible, I believe, or I am not sure how to solve it. The guard is a raw pointer (`*const()`) which means it is `!Send` but we have to store it for the response future to not drop it and lose the guard and for that it must be `Send`.

opentelemetry-instrumentation-tower/src/lib.rs
jan-xyz: https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/431#discussion_r2561682483

francoposa:
> This is a general problem of mixing open-telemetry and tracing and not specific to the open-telemetry-Axum instrumentation.

Those TRACE level logs are how the Tower ecosystem emits its trace spans, this log-tracing bridge is how we properly make the connection to the spans emitted here.

It’s a few extra lines to wire up, but that is how the Rust tracing ecosystem is for now, particularly the instrument macro. These OTel libraries enable us to correctly attach those spans from the same trace context together.

The intent of the Tower instrumentation library when I wrote it was to surface the telemetry from Tower. I doubt we want to intentionally ignore and exclude the traces and child spans created by Tower.

All it needs for now is the example code to wire it up. I am happy to add it myself.

opentelemetry-instrumentation-tower/CHANGELOG.md
```diff
@@ -2,6 +2,49 @@
 
 ## vNext
 
+### Changed
+
+* **BREAKING**: Removed `with_meter()` method. The middleware now uses global meter and tracer providers by default via `opentelemetry::global::meter()` and `opentelemetry::global::tracer()`, with optional overrides via `with_tracer_provider()` and `with_meter_provider()` methods.
```
francoposa: `with_meter` has not been removed yet

opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -249,27 +262,30 @@ impl<ReqExt, ResExt> HTTPMetricsLayerBuilder<ReqExt, ResExt> {
...
-            None => Err(Error {
-                inner: ErrorKind::Config(String::from("no meter provided")),
-            }),
-        }
+
+        let tracer = Arc::new(global::tracer("opentelemetry-instrumentation-tower"));
+
+        let meter: Meter = self
+            .meter
+            .unwrap_or_else(|| global::meter("opentelemetry-instrumentation-tower"));
```
francoposa: After `with_meter` gets removed, the tracer and meter init can be done the same way.

opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -190,9 +203,9 @@ impl fmt::Debug for Error {
     }
 }
 
-impl HTTPMetricsLayerBuilder {
+impl HTTPLayerBuilder {
     pub fn builder() -> Self {
-        HTTPMetricsLayerBuilder {
+        HTTPLayerBuilder {
             meter: None,
             req_dur_bounds: Some(LIBRARY_DEFAULT_HTTP_SERVER_DURATION_BOUNDARIES.to_vec()),
```
francoposa: Drive-by not caused by your PR, but we should probably default to the `_OTEL_DEFAULT_HTTP_SERVER_DURATION_BOUNDARIES` - this non-semantic-convention default is a holdover from when this was my own library.

cijothomas:
> I don't think the tracing bridge should be part of the PR. This is a general problem of mixing open-telemetry and tracing and not specific to the open-telemetry-Axum instrumentation. If you want to use the bridge it's good for you but I doubt that's the general sentiment.

+1. That problem is something we'd want to cover with good examples/doc. 

@francoposa While you modified sample shows all the spans created via `tracing:span`, we cannot determine whether end-users really wanted all those spans. Some users merely wanted extra attributes from those spans attached to log_record only (like in this PR is solving: https://github.com/open-telemetry/opentelemetry-rust/pull/3282/changes)

Lets park this topic to unblock progress, and get this changes in quickly?

cijothomas:
@jan-xyz There is a conflict on readme file. Could you resolve that - since you are actively helping on tower, it makes sense to list yourselves as co-owner for the same.

opentelemetry-instrumentation-tower/CHANGELOG.md
jan-xyz: Ah, I removed it and then re-added it because I couldn't make the tests work without it.

cijothomas (approved):
Thanks!

Issue #309, opened 2025-06-20
[Geneva Exporter] Uploader should handle retriable errors


utpilla:
We need to handle retires:

https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/8a29dec470defdec64f5989866021c372e998a49/opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs#L212-L224

utpilla:
@lalitb Closing this issue since we expect the callers to be responsible for retries. Related issue: #581

PR #442, opened 2025-09-02
fix: [Geneva Exporter] Add retries for ingestion info retrieval

lalitb:
towards:  #309 

## Summary
This PR adds retry logic to the Geneva Config Service client to improve resilience against transient network issues and server overload. The changes implement intelligent error handling with appropriate backoff strategies.

Key changes:

- Adds retry mechanism with up to 3 attempts for Geneva Config Service API calls
- Implements smart backoff strategy using server-provided Retry-After headers or fixed delays
- Introduces new error type for rate limiting scenarios with retry information
- No changes in upload API, as retry for upload failure is the responsibility of caller.

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
@@ -394,11 +394,8 @@ impl GenevaConfigClient {
                         break;
                     }
                     Err(e) => {
-                        last_error = Some(e);
-                        if attempt < MAX_GCS_RETRIES
-                            && is_retriable_error(last_error.as_ref().unwrap())
-                        {
-                            let delay = match last_error.as_ref().unwrap() {
+                        if attempt < MAX_GCS_RETRIES && is_retriable_error(&e) {
+                            let delay = match &e {
                                 GenevaConfigClientError::RequestFailedWithRetryAfter {
                                     retry_after_secs: Some(s),
                                     ..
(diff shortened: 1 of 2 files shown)


utpilla:
> fixes: #309

This PR doesn't fix that issue. It's only working on add retries for the ingestion info calls not the actual upload calls. We should either create a dedicate issue for this PR or use a different phrase such as "Towards" or "Partially Fixes" so that merging this PR doesn't automatically close the issue.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -370,10 +372,23 @@ impl GenevaConfigClient {
...
+        let (fresh_ingestion_gateway_info, fresh_moniker_info) = {
+            let mut attempt = 0;
+            loop {
+                match self.fetch_ingestion_info().await {
+                    Ok(v) => break v,
+                    Err(e) => {
+                        if attempt < MAX_GCS_RETRIES && is_retriable_error(&e) {
+                            attempt += 1;
+                            continue;
+                        } else {
```
utpilla, line 385: We need to respect the [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After) header if present, before sending the request again.

lalitb:
> This PR doesn't fix that issue. It's only working on add retries for the ingestion info calls not the actual upload calls. We should either create a dedicate issue for this PR or use a different phrase such as "Towards" or "Partially Fixes" so that merging this PR doesn't automatically close the issue.

Agree, done.

PR #451, opened 2025-09-24
feat(geneva): add Managed Identity authentication support

gscalderon:
### Summary
Adds Managed Identity (MSI) authentication support to the Geneva exporter, enabling both system-assigned and user-assigned identities. Improves error diagnostics for MSI token acquisition and cleans up examples / formatting.

### Details
- New `AuthMethod::ManagedIdentity { selector }` variant.
- `ManagedIdentitySelector` supports: `System`, `ObjectId(String)`, `ClientId(String)`, `ResourceId(String)`.
- Token acquisition tries multiple scope variants (`/.default`, base, base with trailing slash) and reports candidate scopes + last error upon failure.
- Fixed Geneva Config Service URL path (trailing slash) to satisfy tests.
- Removed debug scaffolding, unused code, and applied consistent formatting.
- Added `basic_msi_test` example demonstrating environment variable setup for MSI.


README.md
```diff
@@ -29,7 +29,7 @@ Check individual folders for usage guidelines and examples.
 ## Supported Rust Versions
 
 OpenTelemetry is built against the latest stable release. The minimum supported
-version is 1.75. The current OpenTelemetry version is not guaranteed to build
+version is 1.85. The current OpenTelemetry version is not guaranteed to build
```
lalitb, line 32: Don't update the compiler requirements for all the packages within this repo. Only do it for `opentelemetry-geneva-exporter/*/`

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -44,6 +48,19 @@ use std::sync::RwLock;
 /// ```powershell
 /// openssl pkcs12 -export -in cert.pem -inkey key.pem -out client.p12 -name "alias"
 /// ```
+#[allow(dead_code)]
+#[derive(Clone, Debug)]
+pub enum ManagedIdentitySelector {
+    /// System-assigned managed identity.
+    System,
+    /// User-assigned by Object (Principal) Id (GUID)
+    ObjectId(String),
```
lalitb, line 57: If the VM has only one user-assigned MSI attached, do we still need or should provide the object-id or client-id - or will the azure-sdk auto-detect the identity and use?

gscalderon: Currently, users need to provide either the object-id or client-id, even if there is only one user-assigned MSI attached. The only exception is for system-assigned MSIs, where these values are not required

opentelemetry-exporter-geneva/geneva-uploader-ffi/Cargo.toml
```diff
@@ -5,7 +5,7 @@ version = "0.2.0"
 edition = "2021"
 homepage = "https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-exporter-geneva/geneva-uploader-ffi"
 repository = "https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-exporter-geneva/geneva-uploader-ffi"
-rust-version = "1.75.0"
+rust-version = "1.85.0"
```
lalitb, line 8: I believe this is un-avoidable, as rust-sdk-for-azure needs this minimal version -  https://github.com/Azure/azure-sdk-for-rust/blob/bbb303181bcd3492f2595777ab822dc889adc493/Cargo.toml#L38?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -70,15 +59,100 @@ pub enum AuthMethod {
...
     ///
-    /// Uses either the system-assigned identity (selector = System) or a user-assigned identity
-    /// specified by object id, client id, or resource id.
-    ManagedIdentity { selector: ManagedIdentitySelector },
+    /// At runtime we validate that exactly one of client_id|object_id|resource_id is provided.
+    UserManagedIdentity {
+        client_id: Option<String>,
+        object_id: Option<String>,
+        resource_id: Option<String>,
+    },
```
lalitb, line 71: Why not model client_id, object_id, and resource_id as an enum as suggested earlier since only one of them should ever be set at a time?

gscalderon: I wasn't sure if this approach was better since it avoids having all the models in the enum, but you're right. I'll update it so it's single and straightforward, as you suggested earlier

lalitb: Yes, the issue with the current approach is that it allocates space for all three fields even when only one of them is actually used.

PR #486, opened 2025-11-10
fix: [Geneva Uploader] Eliminate hash collisions and redundant MD5 computation in schema deduplication - Perf improvement 5-10%

lalitb:
## Changes

  Refactors schema deduplication to use exact field comparison instead of hash-based equality, eliminating collision risk and improving performance by 5-11%.

  - Replace hash-derived schema IDs with auto-incrementing IDs (1, 2, 3...)
  - Use exact field equality check (name + type) for schema deduplication instead of hash comparison
  - Compute MD5 hash once when schema is created and reuse for URL formatting
  - Store fields vector in `CentralSchemaEntry` for comparison (minimal memory overhead) - This can be avoided by encoding the schema for each event, and then compare it with already encoded unique events, this would be slower. It's a tradeoff between memory and CPU.

 Benchmark:

  The performance improvement comes from eliminating redundant MD5 computation - previously calculated twice (once for deduplication, once for URL parameter), now calculated once and reused.

  | Test         | Main (OLD) | PR (NEW)  | Improvement |
  |--------------|------------|-----------|-------------|
  | 0 attrs      | 6.58 µs    | 5.66 µs   | -14.0% ✅    |
  | 4 attrs      | 8.37 µs    | 7.50 µs   | -10.4% ✅    |
  | 8 attrs      | 10.87 µs   | 9.77 µs   | -10.1% ✅    |
  | 16 attrs     | 16.01 µs   | 14.32 µs  | -10.6% ✅    |
  | Batch 1      | 4.04 µs    | 3.72 µs   | -7.9% ✅     |
  | Batch 10     | 8.60 µs    | 7.72 µs   | -10.2% ✅    |
  | Batch 100    | 51.20 µs   | 44.55 µs  | -13.0% ✅    |
  | Batch 1000   | 473.51 µs  | 450.09 µs | -4.9% ✅     |
  | Mixed events | 57.92 µs   | 52.64 µs  | -9.1% ✅     |

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/bond_encoder.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/bond_encoder.rs
@@ -107,18 +107,18 @@ pub(crate) struct FieldDef {
 
 /// Schema definition that can be built dynamically
 #[derive(Clone)]
-pub(crate) struct DynamicSchema {
+pub(crate) struct DynamicSchema<'a> {
     pub struct_name: String,
     pub qualified_name: String,
-    pub fields: Vec<FieldDef>,
+    pub fields: &'a [FieldDef],
 }
 
-impl DynamicSchema {
-    pub(crate) fn new(name: &str, namespace: &str, fields: &[FieldDef]) -> Self {
+impl<'a> DynamicSchema<'a> {
+    pub(crate) fn new(name: &str, namespace: &str, fields: &'a [FieldDef]) -> Self {
         Self {
             struct_name: name.to_string(),
             qualified_name: format!("{namespace}.{name}"),
-            fields: fields.to_vec(),
+            fields,
         }
     }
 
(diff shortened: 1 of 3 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/bond_encoder.rs
```diff
@@ -114,11 +114,11 @@ pub(crate) struct DynamicSchema {
 }
 
 impl DynamicSchema {
-    pub(crate) fn new(name: &str, namespace: &str, fields: Vec<FieldDef>) -> Self {
+    pub(crate) fn new(name: &str, namespace: &str, fields: &[FieldDef]) -> Self {
         Self {
             struct_name: name.to_string(),
             qualified_name: format!("{namespace}.{name}"),
-            fields,
+            fields: fields.to_vec(),
```
utpilla: This cloning is unnecessary. We could use lifetimes for `DynamicSchema` struct. It would be contained and not cascade to other types.

```rust
#[derive(Clone)]
pub(crate) struct DynamicSchema<'a> {
    pub struct_name: String,
    pub qualified_name: String,
    pub fields: &'a [FieldDef],
}
```

PR #487, opened 2025-11-10
feat: Add opentelemetry initial declarative configuration

andborja:
## Changes

Adds initial declarative configuration support.

### DIFF
--- a/opentelemetry-config/examples/application/src/main.rs
+++ b/opentelemetry-config/examples/application/src/main.rs
@@ -7,8 +7,8 @@ pub mod mock_provider;
 pub mod model;
 use std::env;
 
-use opentelemetry_config::{providers::TelemetryProviders, ConfigurationProviderRegistry};
 use crate::mock_provider::MockPeriodicExporterProvider;
+use opentelemetry_config::{providers::TelemetryProviders, ConfigurationProviderRegistry};
 
 pub fn main() -> Result<(), Box<dyn std::error::Error>> {
     let args: Vec<String> = env::args().collect();
@@ -98,7 +98,10 @@ fn initialize_telemetry_registry() -> ConfigurationProviderRegistry {
     let mut registry = ConfigurationProviderRegistry::default();
 
     // Register the custom exporter factory.
-    MockPeriodicExporterProvider::register_into(&mut registry);
+    registry.register_metric_exporter_factory(
+        "custom",
+        MockPeriodicExporterProvider::register_mock_exporter_factory,
+    );
 
     registry
 }
@@ -108,6 +111,6 @@ fn configure_telemetry(
     registry: &ConfigurationProviderRegistry,
     telemetry_config: &str,
 ) -> Result<TelemetryProviders, Box<dyn std::error::Error>> {
-    let providers = TelemetryProviders::configure_from_yaml(&registry, telemetry_config)?;
+    let providers = TelemetryProviders::configure_from_yaml_str(&registry, telemetry_config)?;
     Ok(providers)
 }
--- a/opentelemetry-config/src/lib.rs
+++ b/opentelemetry-config/src/lib.rs
@@ -232,7 +232,7 @@ mod tests {
         let factory_function_option = registry.metrics.provider_factory(name);
         if let Some(factory_function) = factory_function_option {
             let builder = MeterProviderBuilder::default();
-            _ = factory_function(builder, &periodic_config_yaml).unwrap();
+            _ = factory_function(builder, periodic_config_yaml).unwrap();
             // Verify that the factory function was called
             assert_eq!(call_count.get(), 1);
         } else {
--- a/opentelemetry-config/src/providers.rs
+++ b/opentelemetry-config/src/providers.rs
@@ -25,7 +25,7 @@ pub struct TelemetryProviders {
 }
 
 impl TelemetryProviders {
-    pub fn new(
+    fn new(
         meter_provider: Option<SdkMeterProvider>,
         logs_provider: Option<SdkLoggerProvider>,
         traces_provider: Option<SdkTracerProvider>,
@@ -90,7 +90,7 @@ impl TelemetryProviders {
     }
 
     /// Configures the Telemetry providers from a YAML string
-    pub fn configure_from_yaml(
+    pub fn configure_from_yaml_str(
         configuration_registry: &ConfigurationProviderRegistry,
         yaml_str: &str,
     ) -> Result<TelemetryProviders, ProviderError> {
@@ -202,7 +202,7 @@ mod tests {
         let name = "console";
         registry.register_metric_exporter_factory(name, register_mock_reader_factory);
 
-        let providers = TelemetryProviders::configure_from_yaml(&registry, yaml_str).unwrap();
+        let providers = TelemetryProviders::configure_from_yaml_str(&registry, yaml_str).unwrap();
         assert!(providers.meter_provider.is_some());
     }
 
--- a/opentelemetry-config/examples/application/src/mock_provider.rs
+++ b/opentelemetry-config/examples/application/src/mock_provider.rs
@@ -32,15 +32,6 @@ struct CustomData {
 pub(crate) struct MockPeriodicExporterProvider {}
 
 impl MockPeriodicExporterProvider {
-    pub fn register_into(
-        registry: &mut opentelemetry_config::ConfigurationProviderRegistry,
-    ) {
-        registry.register_metric_exporter_factory(
-            "custom",
-            MockPeriodicExporterProvider::register_mock_exporter_factory,
-        );
-    }
-
     pub fn register_mock_exporter_factory(
         mut meter_provider_builder: MeterProviderBuilder,
         periodic_config_yaml: &str,
--- a/opentelemetry-config/examples/custom/src/main.rs
+++ b/opentelemetry-config/examples/custom/src/main.rs
@@ -50,7 +50,7 @@ pub fn main() -> Result<(), Box<dyn std::error::Error>> {
     MockPeriodicExporterProvider::register_into(&mut registry);
 
     // Configure telemetry from the provided YAML file.
-    let providers = match TelemetryProviders::configure_from_yaml(&registry, &config_yaml) {
+    let providers = match TelemetryProviders::configure_from_yaml_str(&registry, &config_yaml) {
         Ok(providers) => providers,
         Err(e) => {
             println!(


opentelemetry-config/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+/// Trait for configuring OTLP Exporter for Periodic Metrics Reader
+pub trait MetricsReadersPeriodicExporterOtlpConfigurator {
+    fn configure(
+        &self,
+        meter_provider_builder: MeterProviderBuilder,
+        config: &PeriodicExporterOtlp,
+    ) -> MeterProviderBuilder;
+}
```
utpilla: Are these traits needed?

opentelemetry-config-stdout/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+
+#[derive(Clone)]
+pub struct ConsolePeriodicExporterConfigurator {}
+
+impl ConsolePeriodicExporterConfigurator {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    pub fn register_into(configurators_manager: &mut opentelemetry_config::ConfiguratorManager) {
```
utpilla: This method seems odd. This way of registering components does not enforce an implementor to have the registration logic. I was hoping to see something similar to what @lalitb had suggested here: https://github.com/open-telemetry/opentelemetry-rust/pull/3226#discussion_r2505801811

Use a trait and force the implementors to provide the registration logic.

opentelemetry-config/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+mod tests {
+    use std::sync::atomic::AtomicI16;
+
+    use super::*;
+
+    #[test]
+    fn test_register_periodic_exporter_configurator() {
+        // Arrange
+        struct MockPeriodicExporterConfigurator {
+            call_count: AtomicI16,
```
utpilla: These are single threaded tests. We don't need to use atomics here. You might want to use `Cell` instead.

opentelemetry-config/src/model/metrics/reader.rs
```diff
@@ -0,0 +1,171 @@
...
+
+#[derive(serde::Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct Periodic {
+    pub exporter: Option<PeriodicExporter>,
+}
+
+#[derive(serde::Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct PeriodicExporter {
```
lalitb: It seems `deny_unknown_fields` will block adding any custom periodic exporters. Something like this won't work:
```yaml
          metrics:
            readers:
              - periodic:
                  exporter:
                    genevaexporter:
                      endpoint: "http://localhost:14268"
                      service_name: "my-service"
```

opentelemetry-config/src/model/metrics/reader.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub console: Option<PeriodicExporterConsole>,
+    pub otlp: Option<PeriodicExporterOtlp>,
+}
+
+#[derive(serde::Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct PeriodicExporterConsole {
+    pub temporality: Option<Temporality>,
+}
```
lalitb: This seems to be design issue to me:
- `opentelemetry-config` defines `PeriodicExporterConsole` (model)
-  `opentelemetry-config` validates YAML against this model.
-  But `opentelemetry-config` does not implement how to use this config.
- Instead, `opentelemetry-config-stdout` provides the implementation in a separate crate.

And possibly same pattern would be done for OTLP. I think the config model and implementation should be together - better to have them within `opentelemetry-config` for both OTLP and Console/stdout.

opentelemetry-config/src/lib.rs
andborja: No. Good catch!. It was old code. Removing it.

opentelemetry-config-stdout/src/lib.rs
lalitb: @andborja - Can you add a working example of an external crate adding a custom exporter (i.e, new exporter different from otlp/console to the YAML). I still have doubts on how this register pattern will work for external crates.

opentelemetry-config/src/model/metrics/reader.rs
andborja: It is actually by design: "Centralized configuration model and decoupled implementations".

I documented it at the readme:
https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/1d84ea6f8b2f5118152b9540fd05ebb1f5be5849/opentelemetry-config/README.md#design-pattern

opentelemetry-config-stdout/src/lib.rs
lalitb: I might be missing some context, but `MockConsoleExporter` doesn’t seem to demonstrate external custom exporter extensibility - it looks like an alternative implementation of the existing `PeriodicExporterConsole` type. That’s an implementation replacement rather than a type extension.

The registry pattern is interesting, but if it currently can’t support external crates adding new exporter types (as mentioned [here](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/487#discussion_r2513105623)), its value might be limited. In that case, it may be cleaner to merge the OTLP and stdout implementations directly into the main `opentelemetry-config` crate and simplify the design instead of maintaining a registry layer.

lalitb: I don't think this approach is scalable. Under the current design, every external exporter requires modifying the `PeriodicExporter` struct in `opentelemetry-config`. This makes us the gatekeeper for all exporter types. For a Geneva exporter, Jaeger exporter, or any other custom exporter, teams would need to:

1. Submit a PR adding their config struct to `opentelemetry-config`.
2. Wait for review and merge.
3. Have their schema maintained in this crate.

External crates should be self-contained - they should define their config types, implement the configurator trait, and work without touching opentelemetry-config source code.

andborja: Updated the implementation so the model is extensible also.

andborja: As discussed, implemented as a function instead.

opentelemetry-config-stdout/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+### 2. Load and Apply Configuration
+
+```rust
+use opentelemetry_config::{ConfigurationProvidersRegistry, providers::TelemetryProvider};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Create configuration registry and register console exporter
+    let mut registry = ConfigurationProvidersRegistry::new();
+    let metrics_registry = registry.metrics_mut();
```
lalitb: Is `metrics_mut` avoidable by directly providing `register_metric_config`?

opentelemetry-config/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+impl MetricsProvidersRegistry {
+    pub fn new() -> Self {
+        Self {
+            periodic_exporter_factories: HashMap::new(),
+        }
+    }
+
+    pub fn register_periodic_exporter_factory(
+        &mut self,
+        name: String,
```
lalitb: The exporter name would be static string - we can make it `&'static str` or `&srr` ?

opentelemetry-config/src/model/metrics/reader.rs
```diff
@@ -0,0 +1,150 @@
...
+        } else {
+            Err(serde::de::Error::custom("Empty map"))
+        }
+    }
+}
+
+#[derive(serde::Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct Periodic {
+    pub exporter: serde_yaml::Value,
```
lalitb: Add TODO to support missing fields -  interval_millis, timeout_millis, etc.

opentelemetry-config/src/providers.rs
```diff
@@ -0,0 +1,230 @@
...
+                if let Some(i) = n.as_i64() {
+                    KeyValue::new(key.to_string(), i)
+                } else if let Some(f) = n.as_f64() {
+                    KeyValue::new(key.to_string(), f)
+                } else {
+                    KeyValue::new(key.to_string(), n.to_string())
+                }
+            }
+            serde_yaml::Value::Bool(b) => KeyValue::new(key.to_string(), *b),
+            _ => KeyValue::new(key.to_string(), format!("{:?}", value)),
```
lalitb: add TODO to support Array - currently it would be stringified.

opentelemetry-config-stdout/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## vNext
+
+## v0.1.0
+
+### Added
+
+- Initial declarative configuration for stdout (console)
```
utpilla:
```suggestion
- Initial declarative configuration for stdout (console) exporter for metrics.
```

opentelemetry-config-stdout/README.md
utpilla: We don't need this indirection and the extra set of public APIs such as `metrics_mut()`, `traces_mut()` etc.

> The initial one is for periodic exporters, but there will be more.

We shouldn't have APIs specific to adding exporters since that's not enforced by our contract. We only need three public APIs for registration:
- `register_metrics_config()
- `register_traces_config()
- `register_logs_config()

opentelemetry-config/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+impl ConfigurationProvidersRegistry {
+    pub fn new() -> Self {
+        Self {
+            metrics: MetricsProvidersRegistry::new(),
+        }
+    }
+
+    pub fn metrics_mut(&mut self) -> &mut MetricsProvidersRegistry {
```
utpilla: This method is not needed for the reason mentioned in https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/487#discussion_r2532499511

opentelemetry-config/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+    trace::SdkTracerProvider,
+};
+use serde_yaml::Value;
+
+pub mod model;
+pub mod providers;
+
+/// Registry for different configuration providers.
+pub struct ConfigurationProvidersRegistry {
+    metrics: MetricsProvidersRegistry,
```
utpilla: 1. Let's use the name `MeterProvider` to be consistent with the term that the SDK uses.
2. Use the singular form of the word since there is only one provider that we configure.

```suggestion
    metrics: MeterProviderRegistry,
```

opentelemetry-config/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+impl MetricsProvidersRegistry {
+    pub fn new() -> Self {
+        Self {
+            periodic_exporter_factories: HashMap::new(),
+        }
+    }
+
+    pub fn register_periodic_exporter_factory(
```
utpilla: For the reasons mentioned in https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/487#discussion_r2532499511

```suggestion
    pub fn register_metrics_config(
```

opentelemetry-config/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+impl Default for ConfigurationProvidersRegistry {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Registry for metrics configuration providers.
+pub struct MetricsProvidersRegistry {
```
utpilla: We don't want this or any of its methods to be public.

opentelemetry-config-stdout/README.md
```diff
@@ -0,0 +1,128 @@
...
+    metrics_registry.register_periodic_exporter_factory(
+        "console".to_string(),
+        opentelemetry_config_stdout::register_console_exporter
+    );
+
+    // Load configuration from YAML file
+    let telemetry_provider = TelemetryProvider::new();
+    let providers = telemetry_provider
+        .configure_from_yaml_file(&registry, "otel-config.yaml")?;
```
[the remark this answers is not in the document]
utpilla: Yes. In fact, we should remove `TelemetryProviders::new()` method and just have a static method `TelemetryProviders::generate_from_yaml_file()`.

opentelemetry-config/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    pub fn with_traces_provider(mut self, traces_provider: SdkTracerProvider) -> Self {
+        self.traces_provider = Some(traces_provider);
+        self
+    }
+    pub fn with_logs_provider(mut self, logs_provider: SdkLoggerProvider) -> Self {
+        self.logs_provider = Some(logs_provider);
+        self
+    }
```
utpilla: All of these APIs are unnecessary.

opentelemetry-config/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+            ProviderError::RegistrationError(details) => {
+                write!(f, "Registration error: {}", details)
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::sync::{atomic::AtomicI16, Arc};
```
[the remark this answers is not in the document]
utpilla: Ah this is because of the `Send` and `Sync` trait bounds on the factory type. I think we should remove them since we only need to run this closure during initialization. We wouldn't need to run them in a multi-threaded or async context.

@lalitb Thoughts?

utpilla (changes requested):
We now seem to have a good approach that offers extensibility well. However, there are a lot of unnecessary APIs and a lot of APIs that shouldn't be public.

I don't think we need more public APIs than these for Metrics:

For `opentelemetry-config`
- `ConfigurationProvidersRegistry` (struct)
- `ConfigurationProvidersRegistry::new()`
- `ConfigurationProvidersRegistry::register_metrics_config()`
- `TelemetryProviders` (struct)
- `TelemetryProviders`::generate_from_yaml_file() -> Result<TelemetryProviders, ConfigurationError>`
- `ConfigurationError` (struct or enum that would be used an error type)
- `TelemetryProviders::meter_provider()` which can be used to retrieve the MeterProvider generated

For `opentelemetry-config-stdout`
- `register_console_exporter()`

opentelemetry-config/src/lib.rs
andborja: Sure.

opentelemetry-config/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+    pub fn metrics_mut(&mut self) -> &mut MetricsProvidersRegistry {
+        &mut self.metrics
+    }
+
+    pub fn metrics(&self) -> &MetricsProvidersRegistry {
+        &self.metrics
+    }
+}
+
+impl Default for ConfigurationProvidersRegistry {
```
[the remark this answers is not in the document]
andborja: I'll remove the ConfigurationProvidersRegistry struct.

opentelemetry-config/src/lib.rs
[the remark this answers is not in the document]
lalitb: Yes, it would be safe to assume that the configuration code would run in a single-threaded initialization context, and doesn't need Send + Sync safety - regardless of whether they are called from sync or async main.

opentelemetry-config/src/providers.rs
```diff
@@ -0,0 +1,230 @@
...
+            metrics_provider: MetricsProvider::new(),
+        }
+    }
+
+    /// Configures the Telemetry providers based on the provided configuration
+    pub fn configure(
+        &self,
+        configuration_registry: &ConfigurationProvidersRegistry,
+        config: &Telemetry,
+    ) -> Result<TelemetryProviders, ProviderError> {
```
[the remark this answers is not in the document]
andborja: Agree. Updated.

opentelemetry-config-stdout/README.md
lalitb: I'm not seeing the benefit of the `metrics_mut()` indirection for users. Can't we keep internal grouping while exposing simpler direct methods like `register_metric_periodic_reader()` instead of having users navigate through intermediate registry objects? Or maybe I am missing something basic in the discussion.

utpilla:
> * ConfigurationProviderRegistry is the composition of the other registries. It adds value to group the registries passed as parameter for configuration.

What additional value is added by `metrics_mut` method if the method used to register the configuration already has the word `metrics` in it and also requires a mutable reference `&mut` to `ConfigurationProviderRegistry`?

> * register_metrics_config cannot be a single method, as the contract is different for periodic and pull readers. It is reader based in the latest revision.

Why do we need different contracts? The only contract needed is access to `MeterProviderBuilder` and the associated yaml config data for the component. What's the benefit of moving away from a single `register_metrics_config` method approach?

> * We need the 3 functions, and the order of relevance is actually different:
>   generate_from(struct) > generate_from_yaml(string) > generate_from_yaml_file(file). The relevance order is because as this is a library, it will be linked from the host application configuration as a primarily use case. The configuration will not live in a separate file for most of cases.

_"The configuration will not live in a separate file for most of cases."_ Where is this coming from? This is a new feature, and we don't have any existing users. The examples in the opentelemetry config are file based. Please only add the bare minimum APIs for now. If there is a legit need in the future, we can always offer more APIs.

opentelemetry-config-stdout/README.md
andborja: It is trying to separate the concerns (metrics, traces, logs) in different registries. Having them in the parent would be a little messy, and you'll still need to add the prefix (metrics, logs, traces) to the function. That prefix is avoided by the metrics_mut() function.

lalitb (commented):
PR has changed quite a bit since my approval, keeping it open till we agree on the design.

utpilla:
@andborja @lalitb and I discussed the design and public API design offline. Here's the approach:

From a user's perspective the workflow would like this:
1. They create a `ConfigurationRegistry`
2. They register configurations by calling `register_metrics_config()` on the `ConfigurationRegistry` instance
3. `register_metrics_config` would register a unique component name and a factory which receives `serde_yaml::Value` and `&mut MeterProviderBuilder`.
4. The factories are offered by the component crates.
5. Once the user has registered their desired factories, they will call `TelemetryProviders::build_from_yaml_str(yaml : &str, configuration_registry)` which would return `TelemetryProviders` (a type that would contain the different signal providers)
6. They could then access the generated `MeterProvider` by calling `TelemetryProviders::meter_provider()` and set it as global `MeterProvvider`.

Public APIs:
- `pub struct ConfigurationRegistry`
- `pub ConfigurationRegistry::register_metrics_config(&mut self, ...)`
- `pub struct TelemetryProviders`
- `pub TelemetryProviders::build_from_yaml_str(yaml: &str)` (static method)
- `TelemetryProviders.meter_provider()` to access the generated `MeterProvider`


From the perspective of components wanting to offer factories for their registration:
- They only need to offer a public method that would accept `&mut MeterProviderBuilder` and `serde_yaml::Value` and make the necessary changes to the `MeterProviderBuilder` and then return the builder.
- The yaml `Value` passed to them would contain information just enough for them to make changes to the builder. For example, a push metric exporter would be provided everything under the `PeriodicReader` section to which they belong.

@andborja Could you please confirm that you agree to this approach? If yes, let's have the PR updated to reflect the abovementioned design.

andborja:
> @andborja @lalitb and I discussed the design and public API design offline. Here's the approach:
> 
> From a user's perspective the workflow would like this:
> 
> 1. They create a `ConfigurationRegistry`
> 2. They register configurations by calling `register_metrics_config()` on the `ConfigurationRegistry` instance
> 3. `register_metrics_config` would register a unique component name and a factory which receives `serde_yaml::Value` and `&mut MeterProviderBuilder`.
> 4. The factories are offered by the component crates.
> 5. Once the user has registered their desired factories, they will call `TelemetryProviders::build_from_yaml_str(yaml : &str, configuration_registry)` which would return `TelemetryProviders` (a type that would contain the different signal providers)
> 6. They could then access the generated `MeterProvider` by calling `TelemetryProviders::meter_provider()` and set it as global `MeterProvvider`.
> 
> Public APIs:
> 
> * `pub struct ConfigurationRegistry`
> * `pub ConfigurationRegistry::register_metrics_config(&mut self, ...)`
> * `pub struct TelemetryProviders`
> * `pub TelemetryProviders::build_from_yaml_str(yaml: &str)` (static method)
> * `TelemetryProviders.meter_provider()` to access the generated `MeterProvider`
> 
> From the perspective of components wanting to offer factories for their registration:
> 
> * They only need to offer a public method that would accept `&mut MeterProviderBuilder` and `serde_yaml::Value` and make the necessary changes to the `MeterProviderBuilder` and then return the builder.
> * The yaml `Value` passed to them would contain information just enough for them to make changes to the builder. For example, a push metric exporter would be provided everything under the `PeriodicReader` section to which they belong.
> 
> @andborja Could you please confirm that you agree to this approach? If yes, let's have the PR updated to reflect the abovementioned design.

Confirmed

opentelemetry-config/examples/application/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+    Ok(())
+}
+
+/// Initializes the telemetry configuration registry with custom providers.
+fn initialize_telemetry_registry() -> ConfigurationProviderRegistry {
+    let mut registry = ConfigurationProviderRegistry::default();
+
+    // Register the custom exporter factory.
+    MockPeriodicExporterProvider::register_into(&mut registry);
```
utpilla: This should be using `register_metric_exporter_factory`.

opentelemetry-config/src/providers.rs
```diff
@@ -0,0 +1,227 @@
...
+            meter_provider_option,
+            logs_provider_option,
+            traces_provider_option,
+        );
+
+        Ok(providers)
+    }
+
+    /// Configures the Telemetry providers from a YAML string
+    pub fn configure_from_yaml(
```
utpilla: Let's add `_str` as a suffix to make the method name more descriptive.

```suggestion
    pub fn configure_from_yaml_str(
```

PR #492, opened 2025-12-16
refactor(actix-web): replace hand-coded constants with semconv metrics

zdannar:
Use standardized metric constants from opentelemetry-semconv crate instead of hand-coded string constants for HTTP server metrics.

Fixes #

## Changes

While using opentelemetry-instrumentation-actix-web crate, I noticed that we are receiving type errors from the otel collector.  After looking at the opentelementry-semantic-versions crate, it appears the standard names are not being used and were coded as constants.

* [X] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/CONTRIBUTING.md) guidelines followed
* [X] Unit tests added/updated (if applicable).  _Tests were run and precommit.sh worked fine._
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes.  _Based on the CONTRIB document, it sounds like the CHANGELOG is generated by the commit message.  I did not change._

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,12 +29,12 @@ jobs:
           - rust: stable
             os: macos-latest
           - rust: stable
-            os: ubuntu-22.04-arm
+            os: ubuntu-24.04-arm
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.rust == 'beta' }}
     steps:
     - name: Harden the runner (Audit all outbound calls)
-      uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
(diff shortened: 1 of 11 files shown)


matteiben-onx:
I believe this change adhere's to stable semantic conventions.  Spec here: https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverresponsebodysize

Starting in semantic conventions v1.22.0, then stabilized in v1.23.0, these were changed:

> BREAKING: Rename http.server.request.size metric to http.server.request.body.size and http.server.response.size metric to http.server.response.body.size

https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md#v1220-2023-10-12
https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md#v1230-2023-11-03

As such, this is technically a breaking change.  However, the project seems to be still pre v1.0.

PR #528, opened 2026-01-22
feat(tower): Add configurable route extraction

### LINKED ISSUE #527: [Tower] Fix span name generation and configuration options to prevent cardinality explosion from `http.route`

jan-xyz:
## Summary

Adds configurable route extraction to the tower instrumentation layer to address cardinality concerns with dynamic path segments.

Closes #527.

The extracted route is used for both:
- Span names: `"{method} {route}"` or just `"{method}"` if no route
- The `http.route` metric attribute

### Route Extractors

| Extractor | Description |
|-----------|-------------|
| `NoRouteExtractor` | No route, method-only span names (e.g., `GET`) - safest for cardinality |
| `PathExtractor` | Uses URL path without query params (e.g., `/users/123`) |
| `AxumMatchedPathExtractor` | Uses Axum's `MatchedPath` for route templates (requires `axum` feature) |
| `FnRouteExtractor` | Custom function-based extraction |

### Default Behavior

- **With `axum` feature**: Uses `AxumMatchedPathExtractor` (low cardinality route templates)
- **Without `axum` feature**: Uses `NoRouteExtractor` (method only)

### Usage

```rust
use opentelemetry_instrumentation_tower::{
    HTTPLayerBuilder,
    PathExtractor,
};

let layer = HTTPLayerBuilder::builder()
    .with_route_extractor(PathExtractor)
    .build()
    .unwrap();
```

For custom normalization (e.g., replacing numeric IDs with placeholders), see the `custom-route-extractor` example.

### DIFF
--- a/opentelemetry-instrumentation-tower/src/lib.rs
+++ b/opentelemetry-instrumentation-tower/src/lib.rs
@@ -263,8 +263,11 @@ pub struct HTTPService<
 
 #[derive(Clone)]
 /// [`Layer`] which applies the OTEL HTTP server metrics and tracing middleware
-pub struct HTTPLayer<RouteExt = DefaultRouteExtractor, ReqExt = NoOpExtractor, ResExt = NoOpExtractor>
-{
+pub struct HTTPLayer<
+    RouteExt = DefaultRouteExtractor,
+    ReqExt = NoOpExtractor,
+    ResExt = NoOpExtractor,
+> {
     state: Arc<HTTPLayerState>,
     route_extractor: RouteExt,
     request_extractor: ReqExt,


opentelemetry-instrumentation-tower/examples/custom-route-extractor/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+    println!("  curl http://localhost:5000/users/alice/profile");
+    println!("  curl http://localhost:5000/users/bob/posts");
+    println!("  curl http://localhost:5000/users/unknown/profile");
+
+    let server = axum::serve(listener, app);
+
+    if let Err(err) = server.await {
+        eprintln!("server error: {err}");
+    }
+}
```
cijothomas: though not the main intent for this example, I suggest to showcase best practices in all examples - specifically, lets add provider shutdowns here.

opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -48,10 +48,181 @@ const NETWORK_PROTOCOL_VERSION_LABEL: &str = semconv::attribute::NETWORK_PROTOCO
...
+/// # Cardinality Considerations
+///
+/// The route should be a low-cardinality value (e.g., a route template like `/users/:id`)
+/// rather than the actual path (e.g., `/users/123`). High-cardinality routes can overwhelm
+/// the OpenTelemetry SDK's built-in cardinality limits and downstream backends.
+pub trait RouteExtractor<B>: Clone + Send + Sync + 'static {
+    /// Extracts the route from the request, if available.
+    ///
+    /// Returns `None` to use method-only span names and skip the `http.route` attribute.
+    fn extract_route(&self, req: &http::Request<B>) -> Option<String>;
```
cijothomas: Do we have to return owned `String` or a ref would do? We seem to be cloning it anyway, so trying to see if an allocation can be avoided.

opentelemetry-instrumentation-tower/CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+  - `NoRouteExtractor` - No route, uses only HTTP method (e.g., `GET`), safest for cardinality
+  - `PathExtractor` - Uses the URL path without query params (e.g., `/users/123`)
+  - `NormalizedPathExtractor` - Normalizes numeric IDs to `{id}` and UUIDs to `{uuid}` (requires `uuid` feature)
+  - `AxumMatchedPathExtractor` - Uses Axum's `MatchedPath` for route templates (requires `axum` feature)
+  - `FnRouteExtractor` - Custom function-based extraction via `with_route_extractor_fn()`
+* New `uuid` feature flag for UUID detection in `NormalizedPathExtractor`
+* Default route extractor depends on features:
+  - With `axum` feature: Uses `AxumMatchedPathExtractor` (route templates, low cardinality)
+  - Without `axum` feature: Uses `NoRouteExtractor` (method only, safest)
+* Route extraction now provides both span names and `http.route` metric attribute from the same source
```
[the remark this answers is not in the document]
cijothomas: That made me realize - should we apply the routing extraction in same way to both span and metric? Metric systems might be more cardinality sensitive than tracing backends.. Something to discuss for future - I think its reasonable to start with same for both traces and metrics.

cijothomas:
@francoposa Could you review please?

opentelemetry-instrumentation-tower/src/lib.rs
francoposa: As far as I can tell after messing with this for a bit:
1. The Tower `Service` trait definition does not allow adding lifetimes to parameters. 
2. To avoid cloning, we would need a lifetime on the `req: &http::Request<B>` parameters in the `Service` implementation.

1 + 2 makes this not possible as far as I can tell so far.

cijothomas (approved):
THanks. 
I left a comment about potential performance improvement. Let's first add some benchmarks and then work on improving perf. Happy to help with both in follow ups.

PR #541, opened 2026-02-09
feat(lambda): read cloud.account.id from symlink in Lambda detector

RaphaelManke:
## Summary

- Reads `cloud.account.id` from the symlink at `/tmp/.otel-aws-account-id` created by the OTel Lambda Extension
- Uses `std::fs::read_link()` to read the symlink target
- Silently skips if the symlink doesn't exist (`if let Ok(...)` guard)

## Depends on

- open-telemetry/opentelemetry-lambda#2127 (extension creates the symlink)

## Changes

- `lambda.rs` — Added readlink logic with `ACCOUNT_ID_SYMLINK_PATH` constant, `detect_with_symlink_path()` for testability
- Tests: happy path and missing symlink

## Test plan

- [x] Unit test: symlink exists → `cloud.account.id` set correctly
- [x] Unit test: missing symlink → attribute absent, no panic


chinmaychahar:
@RaphaelManke could you resolve the merge conflicts on this branch? Thanks!

opentelemetry-aws/src/detector/lambda.rs
```diff
@@ -49,6 +57,15 @@ impl ResourceDetector for LambdaResourceDetector {
             ),
         ];
 
+        if let Ok(account_id) = std::fs::read_link(symlink_path) {
```
chinmaychahar: Can we add validation to the symlink target being read by read_link? If the symlink is corrupted, it could produce garbage data

opentelemetry-aws/src/detector/lambda.rs
```diff
@@ -112,4 +129,83 @@ mod tests {
         let got = detector.detect();
         assert_eq!(Resource::builder_empty().build(), got);
     }
+
+    #[sealed_test]
+    fn test_aws_lambda_detector_with_account_id_symlink() {
+        let symlink_path = std::env::temp_dir().join(".otel-aws-account-id-test");
+        // Clean up any leftover from a previous test run
+        let _ = std::fs::remove_file(&symlink_path);
+        std::os::unix::fs::symlink("123456789012", &symlink_path).unwrap();
```
chinmaychahar, line 157: Since the CI runs on Windows too, can we add tests for both unix and windows?

chinmaychahar (changes requested):

opentelemetry-aws/src/detector/lambda.rs
RaphaelManke: Added validation that checks the symlink target is exactly 12 ASCII decimal digits before accepting it as `cloud.account.id`. This matches the AWS account ID format and silently ignores any corrupted or garbage targets. Added a test to cover this case.

opentelemetry-aws/src/detector/lambda.rs
RaphaelManke: Rather than adding a Windows test variant, I took a step back: since AWS Lambda exclusively runs on Linux, the entire symlink lookup is now gated behind `#[cfg(target_os = "linux")]`. This means the feature simply does not compile on Windows or macOS — there is nothing to test there. All symlink-related tests are also `#[cfg(target_os = "linux")]` and will run in the Linux CI job where they are meaningful.

chinmaychahar: sg

chinmaychahar:
LGTM! @RaphaelManke

chinmaychahar (approved):

chinmaychahar:
@RaphaelManke can you fix linter failures?

chinmaychahar (approved):

cijothomas (approved):

PR #548, opened 2026-02-13
perf: Add benchmarks for tower instrumentation

cijothomas:


opentelemetry-instrumentation-tower/benches/middleware.rs
```diff
@@ -0,0 +1,216 @@
...
+        });
+    });
+
+    // Scenario 2: Tracing only (no-op meter via default global)
+    group.bench_function(BenchmarkId::new("request", "tracing"), |b| {
+        b.to_async(&rt).iter_custom(|iters| async move {
+            let _provider = setup_tracer();
+            // Global meter is not set, so HTTPLayer's meter instruments are no-op
+
+            let layer = HTTPLayerBuilder::builder().build().unwrap();
```
utpilla: I think we could simplify things and reduce some more noise by moving the provider + HTTP layer setup from `iter_custom` to `bench_function`'s closure:

```rust
b.to_async(&rt).iter_custom(|iters| {
        let layer = layer.clone(); // cheap Arc clone
        async move {
            let _service = ServiceBuilder::new()
                .layer(layer)
                .service(tower::service_fn(handler));

            let start = std::time::Instant::now();
            for _ in 0..iters {
                // ... same as before
            }
            start.elapsed()
        }
    });
});
```

opentelemetry-instrumentation-tower/benches/middleware.rs
```diff
@@ -0,0 +1,282 @@
...
+        b.to_async(&rt).iter_custom(|iters| async move {
+            let mut service = tower::service_fn(handler);
+
+            let start = std::time::Instant::now();
+            for _ in 0..iters {
+                let req = http::Request::builder()
+                    .method("GET")
+                    .uri("http://example.com/users/123")
+                    .body(String::new())
+                    .unwrap();
```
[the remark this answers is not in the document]
jan-xyz: I did a quick trial locally :

| Scenario                    | Docs (M4 Pro) | Before (M4 Max) | After (M4 Max) | Δ     | Overhead vs Baseline |
|-----------------------------|-------------------|-----------------|----------------|-------|----------------------|
| Baseline                    | ~131 ns           | ~99 ns          | ~41 ns         | −59%  | -                    |
| No-op                       | ~936 ns           | ~740 ns         | ~692 ns        | −7%   | +~651 ns             |
| Tracing                     | ~1100 ns          | ~885 ns         | ~840 ns        | −5%   | +~799 ns             |
| Tracing (sampled-out)       | ~999 ns           | ~776 ns         | ~734 ns        | −5%   | +~693 ns             |
| Metrics                     | ~1295 ns          | ~1102 ns        | ~1036 ns       | −6%   | +~995 ns             |
| Tracing + Metrics           | ~1436 ns          | ~1278 ns        | ~1192 ns       | −7%   | +~1151 ns            |

francoposa (commented):
I am finding the notes are making it unclear to me what we are trying to measure and improve.

For instance:
> // Scenario 2: Middleware present, but both tracer and meter are no-ops.
> // Measures the pure overhead of the HTTPLayer machinery (attribute extraction,
> // context propagation hooks, etc.) when no real telemetry is produced.
> // **Ideally this should be very close to the baseline.**

and
```rust
//! - **Unconditional attribute extraction**: all `KeyValue` attributes (span attrs,
//!   metric label superset, custom extractors) are built on every request without
//!   first checking whether the tracer or meter are no-ops and would discard them
//!   immediately anyway.
```

If the user is going to not produce telemetry, they can use an if-statement in main to not enable the middleware at all. That's not a case we care to measure.
If _either_ metrics or tracing is going to get used, then all the `KeyValue`s in `label_superset` will be used and must be determined from the request.

Then we have :
```rust
**`Box::pin`** wrapping the response future — heap-allocates on every request.
```
AFAICT this is a requirement that is derived from the Tower Service trait design with [currently no good way around it](https://users.rust-lang.org/t/is-it-possible-to-not-use-box-in-this-code/133876/4), just like the owned Strings we discussed on the previous PR.

For:
```rust
//!   Caching would require a new `with_propagator` builder method since there is no
//!   public API to obtain an owned/`Arc`'d propagator from the global.
```

Is this a limitation of the middleware as currently written or the underlying OTel libs? Is there any recommended path available to improve or work around this?

---
I am not opposed to merging what we have for the sake of keeping this PR simple (after taking the request builder out of the hot loop), but I think we want to follow up to re-arrange the code and the notes better so we can differentiate between three core aspects:
1. What is under the control of the user w.r.t configuration of this middleware only (not the OTel tracers/samplers/exporters).
2. What is under the control of the middleware's performance optimizations
3. What is attributable to OTel instrumentation out of the control of this middleware.

jan-xyz (approved):

cijothomas:
@francoposa addressed the request-builder-in-loop point (now built outside the timed loop, see commit c777b62b) and rebased onto current main (OTel 0.32) with refreshed reference numbers (commit 7844e252). The 3-way reorg of perf notes (user / middleware / OTel) and the `label_superset` / `with_propagator` clarifications are still pending — happy to address as a follow-up so we can land this baseline first. PTAL when you get a chance.

francoposa (approved):
LGTM, let's merge with this so @jan-xyz can rebase the PR to bring the ResponseFuture approach back and drop that allocation.

I also have a branch sitting around one to turn protocol and version to `&str` so we can drop those allocs as well.

opentelemetry-instrumentation-tower/Cargo.toml
lalitb, line 13: This should include both   "src" and "benches", otherwise the published package can end up with a `[[bench]]` target in `Cargo.toml` that point to a benchmark file that was excluded from the package.

lalitb (approved):

PR #549, opened 2026-02-16
feat(aws): Add AWS X-Ray span exporter

JeremieRodon:
## Changes

Adds a new `xray_exporter` module to the `opentelemetry-aws` crate that translates OpenTelemetry `SpanData` into [AWS X-Ray segment documents](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html) and exports them.

Includes:
- `XrayExporter` type implementing the `SpanExporter` trait
- `SegmentTranslator` type for translating OTel Spans into X-Ray Segments/Subsegments
- `SegmentDocumentExporter` trait allowing for custom export logic
- `XrayDaemonClient` and `StdoutClient` types, implementing `SegmentDocumentExporter` (feature-gated)
- Integration tests, benchmarks, and documentation

This PR is huge but only ***adds*** code, with no modifications to existing code files.

## Motivation

I need to export OpenTelemetry data from AWS Lambda functions written in Rust to AWS X-Ray via the locally available X-Ray daemon. The daemon remains the most reliable transport mechanism — AWS's own "Distro for OpenTelemetry" Python implementation recently reverted to using it after performance issues with the Go-based exporter. Rather than keeping this translation logic in my own crate, I'm contributing it here so the broader Rust OpenTelemetry community can benefit.

### Core components
- **`SegmentDocumentExporter`** — trait for pluggable exporter backends
- **`XrayExporter<Client>`** — implements `SpanExporter`, generic over any backend client that implements `SegmentDocumentExporter`
- **`SegmentTranslator`** — configurable translator converting `&[SpanData]` to X-Ray segment documents using an attribute processing pipeline using:
  - Immediate dispatch of 42 semantic convention attributes, when 1 to 1 mapping is possible between Span and Segment
  - 11 deferred `ValueBuilder` implementations that accumulate multiple attribute values and resolve to the correct X-Ray segment value once every span attributes has been seen (e.g., assembling `http.request.url` from its component parts, or choosing the `name` of the segment)
- **`XrayDaemonClient`** — UDP client for the X-Ray daemon, implements `SegmentDocumentExporter` (feature-gated)
- **`StdoutClient`** — debug/development output client, implements `SegmentDocumentExporter` (feature-gated)

### Design highlights
- **Zero-copy where possible**: `SegmentDocument<'a>` borrows from input `SpanData` via `Cow<'a, str>`, minimizing allocations in the hot path
- **Phantom-typed builders**: `DocumentBuilder<'a, DBT>` enforces segment vs. subsegment constraints at the type level (e.g., segments require `trace_id`, subsegments require `parent_id`)
- **Both current and deprecated semantic conventions** are supported for backward compatibility, though that may be far from complete coverage
- **Subsegment nesting** (feature-gated behind `subsegment-nesting`): builds a tree of parent-child relationships with precursor ID computation, for cases where callers want to nest subsegments within their parent segments in a single batch rather than send all the Segment/Subsegment documents in a flat array (which is totally fine for the X-Ray service, but initially I thought it *cool* to implement that and now I don't want to throw the code...)

### Feature flags

| Feature | Default | Description |
|---|---|---|
| `xray-exporter` | Yes | Core exporter module |
| `xray-daemon-client` | No | UDP client for X-Ray daemon |
| `xray-stdout-client` | No | Debug stdout client |
| `subsegment-nesting` | No | Nested subsegments with precursor ID computation |

### X-Ray concept mapping

Segments are produced from `Server` spans; all other span kinds become independent subsegments. Annotations (searchable, capped at 50) are populated from explicitly indexed attributes (I support an `aws.xray.annotations` attribute that allows to specify attribute to index on the span itself); remaining unprocessed attributes go to metadata. 

I mainly tried to follow:
- X-Ray documentation: [https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html)
- ADOT documentation on mapping attributes: [https://aws-otel.github.io/docs/getting-started/x-ray](https://aws-otel.github.io/docs/getting-started/x-ray)
- Go reference implementation: [https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/awsxrayexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/awsxrayexporter)
 
## Testing

- **Unit tests** covering builder validation, ID parsing/formatting, dispatch table routing, value extraction, and `MaybeSkip` serialization behavior
- **Integration tests** covering: end-to-end translation pipeline, semantic convention attribute mapping, constraint validation, error handling, JSON serialization format, parent-child relationships, and resource translation (EC2, ECS, EKS, Lambda, Beanstalk)
- **Benchmarks** measuring translation and UDP export throughput at batch sizes of 10, 100, and 512 spans, with `pprof` flamegraph support

> **Note on AI-assisted content:** I wrote all the code myself, but used AI extensively for documentation, test generation, and benchmarks. Even if I read and corrected all of them, there may be some redundant or imperfect tests in the mix, but I'm confident the overall coverage is *ok*.

## Benchmarks

Performance measurements on the full translation + UDP export pipeline:

### Without Subsegment Nesting (default)

| Batch Size | Translation Only | Export Only | Translation + Export | Per-Span |
|------------|------------------|-------------|----------------------|----------|
| 10 spans   | **18.8 µs**      | **64.0 µs** | **107.0 µs**         | 10.7 µs  |
| 100 spans  | **196.4 µs**     | **662.1 µs**| **1.10 ms**          | 11.0 µs  |
| 512 spans  | **1.05 ms**      | **3.43 ms** | **5.85 ms**          | 11.4 µs  |

**Key insights:**
- Translation overhead: **~2.0 µs per span** without nesting, **~2.3 µs per span** with nesting
- Subsegment nesting adds ~18% overhead to translation (tree building + precursor ID computation)
- Network export dominates total latency (specifically serialization, per previous tests that are no longer there)

My system (not ideal for benchmark I'll admit...):
OS: NixOS 26.05.20260123.c5296fd (Yarara) x86_64 
Kernel: 6.12.67
CPU: 11th Gen Intel i7-1185G7 (8) @ 4.800GHz 
Memory: 32GB

<details>
<summary>Expand for more benchmark breakdown</summary>

### With Subsegment Nesting (feature-gated)

| Batch Size | Translation Only |
|------------|------------------|
| 10 spans   | **22.2 µs**      |
| 100 spans  | **232.6 µs**     |
| 512 spans  | **1.27 ms**      |

All measurements represent mean values with 95% confidence intervals, in nanoseconds:

**Translation + Export (without nesting):**
- 10 spans: 106,996 ns ± 387 ns
- 100 spans: 1,100,215 ns ± 5,187 ns
- 512 spans: 5,845,351 ns ± 15,297 ns

**Translation Only (without nesting):**
- 10 spans: 18,822 ns ± 42 ns
- 100 spans: 196,392 ns ± 163 ns
- 512 spans: 1,049,404 ns ± 1,859 ns

**Translation Only (with nesting):**
- 10 spans: 22,214 ns ± 47 ns
- 100 spans: 232,647 ns ± 364 ns
- 512 spans: 1,267,295 ns ± 2,639 ns

**Export Only:**
- 10 spans: 63,961 ns ± 378 ns
- 100 spans: 662,097 ns ± 3,058 ns
- 512 spans: 3,434,034 ns ± 23,957 ns

Run `cargo bench --package opentelemetry-aws --features xray-daemon-client` to reproduce.
</details>

## Context and caveats

This is my first contribution of this magnitude to an open-source project — I'm happy to take feedback and iterate.

I don't consider this work definitive. The X-Ray translation space is surprisingly inconsistent: the ADOT documentation, the Go "reference implementation", and the current Python implementation by AWS all describe *not-so-slightly* different behaviors. I made pragmatic choices where these sources diverged, guided by my own needs, and some may not suit every use case.

I've tried to lay solid foundations that can be extended, but contributions and suggestions are very welcome. In particular, additional semantic convention mappings and edge cases would benefit from community input.

## Related work

I'm also developing a separate crate to streamline the `tracing → OpenTelemetry → X-Ray` integration specifically for AWS Lambda functions, leveraging the AWS SDK for Rust's existing `tracing` instrumentation to provide X-Ray capabilities out of the box. This PR provides the shared translation layer that crate will depend on.


chinmaychahar:
Thanks for the PR! @JeremieRodon , will review it shortly

chinmaychahar:
Thanks for this well-structured and thorough implementation, @JeremieRodon! 

I'm still going through the full review, but wanted to share an initial observation. The xray-exporter feature is default currently and I see the existing pattern in Cargo.toml was default = `["trace", "internal-logs"]` only. Could we keep xray-exporter as opt-in? Want to understand your design choices.

I'll follow up with more detailed inline comments as I work through the rest. Great work on the documentation and test coverage!

opentelemetry-aws/Cargo.toml
```diff
@@ -31,15 +46,29 @@ opentelemetry-semantic-conventions = { workspace = true, optional = true, featur
     "semconv_experimental",
 ] }
 tracing = {version = "0.1", optional = true}
+serde = { version = "1.0", features = ["derive"] }
```
chinmaychahar: could we mark this as optional = true as well? it's not a lightweight dependency

chinmaychahar:
Could you also resolve the merge conflicts? Thanks!

JeremieRodon:
Thanks @chinmaychahar !

> The xray-exporter feature is default currently and I see the existing pattern in Cargo.toml was default = ["trace", "internal-logs"] only. Could we keep xray-exporter as opt-in? Want to understand your design choices.

I don't remember but I suppose it is a development artifact I forgot to revert before submitting. Of course it should be opt-in.

I'll resolve the merge conflict. Thank you for your time, appreciate it.

opentelemetry-aws/src/xray_exporter/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+//! | `xray-daemon-client` | Enables the [`daemon_client`] sub-module with [`XrayDaemonClient`] for sending segments to the X-Ray daemon over UDP |
+//! | `xray-stdout-client` | Enables the [`stdout_client`] sub-module with [`StdoutClient`] for writing segments to stdout (useful for debugging) |
+//! | `subsegment-nesting` | Enables the use of subsegment nesting via [`SegmentTranslator::always_nest_subsegments`] |
+//! | `internal-logs` | Enables internal `tracing` instrumentation throughout the module |
+//!
+//! [`XrayDaemonClient`]: daemon_client::XrayDaemonClient
+//! [`StdoutClient`]: stdout_client::StdoutClient
+//! [`SpanExporter`]: opentelemetry_sdk::trace::SpanExporter
+
+use core::{error::Error, fmt, future::Future};
```
chinmaychahar: I noticed `core::error::Error` was stabilized in [Rust 1.81](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0/) and this crate declares rust-version as 1.75.0 

The rest of this workspace uses std::error::Error, could you switch to that? Same for the doc example at line 163.

opentelemetry-aws/src/xray_exporter/daemon_client/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+                .unwrap_or_else(|| {
+                    #[cfg(feature = "internal-logs")]
+                    tracing::warn!(
+                        "No valid {} env variable detected, falling back on default",
+                        Self::AWS_XRAY_DAEMON_ADDRESS_ENV_VAR
+                    );
+                    SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), Self::DEFAULT_DAEMON_PORT)
+                }),
+        )
+        .expect("could not bind daemon")
```
chinmaychahar, line 219: Suggestion: The quick-start example uses default(), which panics if the socket can't bind. Maybe show new() in the example instead? It does the same thing but lets the user handle the error instead of crashing

opentelemetry-aws/src/xray_exporter/translator/attribute_processing/value_builder/http_request_url_builder.rs
```diff
@@ -0,0 +1,847 @@
...
+
+impl<'value> ValueBuilder<'value> for HttpRequestUrlBuilder<'value> {
+    fn resolve(self, segment_builder: &mut AnyDocumentBuilder<'value>) -> Result<()> {
+        let url = if let Some(url_full) = self.url_full {
+            // Full URL available so no need to assemble
+            Some(Cow::Borrowed(url_full))
+        } else if let Some(domain) = self.url_domain.or(if self.span_kind_is_server {
+            self.server_address
+        } else {
+            self.client_address
```
chinmaychahar, line 92: should it be just server.address as fallback?  client.address is the caller's IP, not the destination so the target URL should use server.address. Lemme know if my understanding is correct and this sounds reasonable!

opentelemetry-aws/src/xray_exporter/translator/mod.rs
```diff
@@ -0,0 +1,1256 @@
...
+                // - Update the last_seens_endtime
+                // - Add the id of the present segment to the precursor list
+                if let Some(end_time) = end_time {
+                    last_seens_endtime = Some(end_time);
+                    // MARKER ALLOC
+                    precursors.push(id);
+                }
+            }
+
+            let document_builders_parent_index = &(trace_id, parent_id);
```
chinmaychahar, line 691: we can add a check here, something like if `document_builders_index != document_builders_parent_index`

opentelemetry-aws/src/xray_exporter/translator/mod.rs
```diff
@@ -0,0 +1,1256 @@
...
+                .get_mut(document_builders_index)
+                .expect("builders in subsegments are always also in document_builders")
+            else {
+                continue;
+            };
+
+            if last_seen_parent_id != Some(parent_id) {
+                last_seen_parent_id = Some(parent_id);
+                // We changed parent, clear precursors
+                precursors.clear();
```
chinmaychahar, line 664: Should add last_seens_endtime = None here

opentelemetry-aws/src/xray_exporter/translator/attribute_processing/value_builder/http_response_content_length_builder.rs
chinmaychahar, line 1: NIT: the filename has a small typo 😄

chinmaychahar (changes requested):

chinmaychahar:
The workspace has an examples/ directory. A runnable example would help users for a feature this size, what do you think?

chinmaychahar:
Could you also fix the clippy warning? Also, the full CI suite hasn't run yet, @cijothomas can you help us trigger it once the fixes are in?

chinmaychahar:
Also, two doc updates needed since xray-exporter is now opt-in: 1. CHANGELOG.md still says enabled by default and 2. README.md feature table shows xray-exporter default = Yes

opentelemetry-aws/src/xray_exporter/translator/attribute_processing/value_builder/http_request_url_builder.rs
JeremieRodon: Based on [ADOT OTel -> X-Ray doc](https://aws-otel.github.io/docs/getting-started/x-ray#otel-span-http-attributes-translation) for the `http.request.url` X-Ray field, they seem to use either one depending on the span `Kind`.
Furthermore, the [Go reference implementation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/awsxrayexporter/internal/translator/http.go#L139) is indeed doing the same kind of branching as the code you highlighted.
I confess I mostly followed their choices without thinking too much about whether or not they make sense... As I said in my initial message, it is hard because the doc and the various AWS implementations often disagree. But on that particular topic they seem aligned so I didn't really looked past that.

I'm not sure *why* this is like that. Could we follow the ref impl for now (assuming I did not messed up while trying to replicate the logic), and in the future if real use-cases seem to suffer, we can always push a bug fix?

opentelemetry-aws/src/xray_exporter/translator/mod.rs
JeremieRodon: Is it really possible that an OTel span references itself as a parent?
I will suppose it is seeing you deem the check necessary...

JeremieRodon:
> The workspace has an examples/ directory. A runnable example would help users for a feature this size, what do you think?

I will try to think of something

opentelemetry-aws/src/xray_exporter/translator/attribute_processing/value_builder/http_request_url_builder.rs
chinmaychahar: Fair, if the ADOT docs and Go reference impl both do this, let's keep it consistent with them. Thanks for checking!

opentelemetry-aws/src/xray_exporter/translator/mod.rs
chinmaychahar: probably not in a normal operation but a defensive check will avoid an unexpected input. What do you suggest?

chinmaychahar:
Thanks for the fixing serde and the error core import! A few doc comments in mod.rs still mention "enabled by default", nothing else looks majorly blocking to me

cijothomas:
@JeremieRodon can you rebase from main and fix changelog conflicts?

JeremieRodon:
> @JeremieRodon can you rebase from main and fix changelog conflicts?

Done.


> The workspace has an examples/ directory. A runnable example would help users for a feature this size, what do you think?

I did not have time to work on that yet.

I *do* have examples of usage of the `awssdk-instrumentation` crate that I built on top of this feature, but as it leans heavily toward the AWS Lambda use-case, it is in the form of a demo repository instead of Rust examples (see [yak-mania](https://github.com/RustyServerless/yak-mania) and [awssdk-instrumentation](https://github.com/RustyServerless/awssdk-instrumentation) if curious).

When we consider *only* the `xray-exporter` feature added to this crate, there is really no more to show than what I included in the module documentation. Should I extract the doc example usage into the examples repository with some Yak shaving shenanigans? It is low value but not very time-consuming so if people tend to lookup the `examples/` directory first it may be worth it.

chinmaychahar:
Thanks for the rebase and fixes! I will let you take the call on the examples. If you think that the usage examples in the module docs cover it well, sounds good.

NIT: CHANGELOG.md still says "enabled by default" for the xray-exporter feature. Once that's updated, the PR is good to go for me

JeremieRodon:
> Thanks for the rebase and fixes! I will let you take the call on the examples. If you think that the usage examples in the module docs cover it well, sounds good.

I think it is OK as it is. By itself it is "just" an new OpenTelemetry Exporter. 

I doubt anyone will want to use it stand-alone because it is *a priori* tied to AWS use-cases, which will almost certainly depend on the *AWS SDK for Rust*, which in turn will need to be instrumented to automatically extract useful info such as DynamoDB table names, S3 buckets, etc... This is outside of the scope of the `opentelemetry-aws` crate (I think?) which is why I made the `awssdk-instrumentation` crate based on the `xray-exporter`. I decided to contribute the exporter here because it felt like the correct place for it to be.

chinmaychahar:
Thanks! LGTM 🚀

chinmaychahar (approved):

chinmaychahar:
@cijothomas could we trigger the CI checks for this PR?

JeremieRodon:
Just fixed 2 CI errors, one in `doc` and a `lint` in the xray_exporter benchmark. For reasons that escapes me, I did not succeeded at running the `script/lint.sh`, or rather it does not seem to care about the benches folder on my computer...

I did not know what to do with the other CI errors, please advise.
CI run link: [https://github.com/open-telemetry/opentelemetry-rust-contrib/actions/runs/27258143706](https://github.com/open-telemetry/opentelemetry-rust-contrib/actions/runs/27258143706)

chinmaychahar:
I'll take a look

chinmaychahar:
@JeremieRodon Two things to fix from CI:

1. cargo-machete reports url as unused so we can add url to the ignored list in [package.metadata.cargo-machete]. I see tracing is already there.
2. pprof doesn't compile on Windows error - can you move it to a unix-only dev-dependency section so it's skipped on Windows CI?

The other failures look like timeouts, not your code. A re-run should fix those.

JeremieRodon:
@chinmaychahar thanks for the explanations. I did what I though was right, hopefully we are _finally_ set!

chinmaychahar:
Perfect, thanks!

chinmaychahar (approved):

cijothomas (approved):
LGTM.

PR #551, opened 2026-02-26
feat(geneva-uploader): Make rustls-tls the default TLS backend, gate cert auth behind feature flag

lalitb:
Description:

   The `geneva-uploader` crate previously hardcoded `native-tls` (OpenSSL) for all
   TLS operations. This breaks builds targeting musl/Alpine (e.g., the otel-arrow
   Rust collector Docker image) where `openssl-sys` fails to cross-compile.

   Since the primary auth methods going forward are MSI and Workload Identity (not
   certificate-based mTLS), this PR switches the default TLS backend to `rustls`
   and makes certificate auth opt-in.

   ### Feature flags

   | Feature | Description | Default |
   |---------|-------------|---------|
   | `rustls-tls` | Uses rustls for TLS. No system OpenSSL needed. | ✅ |
   | `cert-auth` | Certificate-based mTLS auth (PKCS#12). Implies `native-tls`. | ❌ |
   | `native-tls` | Use native-tls/OpenSSL directly. | ❌ |
   | `self_signed_certs` | Accept self-signed certs (dev only). | ❌ |

   ### Migration

   Most consumers (MSI/Workload Identity auth) need no changes — the default
   feature set just works without OpenSSL.

   For certificate auth, explicitly opt in:

   ```toml
   geneva-uploader = { version = "...", features = ["cert-auth", "self_signed_certs"] }
```

### Changes

   - `Cargo.toml`: Disabled default features on `azure_core` and `azure_identity` to
     prevent transitive `native-tls` via `typespec_client_core` → `reqwest`
   - `geneva-uploader-ffi/Cargo.toml`: Added `features = ["cert-auth"]` since FFI exposes certificate auth
   - `opentelemetry-exporter-geneva/Cargo.toml`: Added `cert-auth` in dev-dependencies for examples

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
@@ -4,19 +4,15 @@ pub(crate) mod client;
 mod tests {
     use crate::config_service::client::{AuthMethod, GenevaConfigClient, GenevaConfigClientConfig};
     #[cfg(feature = "cert-auth")]
-    use openssl::{pkcs12::Pkcs12, pkey::PKey, x509::X509};
-    #[cfg(feature = "cert-auth")]
-    use rcgen::generate_simple_self_signed;
-    #[cfg(feature = "cert-auth")]
-    use std::io::Write;
-    #[cfg(feature = "cert-auth")]
-    use std::path::PathBuf;
-    #[cfg(feature = "cert-auth")]
-    use tempfile::NamedTempFile;
-    #[cfg(feature = "cert-auth")]
-    use wiremock::matchers::{method, path};
-    #[cfg(feature = "cert-auth")]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs [hunk 1]
```diff
@@ -265,12 +270,14 @@ impl GenevaConfigClient {
         let agent_identity = "GenevaUploader";
         let agent_version = "0.1";
 
+        #[allow(unused_mut)]
```
utpilla: Let's not do a blanket suppression of the lint here. This is more explanatory as to when the `client_builder` is actually mutated after creation.

```rust
#[cfg(feature = "cert-auth")]
let mut client_builder = Client::builder()...;
#[cfg(not(feature = "cert-auth"))]
let client_builder = Client::builder()...;
```

PR #561, opened 2026-03-10
refactor(tower): replace Box::pin with concrete ResponseFuture 

jan-xyz:
Eliminates heap allocation per request by introducing a concrete ResponseFuture type that embeds the inner service future directly. The OTel context is attached manually during poll, keeping span propagation correct while also relaxing the S::Future bound from Send + 'static to just Send.

Fixes https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/548#issuecomment-4000873915

## Changes

Replace the box implementation which causes heap allocations with a response future. This has a small performance benefit:

| Scenario | Box::pin (old) | pin-project (new) | Change |
|---|---|---|---|
| **Baseline** | 102.00 ns | 101.24 ns | noise |
| **No-op** | 750.27 ns | 753.00 ns | noise (+0.7%) |
| **Tracing** | 930.06 ns | 897.82 ns | **-3.5%** |
| **Tracing (sampled-out)** | 804.22 ns | 779.73 ns | **-2.5%** |
| **Metrics** | 1115.8 ns | 1072.0 ns | **-3.5%** |
| **Tracing + Metrics** | 1246.4 ns | 1238.7 ns | **-2.8%** |

(I used the benchmark from [this PR](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/548))


francoposa (approved):
LGTM, I think @cijothomas would have to be the one to double-check the usage of the context guard although it is covered by tests

jan-xyz:
Hey, not sure if this fell through the cracks or you're just swamped with other work. I just thought I ask anyway to see if there is something I can do to move this forward?

cijothomas (approved):

PR #562, opened 2026-03-13
feat(etw-traces): add opentelemetry-etw-traces crate

jevidako:
Fixes #

## Changes
Adds a new OpenTelemetry SpanProcessor that exports trace spans to ETW (Event Tracing for Windows) using TraceLogging Dynamic, encoded in Microsoft Common Schema v4.0 format (Part A / Part B / Part C).

   - Processor implementing SpanProcessor for direct ETW export
   - Builder pattern with provider name validation (cross-compatible and ETW-only modes)
   - Configurable event name (defaults to "Span")
   - Optional resource attribute promotion to Part C via with_resource_attributes()
   - Span attributes exported as individually typed ETW fields
   - Span events and links serialized as JSON


jevidako:
> @jevidako - Thanks for the PR. This is still the draft, so wondering if this is ready for review ?

Hi @lalitb, actually it should be ready for review. thanks!

opentelemetry-etw-traces/src/exporter/options.rs
```diff
@@ -0,0 +1,59 @@
+use std::borrow::Cow;
+
+/// Configuration options for the ETW trace exporter.
+#[derive(Debug)]
+pub(crate) struct Options {
+    /// Name of the ETW provider to register.
+    provider_name: String,
+
+    /// ETW event name for span events.
```
cijothomas: Don't need to allow customization. Spans always take Span table - this is good enough.

opentelemetry-etw-traces/src/exporter/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+            resource: Default::default(),
+            resource_attribute_keys,
+            options,
+        }
+    }
+
+    /// Exports a single span to ETW.
+    ///
+    /// Event write order follows Common Schema: reset → __csver__ → PartA → PartC → PartB → write
+    pub(crate) fn export_span_data(&self, span_data: &SpanData) {
```
cijothomas, line 95: we should be checking if there is a listener before serializing and writing.

opentelemetry-etw-traces/src/exporter/options.rs
jevidako: I see, but for the solutions that already have partial implementation and will integrate this once it’s ready, having the option to plug into the current system design would be quite useful. For instance, we have two projects where we plan to adopt this crate once it is merged, and both already use predefined (project name–based) deployed providers. If you feel that customization should not be allowed by default, perhaps this could be introduced as an optional feature at least?

opentelemetry-etw-traces/src/exporter/common.rs
```diff
@@ -0,0 +1,338 @@
...
+/// internal heap allocations.
+pub(crate) fn system_time_to_str(time: &SystemTime) -> String {
+    let (buf, len) = format_timestamp(time);
+    let s = std::str::from_utf8(&buf[..len]).expect("timestamp buffer contains only ASCII");
+    s.to_owned()
+}
+
+/// Writes timestamp into a stack `[u8; 32]` buffer.
+/// Returns the buffer and the number of bytes written.
+/// Maximum length of a formatted timestamp: "YYYY-MM-DD HH:MM:SS.NNNNNNNNN" = 29 chars.
```
urosjovanovic: Note that this format is not per RFC3339 (e.g. 2026-05-15T11:23:45.123456789Z) and may differ compared to how other timestamps (e.g. span start/end) show up.

opentelemetry-etw-traces/src/exporter/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+/// Holds a TLD provider and options. Uses thread-local `EventBuilder`
+/// for zero-allocation reuse across export calls.
+pub(crate) struct ETWExporter {
+    provider: Pin<Arc<tld::Provider>>,
+    resource: Resource,
+    options: Options,
+    resource_attribute_keys: Vec<Cow<'static, str>>,
+}
+
+impl ETWExporter {
```
urosjovanovic, line 63: Do we need to be more explicit here - e.g. EtwSpanExporter? Not sure what the convention is for other types (logs, metrics).

opentelemetry-etw-traces/src/exporter/common.rs
```diff
@@ -0,0 +1,338 @@
...
+}
+
+fn array_to_json(arr: &opentelemetry::Array) -> String {
+    serde_json::to_string(&array_to_value(arr)).unwrap_or_default()
+}
+
+/// Serializes span links to a JSON string (array of {toTraceId, toSpanId}).
+pub(crate) fn links_to_json(links: &[opentelemetry::trace::Link]) -> String {
+    if links.is_empty() {
+        return "[]".to_string();
```
urosjovanovic: This will likely be the default case, can we avoid instantiating a new string every time?

opentelemetry-etw-traces/src/exporter/mod.rs
jevidako: Done :)

opentelemetry-etw-traces/src/exporter/mod.rs
jevidako: In opentelemetry-etw-logs, it’s only named EtwExporter, so I kept the same naming for consistency.

opentelemetry-etw-traces/README.md
```diff
@@ -0,0 +1,90 @@
+# OpenTelemetry Trace Exporter for ETW
+
+![OpenTelemetry — An observability framework for cloud-native software.][splash]
+
+[splash]: https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/main/assets/logo-text.png
+
+| Status        |           |
+| ------------- |-----------|
+| Stability     | beta      |
+| Owners        |           |
```
cijothomas: @jevidako Can you list yourselves as owner?

opentelemetry-etw-traces/src/lib.rs
```diff
@@ -0,0 +1,55 @@
+//! # OpenTelemetry ETW Traces Exporter
```
cijothomas, line 1: https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/586/changes - we recently added integration test for ETW exporter. Could you copy the same for this one too, so we can have integration test running in CI. No need to be very detailed - a simple test is good enough to begin with. We can expand them later in future PRs.

opentelemetry-etw-traces/src/exporter/common.rs
```diff
@@ -0,0 +1,341 @@
...
+use std::borrow::Cow;
+use std::time::SystemTime;
+use tracelogging_dynamic as tld;
+
+use super::otel_id_ext::{SpanIdExt, TraceIdExt};
+
+/// Converts an OpenTelemetry `SpanKind` to a u8 value matching the OTel spec.
+pub(crate) fn span_kind_to_u8(kind: &SpanKind) -> u8 {
+    match kind {
+        SpanKind::Internal => 1,
```
cijothomas: this mapping is incorrect (CommonSchema uses a slightly diff values than protobuf.)
Check the user-events trace exporter to get the correct values.

opentelemetry-etw-traces/src/exporter/part_b.rs
```diff
@@ -0,0 +1,98 @@
...
+        tld::OutType::Default,
+        field_tag,
+    );
+    event.add_filetime(
+        "startTime",
+        tld::win_filetime_from_systemtime!(span_data.start_time),
+        tld::OutType::Default,
+        field_tag,
+    );
+    event.add_str8(
```
cijothomas: parentId is optional in CommonSchema and only needed for non-root spans. Let's do this conditionally. (like user-events-trace exporter already does IIRC)

opentelemetry-etw-traces/src/exporter/part_b.rs
```diff
@@ -0,0 +1,98 @@
...
+    );
+    event.add_u8(
+        "kind",
+        common::span_kind_to_u8(&span_data.span_kind),
+        tld::OutType::Default,
+        field_tag,
+    );
+    event.add_filetime(
+        "startTime",
+        tld::win_filetime_from_systemtime!(span_data.start_time),
```
cijothomas: this is not the right format.... check the etw-logs for the example to copy from.

opentelemetry-etw-traces/src/exporter/part_c.rs
```diff
@@ -0,0 +1,74 @@
...
+/// added as typed fields.
+pub(crate) fn populate_part_c(
+    event: &mut tld::EventBuilder,
+    span_data: &SpanData,
+    resource: &super::Resource,
+    field_tag: u32,
+) {
+    let resource_attr_count = resource.attributes_from_resource.len();
+    let span_attr_count = span_data.attributes.len();
+    let additional_span_data = 1u8; // 'events'
```
cijothomas: events should be ignored. There is no Common Schema mapping for SpanEvents, and is not supported in any etw/user-events exporter.

opentelemetry-etw-traces/src/exporter/part_b.rs
```diff
@@ -0,0 +1,98 @@
...
+/// PartB {
+///     _typeName: "Span"
+///     name: str8
+///     kind: u8
+///     startTime: filetime
+///     parentId: str8
+///     [links: str8 (JSON)]    // only if present
+///     [statusMessage: str8]   // only if status has description
+///     success: bool32
+/// }
```
cijothomas, line 25: https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-user-events-trace/src/trace/exporter.rs#L24

a set of well known fields from Common Schema PartB should be handled, else they get dropped as PartC which is incorrect.

opentelemetry-etw-traces/src/exporter/mod.rs
cijothomas: this is fine. These are internal anyway.

cijothomas (changes requested):
Thanks for contributing! I have review and left comments. 
Most of them pointing to gaps about CommonSchema mapping - user-events-trace exporter can be used as a good example to follow. (Or you can use the Common Schema doc inside Microsoft too).

opentelemetry-etw-traces/README.md
cijothomas: @urosjovanovic too, if okay!

cijothomas:
Reviewing this PR made be realize we have some gaps in user-events-trace too! Opened a PR to fix them - https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/632 we'd definitely want them to be taken care in this PR too.

opentelemetry-etw-traces/src/exporter/common.rs
jevidako: From 'https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-user-events-trace/src/trace/exporter.rs'

I've found following - 
            _eb.add_value(
                "kind",
                match span.span_kind {
                    SpanKind::Internal => 0,
                    SpanKind::Server => 1,
                    SpanKind::Client => 2,
                    SpanKind::Producer => 3,
                    SpanKind::Consumer => 4,
                },
                FieldFormat::UnsignedInt,
                0,
            );_

Changed mapping to start from 0 (instead of 1), to align with 'user-events trace'

opentelemetry-etw-traces/src/exporter/part_c.rs
jevidako: I’ll remove events for now - just noticed the announcement on 'https://opentelemetry.io/blog/2026/deprecating-span-events/'

Since one of our Rust projects still relies on events (while migration to newer versions is not completed) would it be possible to introduce this as an optional feature in a follow-up version, if needed?

cijothomas: I don't think we need to support this in the exporter. For maintaining support for the existing projects, we can discuss some options to maintain compatibility. Will chat offline for that.

opentelemetry-etw-traces/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+license = "Apache-2.0"
+edition = "2021"
+rust-version = "1.75.0"
+
+[dependencies]
+chrono = { version = "0.4.40", default-features = false, features = ["alloc", "std"] }
+opentelemetry = { workspace = true, features = ["trace"] }
+opentelemetry_sdk = { workspace = true, features = ["trace"] }
+tracelogging_dynamic = "1.2.4"
+tracing = { version = "0.1", optional = true }
```
cijothomas: not required right? this is likely why cargo machete is failing in CI.

opentelemetry-etw-traces/src/exporter/part_b.rs
```diff
@@ -0,0 +1,165 @@
...
+/// Layout (TLD struct, dynamic field count):
+/// ```text
+/// PartB {
+///     _typeName: "Span"
+///     name: str8
+///     kind: u8
+///     startTime: str8 (RFC3339)
+///     [parentId: str8]                      // only for non-root spans
+///     [links: str8 (JSON)]                  // only if present
+///     [statusMessage: str8]                 // only if status has description
```
cijothomas, line 20: statusMessage is only conditional based on whether httpStatusCode is in or not. The open PR in user-events trace is addressing this, so you can copy the idea form it. 
Alternatively, leave a TODO for this to unblock this PR quickly.

opentelemetry-etw-traces/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+serde_json = "1.0"
```
cijothomas: lets keep it conditional, like etw-logs does.

opentelemetry-etw-traces/src/exporter/otel_id_ext.rs
```diff
@@ -0,0 +1,147 @@
+//! Zero-allocation hex string conversions for OpenTelemetry [`TraceId`] and [`SpanId`].
```
cijothomas: I like the direction of avoiding heap allocation for traceid/spanid. I think this is too much boilerplate code to maintain too.
For this PR - lets stick with the allocating to_string() option for now, similar to eta-logs and user-events-traces does. 

And separately a common solution can be found and then applied to all of them consistently.

opentelemetry-etw-traces/src/exporter/part_c.rs
```diff
@@ -0,0 +1,75 @@
...
+) {
+    let resource_attr_count = resource.attributes_from_resource.len();
+    let span_partc_attr_count = span_data
+        .attributes
+        .iter()
+        .filter(|kv| common::well_known_part_b_field(kv.key.as_str()).is_none())
+        .count();
+    let part_c_count = resource_attr_count + span_partc_attr_count;
+
+    event.add_struct(
```
cijothomas: should we do this only if the count is >0 right?

opentelemetry-etw-traces/src/exporter/part_b.rs
jevidako: Thanks for the input, I've left TODO for now, and it can be addressed in the following PRs.

opentelemetry-etw-traces/CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
...
+Released 2026-Mar-11
+
+- Initial release.
+- `Processor` implementing `SpanProcessor` that exports spans to ETW using TraceLogging Dynamic.
+- Common Schema v4.0 encoding (Part A / Part B / Part C).
+- Builder pattern with provider name validation (cross-compatible and ETW-only modes).
+- Configurable event name (defaults to `"Span"`).
+- Optional resource attribute promotion to Part C via `with_resource_attributes()`.
+- Span attributes exported as individually typed ETW fields.
+- Span events and links serialized as JSON.
```
cijothomas: minor: Span events are not exported, so lets update changelog to reflect it.

opentelemetry-etw-traces/src/exporter/common.rs
```diff
@@ -0,0 +1,222 @@
...
+
+/// Well-known OpenTelemetry semantic-convention attributes that map to
+/// Common Schema Part B field names. Attributes whose key matches one of
+/// these entries are emitted as Part B fields (using the mapped name)
+/// instead of Part C.
+///
+/// Stored as a static slice (rather than a `HashMap`) to guarantee a
+/// stable, deterministic ordering of fields in the emitted ETW event.
+pub(crate) const WELL_KNOWN_PART_B_ATTRIBUTES: &[(&str, &str)] = &[
+    // Database
```
cijothomas, line 13: this was updated in user-events-traces too recently - can you pick the latest?

opentelemetry-etw-traces/src/exporter/options.rs
cijothomas: Got it, let's keep with_event_name — this matches the .NET Geneva exporter's TableNameMappings for traces, which has shipped as Stable for years with the same shape (one configurable table name per exporter, defaulting to Span).

Minor tweaks please:

1. Update the rustdoc on with_event_name  to make it explicit that this sets the single ETW EventName used for every exported span - there is no per-span override. 
2. Avoid the phrase "span events" - it collides with the OTel SpanEvent concept.

cijothomas (approved):
LGTM! Thanks for working on this

PR #565, opened 2026-03-21
feat(geneva-uploader): Add On Behalf Of (OBO) support

gscalderon:
## Changes
   
Add On Behalf Of (OBO) support to the Geneva Uploader, enabling data to be uploaded on behalf of a different service identity.
   
### What OBO does
When a monitoring agent collects data from multiple services, OBO allows tagging each upload with the originating service's identity so Geneva routes/stores data under the correct service.
   
### Changes made
1. **Config**: Added `obo_identity` and `obo_annotations` optional fields to `GenevaClientConfig`
2. **Upload URI**: Appends `&onbehalfid=` and `&onbehalfannotations=` (URL-encoded) query parameters to the GIG upload URI when OBO is configured
3. **Bond schema**: Adds `onbehalfServiceId` and `onbehalfAnnotations` columns to Bond schema via `determine_fields()` and `determine_span_fields()`
4. **Bond row data**: Writes OBO values into each log/span row via `write_row_data()` and `write_span_row_data()`
5. **Backward compatible**: All changes are behind `Option<String>` — no behavior change when OBO is not configured
   
### Reference
OBO logic ported from the C# PipelineAgent (AMACoreAgent) implementation.
   
* [x] CONTRIBUTING guidelines followed
* [x] Unit tests added/updated (11 new tests: 8 Bond blob tests + 3 URI tests)
* [x] Changes in public API reviewed

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
@@ -185,6 +185,27 @@ pub struct GenevaConfig {
     pub role_instance: *const c_char,
     pub auth: GenevaAuthConfig, // Active member selected by auth_method
     pub msi_resource: *const c_char, // Azure AD resource URI for MSI auth (auth methods 0, 3, 4, 5). Not used for auth methods 1, 2. Nullable.
+    pub obo_map: *const FfiOboEventMap, // Optional OBO event map. Nullable.
+}
+
+/// FFI-safe OBO event configuration
+#[repr(C)]
+pub struct FfiOboEventConfig {
+    /// Event name (null-terminated UTF-8 string)
+    pub event_name: *const c_char,
+    /// OBO identity (null-terminated UTF-8 string)
+    pub identity: *const c_char,
+    /// OBO annotations (null-terminated UTF-8 string, can be null)
+    pub annotations: *const c_char,
+}
+
+/// FFI-safe OBO event map (array of configs)
+#[repr(C)]
+pub struct FfiOboEventMap {
+    /// Pointer to array of FfiOboEventConfig
+    pub entries: *const FfiOboEventConfig,
+    /// Number of entries
+    pub count: usize,
 }
 
 /// Error codes returned by FFI functions
@@ -263,6 +284,65 @@ unsafe fn write_error_if_provided(
         unsafe {
             *err_msg_out.add(bytes_to_copy) = 0;
         }
+    }
+}
+
+/// Converts an FFI OBO event map to the Rust-side `OboEventMap`.
+///
+/// Returns `None` when the pointer is null, the entries pointer is null,
+/// or the count is zero. Individual entries with null event_name or identity
+/// are silently skipped.
+///
+/// # Safety
+/// - `ffi_map` must be null or point to a valid `FfiOboEventMap`
+/// - All non-null string pointers inside entries must be valid null-terminated UTF-8
+unsafe fn convert_obo_event_map(
+    ffi_map: *const FfiOboEventMap,
+) -> Option<geneva_uploader::client::OboEventMap> {
+    if ffi_map.is_null() {
+        return None;
+    }
+    let map_ref = unsafe { &*ffi_map };
+    if map_ref.entries.is_null() || map_ref.count == 0 {
+        return None;
+    }
+
+    let mut obo_map = std::collections::HashMap::new();
+    let entries = unsafe { std::slice::from_raw_parts(map_ref.entries, map_ref.count) };
+
+    for entry in entries {
+        if entry.event_name.is_null() || entry.identity.is_null() {
+            continue;
+        }
+        let event_name = unsafe { CStr::from_ptr(entry.event_name) }
+            .to_string_lossy()
+            .into_owned();
+        let identity = unsafe { CStr::from_ptr(entry.identity) }
+            .to_string_lossy()
+            .into_owned();
+        let annotations = if entry.annotations.is_null() {
+            None
+        } else {
+            Some(
+                unsafe { CStr::from_ptr(entry.annotations) }
+                    .to_string_lossy()
+                    .into_owned(),
+            )
+        };
+
+        obo_map.insert(
+            event_name,
+            geneva_uploader::client::OboEventConfig {
+                identity,
+                annotations,
+            },
+        );
+    }
+
+    if obo_map.is_empty() {
+        None
+    } else {
+        Some(obo_map)
     }
 }
 
@@ -511,6 +591,7 @@ pub unsafe extern "C" fn geneva_client_new(
         role_name,
         role_instance,
         msi_resource,
+        obo_event_map: unsafe { convert_obo_event_map(config.obo_map) },
     };
 
     // Create client
@@ -1553,6 +1634,7 @@ mod tests {
             role_name: "testrole".to_string(),
             role_instance: "testinstance".to_string(),
             msi_resource: None,
+            obo_event_map: None,
         };
         let client = GenevaClient::new(cfg).expect("failed to create GenevaClient with MockAuth");
         Box::new(GenevaClientHandle {
@@ -1715,6 +1797,7 @@ mod tests {
                 // The union is never accessed for SystemManagedIdentity (auth_method 0).
                 auth: std::mem::zeroed(),
                 msi_resource: ptr::null(),
+                obo_map: ptr::null(),
             };
 
             let mut out: *mut GenevaClientHandle = std::ptr::null_mut();
@@ -1749,6 +1832,7 @@ mod tests {
                 role_instance: role_instance.as_ptr(),
                 auth: std::mem::zeroed(), // Union not accessed for invalid auth method
                 msi_resource: ptr::null(),
+                obo_map: ptr::null(),
             };
 
             let mut out: *mut GenevaClientHandle = std::ptr::null_mut();
@@ -1788,6 +1872,7 @@ mod tests {
                     },
                 },
                 msi_resource: ptr::null(),
+                obo_map: ptr::null(),
             };
 
             let mut out: *mut GenevaClientHandle = std::ptr::null_mut();
@@ -1826,6 +1911,7 @@ mod tests {
                     },
                 },
                 msi_resource: ptr::null(),
+                obo_map: ptr::null(),
             };
 
             let mut out: *mut GenevaClientHandle = std::ptr::null_mut();
@@ -1864,6 +1950,7 @@ mod tests {
                     },
                 },
                 msi_resource: ptr::null(),
+                obo_map: ptr::null(),
             };
 
             let mut out: *mut GenevaClientHandle = std::ptr::null_mut();
@@ -1902,6 +1989,7 @@ mod tests {
                     },
                 },
                 msi_resource: ptr::null(),
+                obo_map: ptr::null(),
             };
 
             let mut out: *mut GenevaClientHandle = std::ptr::null_mut();
@@ -1943,6 +2031,7 @@ mod tests {
                     },
                 },
                 msi_resource: ptr::null(),
+                obo_map: ptr::null(),
             };
 
             let mut out: *mut GenevaClientHandle = std::ptr::null_mut();
@@ -1986,6 +2075,7 @@ mod tests {
                     },
                 },
                 msi_resource: ptr::null(),
+                obo_map: ptr::null(),
             };
 
             let mut out: *mut GenevaClientHandle = std::ptr::null_mut();
@@ -2070,6 +2160,7 @@ mod tests {
             role_name: "testrole".to_string(),
             role_instance: "testinstance".to_string(),
             msi_resource: None,
+            obo_event_map: None,
         };
         let client = GenevaClient::new(cfg).expect("failed to create GenevaClient with MockAuth");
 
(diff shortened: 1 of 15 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
```diff
@@ -197,6 +199,24 @@ impl GenevaUploader {
             schema_ids,
             row_count
         ).map_err(|e| GenevaUploaderError::InternalError(format!("Failed to write query string: {e}")))?;
+
+        // Append OBO query parameters if this event has OBO config
+        if let Some(config) = obo_config {
```
lalitb: If I understand the equivalent .Net code (which is our source of truth) - I believe the ingestion backend expects special value in "User-Agent" header for OBO. And the backend will reject the event without that header value. If this is true, we should be overriding the User-Agent per request when obo_config.is_some() and identity is non-empty.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -519,6 +557,17 @@ impl OtlpEncoder {
             fields.push((FIELD_BODY.into(), BondDataType::BT_STRING));
         }
 
+        // OBO fields — only if this event has OBO config
+        if obo_config.is_some() {
```
lalitb: If I understand the .Net code correctly, OBO should only be considered active when the resolved identity is meaningfully present, and blank annotations should be suppressed consistently. Using `obo_config.is_some()` here looks broader than that and may cause OBO fields to be included in the schema when they should not be. It would be safer to normalize identity/annotations once up front and use that normalized result consistently across URI emission, schema selection, and row writing.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -789,6 +853,17 @@ impl OtlpEncoder {
             }
         }
 
+        // OBO fields — only if this event has OBO config
+        if obo_config.is_some() {
```
lalitb: same as [here](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/565/changes#r3102331744)

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
```diff
@@ -511,6 +511,7 @@ pub unsafe extern "C" fn geneva_client_new(
         role_name,
         role_instance,
         msi_resource,
+        obo_event_map: None,
```
lalitb: OBO is currently not supported through the FFI surface because client creation hardcodes `obo_event_map: None`. I believe this should be fine for now, as our Windows and Linux agents don't need this feature. But better to confirm once.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -156,15 +172,23 @@ impl LogBatchAccumulator {
...
         let timestamp = record
             .time_unix_nano()
             .filter(|&t| t != 0)
             .or_else(|| record.observed_time_unix_nano())
             .unwrap_or(0);
         let routing_event_name = normalized_event_name(record).unwrap_or("Log");
 
-        let (field_info, dynamic_fields_start) = OtlpEncoder::determine_fields(record);
+        // Look up per-event OBO config
+        let obo_config = obo_event_map.and_then(|map| map.get(routing_event_name));
```
lalitb: This is a strict exact-match lookup. For example, `^Span$` would not match Span here. My understanding looking into the .Net implementation is that it normalizes simple anchored literal forms before lookup, so this may be narrower than that behavior. If that compatibility matters, consider normalizing those forms here or documenting (at API surface) that Rust expects literal keys only.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -2051,4 +2156,216 @@ mod tests {
...
+            OboEventConfig {
+                identity: identity.to_string(),
+                annotations: annotations.map(|s| s.to_string()),
+            },
+        );
+        map
+    }
+
+    #[test]
+    fn test_obo_fields_in_log_schema() {
```
lalitb, line 2276: These tests show that encoding succeeds, but I think they could be strengthened a bit to verify the intended OBO schema behavior more directly. In particular, it would help to add coverage for mixed OBO/non-OBO events and for empty or blank annotations, so we can confirm the OBO fields are included only when expected.

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
kamilon: All consumers of this library will need to support obo. We need to expose this over the FFI surface.

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
kamilon: That header check is no longer performed.

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
kamilon, line 2744: Placing this in a random place, but I don't see any tests where obo is not none.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
kamilon: Great catch here! The normalization done in the .NET code is actually a workaround. The schema for this is that these are regex matches. I've never personally seen anything but the simple anchored format so that's probably "good enough". Will recently gained access to all the current configs and should be able to scan through them to determine if we should add full regex support or not. We have told customers using the .NET codebase that we only support the simple anchored form but the other agents don't have that luxury.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -330,18 +360,22 @@ impl OtlpEncoder {
         &self,
         spans: impl IntoIterator<Item = &'a Span>,
         metadata_fields: &MetadataFields,
+        obo_event_map: Option<&OboEventMap>,
     ) -> Result<Vec<EncodedBatch>, String> {
         // All spans use "Span" as event name for routing - no grouping by span name
```
kamilon, line 402: Is this true?

gscalderon: The hardcoded `"Span"` event name was already before this PR and isn't OBO-specific. Confirmed `"Log"` and `"Span"` are defaults but can be overridden for 1P via GCS config. Tracking this separately in #591.

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
gscalderon: Added 5 FFI tests where OBO is not `None`:
 - `test_convert_obo_event_map_valid` — single entry with full data
 - `test_convert_obo_event_map_null_annotations` — entry with null annotations
 - `test_convert_obo_event_map_null` — null pointer (returns `None`)
 - `test_convert_obo_event_map_multiple_entries` — multiple entries
 - `test_client_creation_with_obo_map` — full client creation with real OBO data

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
gscalderon: Added 4 new tests:
 - `test_obo_empty_identity_not_active` — empty identity doesn't activate OBO
 - `test_obo_whitespace_identity_not_active` — whitespace-only identity doesn't activate OBO
 - `test_obo_blank_annotations_suppressed` — blank annotations are suppressed
 - `test_obo_event_config_helpers` — directly tests `is_active()` and `active_annotations()`
 
 The existing `test_mixed_batch_obo_and_non_obo_events` already covers the mixed case.

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
gscalderon: Exposed OBO on the FFI surface:
 - Added `FfiOboEventConfig` and `FfiOboEventMap` C-compatible structs
 - Added `obo_map` field on `GenevaConfig`
 - Added `convert_obo_event_map()` function that converts the FFI structs to Rust's `Option<OboEventMap>`
 - Updated the C header `geneva_ffi.h` with the corresponding typedefs

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
gscalderon: Addressed — same `is_active()` / `active_annotations()` helpers applied to `determine_span_fields` and `write_span_row_data`.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
gscalderon: Good catch! Added `is_active()` and `active_annotations()` helpers to `OboEventConfig` that normalize identity/annotations consistently:
 - `is_active()` returns `true` only when identity is non-empty after trimming
 - `active_annotations()` returns `None` when annotations are blank
 
 Updated all OBO checks to use these helpers — in `determine_fields`, `determine_span_fields`, `write_row_data`, `write_span_row_data`, `create_upload_uri`, and `upload_batch`. Also added tests covering empty identity, whitespace-only identity, and blank annotations.

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
gscalderon: No action needed here, confirmed the header check is no longer performed by the backend.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
gscalderon: Added normalization for the simple anchored form (`^EventName$`) — the lookup tries the literal event name first, then falls back to the anchored format. This matches the .NET workaround behavior.

@kamilon if Will's scan reveals more complex regex patterns in the configs, we can extend this to full regex matching. Otherwise, this should be "good enough" for the common case.

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
```diff
@@ -266,6 +287,65 @@ unsafe fn write_error_if_provided(
...
+
+    let mut obo_map = std::collections::HashMap::new();
+    let entries = unsafe { std::slice::from_raw_parts(map_ref.entries, map_ref.count) };
+
+    for entry in entries {
+        if entry.event_name.is_null() || entry.identity.is_null() {
+            continue;
+        }
+        let event_name = unsafe { CStr::from_ptr(entry.event_name) }
+            .to_string_lossy()
```
lalitb: Can we use the same UTF-8 validation path for  `event_name`, `identity`, and `annotations`, as the rest of `geneva_client_new` instead of `to_string_lossy()`? The FFI docs say these strings must be UTF-8, and the existing config fields return `InvalidConfig` on invalid UTF-8. Silently replacing bad bytes in OBO event names or identities could create a wrong OBO mapping instead of failing fast.

lalitb (approved):
Thanks @gscalderon for working on this. This looks good now.

lalitb:
@gscalderon  - Can you fix the CI please, so we can merge the PR:

```
home/runner/work/opentelemetry-rust-contrib/opentelemetry-rust-contrib/stress)
   Compiling opentelemetry-exporter-geneva v0.5.0 (/home/runner/work/opentelemetry-rust-contrib/opentelemetry-rust-contrib/opentelemetry-exporter-geneva/opentelemetry-exporter-geneva)
error[E0063]: missing field `obo_event_map` in initializer of `GenevaClientConfig`
  --> opentelemetry-exporter-geneva/geneva-uploader/tests/geneva_test_server_integration.rs:13:36
   |
13 |     let client = GenevaClient::new(GenevaClientConfig {
   |                                    ^^^^^^^^^^^^^^^^^^ missing `obo_event_map`

For more information about this error, try `rustc --explain E0063`.
error: could not compile `geneva-uploader` (test "geneva_test_server_integration") due to 1 previous error
warning: build failed, waiting for other jobs to finish...
``

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
gscalderon: Fixed in 2ab45a78 — now uses c_str_to_string for all three fields, returning GenevaError::InvalidConfig on invalid UTF-8, same as the other config fields. Added 3 new tests for invalid UTF-8 input.

gscalderon:
@lalitb addressed comments in 2ab45a78 and updated the branch — CI is green, thanks for the review and help here!

PR #577, opened 2026-04-10
fix: [Geneva uploader] fix CodeQL finding by validating TLS with a generated CA

lalitb:
## Summary

This change removes the insecure TLS certificate-validation bypass from the Geneva config client and replaces it with real regression coverage that exercises certificate verification correctly.

## What changed

- removed `danger_accept_invalid_certs(true)` from the config-service client
- removed the `self_signed_certs` default/insecure feature path
- kept the production path on normal certificate validation
- added active tests that:
  - generate a CA at runtime
  - generate CA-signed server and client certificates at runtime
  - verify the handshake succeeds when the generated CA is trusted
  - verify the handshake fails when the generated CA is not trusted

## Why

CodeQL correctly flagged the previous implementation because it disabled TLS certificate validation, which could allow man-in-the-middle attacks. The new approach preserves secure verification and proves the behavior with targeted TLS tests instead of relying on an insecure bypass.

## Verification

- `cargo fmt --all`
- `cargo test -p geneva-uploader test_tls_ -- --nocapture`

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
@@ -3,10 +3,27 @@ pub(crate) mod client;
 #[cfg(test)]
 mod tests {
     use crate::config_service::client::{AuthMethod, GenevaConfigClient, GenevaConfigClientConfig};
-    use openssl::{pkcs12::Pkcs12, pkey::PKey, x509::X509};
+    use openssl::{
+        asn1::Asn1Time,
+        bn::{BigNum, MsbOption},
+        hash::MessageDigest,
+        pkcs12::Pkcs12,
+        pkey::{PKey, Private},
+        rsa::Rsa,
+        ssl::{SslAcceptor, SslMethod, SslVersion},
+        x509::{
+            extension::{
+                AuthorityKeyIdentifier, BasicConstraints, ExtendedKeyUsage, KeyUsage,
+                SubjectAlternativeName, SubjectKeyIdentifier,
+            },
+            X509NameBuilder, X509,
+        },
+    };
     use rcgen::generate_simple_self_signed;
-    use std::io::Write;
+    use std::io::{Read, Write};
+    use std::net::TcpListener;
     use std::path::PathBuf;
+    use std::thread;
     use tempfile::NamedTempFile;
     use wiremock::matchers::{method, path};
     use wiremock::{Mock, MockServer, ResponseTemplate};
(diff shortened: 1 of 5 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
```diff
@@ -70,14 +88,236 @@
...
+    }
+
+    fn generate_ca_signed_tls_material() -> GeneratedTlsMaterial {
+        let (ca_cert, ca_key) = generate_ca();
+        let (server_cert, server_signing_key) =
+            generate_signed_leaf(&ca_cert, &ca_key, "localhost", Some("localhost"), true);
+
+        let (client_cert, client_signing_key) =
+            generate_signed_leaf(&ca_cert, &ca_key, "GenevaUploader Test Client", None, false);
+        let client_password = "test".to_string();
```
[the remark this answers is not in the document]
lalitb: This is fixed.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
```diff
@@ -134,6 +389,94 @@ mod tests {
         assert_eq!(token_endpoint, jwt_endpoint);
     }
 
+    #[tokio::test]
+    async fn test_tls_rejects_untrusted_runtime_generated_ca() {
```
utpilla: nit: Not introduced in this PR but the `test_` prefix on the test names in this file is redundant. The `#[tokio::test]` attribute is marker for the test methods.

utpilla: The new `test_tls_rejects_untrusted_runtime_generated_ca` and `test_tls_accepts_runtime_generated_ca` tests use OpenSSL-based cert generation (same as the other tests in this file), but don't have the `#[cfg_attr(target_os = "macos", ignore)]` annotation that the other tests (`test_get_ingestion_info_mocked`, `test_error_handling_with_non_success_status`, `test_missing_ingestion_gateway_info`, `test_invalid_certificate_path`) have.

Do the new tests pass on macOS, or do they need the same ignore annotation?

utpilla (approved):

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
lalitb: These two new tests do pass on macOS, so I intentionally left off the `#[cfg_attr(target_os = "macos", ignore)]`.

The older ignored tests in this file are using the older generated PKCS#12 / `native_tls` path, which is what is not compatible on macOS. I will fix them separately.

lalitb: Good point. fixed now.

PR #588, opened 2026-04-24
fix: wrap role/roleInstance in ext_cloud struct in PartA (fixes #584)

### LINKED ISSUE #584: ETW logs exporter: service.name and service.instance.id not wrapped in ext_cloud struct in PartA
The `opentelemetry-etw-logs` exporter writes `service.name` and `service.instance.id` resource attributes as flat `role` and `roleInstance` fields directly inside PartA, without wrapping them in an `ext_cloud` struct.

In contrast, the user-events exporters (`opentelemetry-user-events-logs` and `opentelemetry-user-events-trace`) correctly export these as `ext_cloud_role` and `ext_cloud_roleInstance`, following the Common Schema convention.

**Current behavior (ETW logs):**
```
PartA.role = "my-service"
PartA.roleInstance = "instance-1"
```

**Expected behavior (matching user-events exporters):**
```
PartA.ext_cloud.role = "my-service"
PartA.ext_cloud.roleInstance = "instance-1"
```

Since ETW uses nested structs (e.g., `ext_dt` already wraps `traceId`/`spanId`), the fix would add an `ext_cloud` struct wrapper around these fields.

**Relevant code:**
- ETW logs: `opentelemetry-etw-logs/src/exporter/part_a.rs` (`populate_resource` function)
- User-events logs (correct): `opentelemetry-user-events-logs/src/logs/exporter.rs`
- User-events trace (correct): `opentelemetry-user-events-trace/src/trace/exporter.rs`

cijothomas:
Fixes #584

## Summary

Wrap `service.name` and `service.instance.id` resource attributes inside an `ext_cloud` struct in PartA, matching the Common Schema convention used by the user-events exporters.

## Before (bug)

```
PartA.role = "my-service"
PartA.roleInstance = "instance-1"
```

## After (fix)

```
PartA.ext_cloud.role = "my-service"
PartA.ext_cloud.roleInstance = "instance-1"
```

### `opentelemetry-etw-logs/src/exporter/part_a.rs`
- **`get_resource_count`**: Returns 1 if either `cloud_role` or `cloud_role_instance` is present (the `ext_cloud` struct counts as one field in PartA), 0 otherwise.
- **`populate_resource`**: Wraps `role`/`roleInstance` in an `event.add_struct("ext_cloud", ...)` call, following the same pattern as `ext_dt` for `traceId`/`spanId`.

### `opentelemetry-etw-logs/src/lib.rs`
- Updated integration test assertions from `PartA.role` → `PartA.ext_cloud.role` and `PartA.roleInstance` → `PartA.ext_cloud.roleInstance`.

### DIFF
--- a/opentelemetry-etw-logs/CHANGELOG.md
+++ b/opentelemetry-etw-logs/CHANGELOG.md
@@ -3,6 +3,16 @@
 ## vNext
 
 - Fixed a panic that would trigger if logging from inside a blocked on async block due to nested `block_on()`s.
+- **Bug fix**: `service.name` and `service.instance.id` resource attributes are
+  now correctly wrapped in an `ext_cloud` struct in PartA
+  (`PartA.ext_cloud.role`, `PartA.ext_cloud.roleInstance`), matching the Common
+  Schema convention used by the user-events exporters. Previously these were
+  emitted as flat fields directly in PartA (`PartA.role`,
+  `PartA.roleInstance`). **Note**: if you have downstream consumers that relied
+  on the previous flat field layout, this is a breaking change for those
+  consumers.
+- Added integration tests that validate the ETW event payload end-to-end using
+  a live ETW listener.
 
 ## v0.10.1
 
--- a/opentelemetry-etw-logs/CHANGELOG.md
+++ b/opentelemetry-etw-logs/CHANGELOG.md
@@ -11,8 +11,6 @@
   `PartA.roleInstance`). **Note**: if you have downstream consumers that relied
   on the previous flat field layout, this is a breaking change for those
   consumers.
-- Added integration tests that validate the ETW event payload end-to-end using
-  a live ETW listener.
 
 ## v0.10.1
 
--- a/opentelemetry-etw-logs/CHANGELOG.md
+++ b/opentelemetry-etw-logs/CHANGELOG.md
@@ -3,14 +3,15 @@
 ## vNext
 
 - Fixed a panic that would trigger if logging from inside a blocked on async block due to nested `block_on()`s.
-- **Bug fix**: `service.name` and `service.instance.id` resource attributes are
-  now correctly wrapped in an `ext_cloud` struct in PartA
-  (`PartA.ext_cloud.role`, `PartA.ext_cloud.roleInstance`), matching the Common
-  Schema convention used by the user-events exporters. Previously these were
-  emitted as flat fields directly in PartA (`PartA.role`,
-  `PartA.roleInstance`). **Note**: if you have downstream consumers that relied
-  on the previous flat field layout, this is a breaking change for those
-  consumers.
+- **Bug fix**: `service.name` and `service.instance.id` resource attributes
+  were previously emitted as bare fields in PartA (`PartA.role`,
+  `PartA.roleInstance`) without the required `ext_cloud` namespace. They are
+  now emitted inside a nested `ext_cloud` struct (`PartA.ext_cloud.role`,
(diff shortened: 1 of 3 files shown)


opentelemetry-etw-logs/src/exporter/part_a.rs
```diff
@@ -62,21 +62,25 @@ fn populate_part_a_from_context(
 }
 
 fn get_resource_count(resource: &super::Resource) -> u8 {
```
utpilla: Let's rename this method to match the new logic.

```suggestion
fn has_ext_cloud(resource: &super::Resource) -> u8 {
```

opentelemetry-etw-logs/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
...
 
 - Fixed a panic that would trigger if logging from inside a blocked on async block due to nested `block_on()`s.
+- **Bug fix**: `service.name` and `service.instance.id` resource attributes are
+  now correctly wrapped in an `ext_cloud` struct in PartA
+  (`PartA.ext_cloud.role`, `PartA.ext_cloud.roleInstance`), matching the Common
+  Schema convention used by the user-events exporters. Previously these were
+  emitted as flat fields directly in PartA (`PartA.role`,
+  `PartA.roleInstance`). **Note**: if you have downstream consumers that relied
+  on the previous flat field layout, this is a breaking change for those
+  consumers.
```
utpilla: The entry seems to conflate two things:

1. **The actual bug**: `role` and `roleInstance` were emitted as bare fields directly in PartA (`PartA.role`, `PartA.roleInstance`) with no `ext_cloud` association. This doesn't conform to Common Schema regardless of encoding style (flat fields vs nested struct). The fix for this bug could have been simply adding the `ext_cloud` prefix to the flat field names (i.e., `ext_cloud_role`, `ext_cloud_roleInstance`). No structural change was required.

2. **An encoding design choice**: Instead of flat prefixed names, this PR uses a nested struct (`ext_cloud { role, roleInstance }`). Per the Common Schema, PartA extensions can be provided either as a nested struct or as flat fields (both are valid,) with nested structs being consistent with how `ext_dt` is already handled in this crate, but it is a separate design decision from the bug fix itself.


For downstream consumers upgrading, there are two breaking changes at once: the addition of the `ext_cloud` namespace (the bug fix) and the switch to a nested struct encoding (the design choice). Both could independently break consumers depending on what they relied on.

Also, the entry says this matches "the Common Schema convention used by the user-events exporters," but the user-events-logs crate actually uses flat field names (`ext_cloud_role`, `ext_dt_traceId`), not nested structs.

Suggested rewording:

> **Bug fix**: `service.name` and `service.instance.id` resource attributes were previously emitted as bare fields in PartA (`PartA.role`, `PartA.roleInstance`) without the required `ext_cloud` namespace. They are now emitted inside a nested `ext_cloud` struct (`PartA.ext_cloud.role`, `PartA.ext_cloud.roleInstance`), consistent with how `ext_dt` is already structured in this crate and following the Common Schema TLD mapping spec's preferred nested struct convention. **Note**: downstream consumers that looked for the bare field names (`role`, `roleInstance`) in PartA will need to be updated.

utpilla (approved):
Approved with suggestions.

opentelemetry-etw-logs/CHANGELOG.md
utpilla: > Also, the entry says this matches "the Common Schema convention used by the user-events exporters," but the user-events-logs crate actually uses flat field names (ext_cloud_role, ext_dt_traceId), not nested structs

Created an issue #589 to track the discrepancy.

PR #514, opened 2026-01-06
fix: Make etw logs exporter synchronous

berkus:
Remove block_on calls and remove otel Exporter trait impl.

Fixes #511

## Changes

ETW Logs exporter now does not invoke asynchronous Exporter impl under the hood. The call was a fake (using block_on) anyway, and it could cause problems if, for example, a tracing log macro was called under another block_on() async block - nesting block_on()s is forbidden.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Unit tests added/updated (if applicable)
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [x] Changes in public API reviewed (if applicable)

PR #590, opened 2026-04-25
fix: remove block_on from user-events-logs processor (fixes #515)

### LINKED ISSUE #515: user-events exporter - model as processor
Similar to etw-logs, for same bug.
https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/514

cijothomas:
Fixes #515

## Summary

Remove `futures_executor::block_on` from the user-events-logs processor, making it fully synchronous. This is the same change that PR #514 applied to `opentelemetry-etw-logs`.

## Motivation

The `Processor::emit()` was calling `futures_executor::block_on(self.exporter.export(...))` to invoke the async `LogExporter` trait. This can panic if called from inside another `block_on()` (nested runtime), e.g. when logging from within an async runtime context.

Since the exporter is inherently synchronous (writes directly to a user_events tracepoint with no I/O), the async wrapper was unnecessary.

### `opentelemetry-user-events-logs/src/logs/exporter.rs`
- Removed `impl LogExporter for UserEventsExporter` block
- Moved `shutdown()`, `event_enabled()`, and `set_resource()` to direct `pub(crate)` methods on `UserEventsExporter` (alongside the existing `export_log_data()`)

### `opentelemetry-user-events-logs/src/logs/processor.rs`
- `emit()` now calls `self.exporter.export_log_data(record, scope)` directly instead of `block_on(self.exporter.export(...))`
- Removed `LogExporter`, `LogBatch`, and `futures_executor` imports

### `opentelemetry-user-events-logs/Cargo.toml`
- Removed `futures-executor` from dependencies

### `opentelemetry-user-events-logs/CHANGELOG.md`
- Added entry documenting the change

### DIFF
--- a/opentelemetry-user-events-logs/CHANGELOG.md
+++ b/opentelemetry-user-events-logs/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## vNext
 
+- Removed `futures-executor` dependency and `LogExporter` trait implementation.
+  The processor now calls the exporter synchronously, avoiding potential panics
+  from nested `block_on()` calls (e.g. when logging from inside an async
+  runtime).
+
 ## v0.16.0
 
 Released 2026-May-13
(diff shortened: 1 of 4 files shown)


lalitb (approved):
thanks.

PR #592, opened 2026-04-28
feat: [Geneva uploader]   Support managed identity auth by resource ID

lalitb:
This change lets `geneva-uploader` get a managed identity token by Azure resource ID when a local managed identity endpoint is available.
                                                                                                              
  This is needed for environments such as Azure Arc extensions, where managed identity is exposed through a local endpoint and the desired identity is selected by resource ID.                               
                                                                                                                                                                                                              
  The code is not hardcoded to Azure Arc. It uses the standard local managed identity endpoint contract exposed through:                                                                                      
                                                                                                                                                                                                              
  ```text                                                                                                                                                                                                     
  IDENTITY_ENDPOINT
  IDENTITY_HEADER
  ```

  ## Flow

  ```mermaid                                                                                                                                                                                                  
  sequenceDiagram 
      participant U as geneva-uploader
      participant MSI as Local managed identity endpoint                                                                                                                                                      
      participant Entra as Microsoft Entra ID
      participant GCS as Geneva Config Service                                                                                                                                                                
                  
      U->>MSI: GET $IDENTITY_ENDPOINT                                                                                                                                                                         
      Note over U,MSI: api-version=2018-02-01<br/>resource=msi_resource<br/>msi_res_id=resource_id<br/>Metadata: true<br/>X-IDENTITY-HEADER: header
      MSI->>MSI: Validate caller and identity selector                                                                                                                                                        
      MSI->>Entra: Request token for selected managed identity                                                                                                                                                
      Entra-->>MSI: Access token                                                                                                                                                                              
      MSI-->>U: JSON response with access_token                                                                                                                                                               
      U->>GCS: GET MonitoringStorageKeys                                                                                                                                                                      
      Note over U,GCS: Authorization: Bearer access_token
      GCS-->>U: Ingestion configuration                                                                                                                                                                       
  ```                                                                                                                                                                                                     
                  
  ## What changes                                                                                                                                                                                             
                  
  Today, this auth method relies on `azure_identity` to get the token. That does not work in all local MSI endpoint environments when the identity is selected by resource ID.                                
   
  With this change, when these environment variables are present:                                                                                                                                             
                  
  ```text
  IDENTITY_ENDPOINT
  IDENTITY_HEADER
  ```

  `geneva-uploader` calls the local MSI endpoint directly:                                                                                                                                                    
   
  ```text                                                                                                                                                                                                     
  GET $IDENTITY_ENDPOINT
    ?api-version=2018-02-01
    &resource=<msi_resource>
    &msi_res_id=<resource_id>                                                                                                                                                                                 
  Metadata: true
  X-IDENTITY-HEADER: <IDENTITY_HEADER>                                                                                                                                                                        
  ```             

  The response contains an `access_token`. `geneva-uploader` uses that token as the bearer token when calling Geneva Config Service.                                                                          
   
  If those environment variables are not present, the existing `azure_identity` flow is still used.                                                                                                           
                  
  ## Local MSI response                                                                                                                                                                                       
                  
  The local MSI endpoint returns the standard managed identity token response shape.

  Example:

  ```json
  {
    "access_token": "<access token>",
    "expires_on": "1777334267",                                                                                                                                                                               
    "resource": "https://monitor.core.windows.net",
    "token_type": "Bearer"                                                                                                                                                                                    
  }               
  ```

  `geneva-uploader` only reads `access_token` from this response.                                                                                                                                             
   
  It does not log the token or persist it. The token is kept in memory and used as:                                                                                                                           
                  
  ```text
  Authorization: Bearer <access_token>
  ```

  when calling Geneva Config Service.                                                                                                                                                                         
   
  ## Why                                                                                                                                                                                                      
                  
  Azure managed identity supports selecting an identity by resource ID using the `msi_res_id` query parameter.

  Some environments, including Azure Arc extensions, expose managed identity through a local endpoint using `IDENTITY_ENDPOINT` and `IDENTITY_HEADER`. In those environments, using `msi_res_id` is the       
  correct way to request a token for a resource-ID-selected identity.
                                                                                                                                                                                                              
  This change adds that path explicitly while keeping the previous SDK-based behavior as fallback.

  ## Retry behavior

  This change does not add a new retry loop around local MSI token acquisition.                                                                                                                               
   
  The request is made through the existing `reqwest` client path. If the local MSI endpoint returns a non-success status, `geneva-uploader` returns an MSI auth error and the existing caller/export retry    
  behavior applies.
                                                                                                                                                                                                              
  This keeps the first implementation small and avoids hiding configuration errors such as:

  - missing or invalid identity header                                                                                                                                                                        
  - invalid `resource_id`
  - identity not assigned or not available                                                                                                                                                                    
  - invalid `msi_resource`

  Future improvement: add targeted retry for transient local MSI endpoint statuses that are recommended by the managed identity documentation, such as:                                                       
   
  - 404                                                                                                                                                                                                       
  - 410           
  - 429
  - 5xx
  - timeout / connection reset

  That retry should use bounded exponential backoff and should not retry permanent 4xx errors.                                                                                                                
   
  ## Safety                                                                                                                                                                                                   
                  
  This does not log tokens or the identity header.

  The token is only kept in memory and is used only as:

  ```text
  Authorization: Bearer <token>
  ```
                                                                                                                                                                                                              
  for the Geneva Config Service request.
                                                                                                                                                                                                              
  ## Tests        

  Added a unit test that verifies:

  - the local MSI endpoint is called
  - `msi_res_id` is sent
  - `client_id` is not sent
  - `Metadata` and `X-IDENTITY-HEADER` headers are sent                                                                                                                                                       
  - the returned token is used as the bearer token for the Geneva Config Service request
                                                                                                                                                                                                              
  Validation:                                                                                                                                                                                                 
   
  ```bash                                                                                                                                                                                                     
  cargo test -p geneva-uploader --lib
  ```

  Result:

  ```text
  38 passed; 0 failed; 9 ignored
  ```                                                    


  ## References
                                                                                                                                                                                                              
  - Local managed identity endpoint contract (`IDENTITY_ENDPOINT`, `IDENTITY_HEADER`, `X-IDENTITY-HEADER`):                                                                                                   
    https://learn.microsoft.com/en-us/azure/app-service/overview-managed-identity#rest-endpoint-reference                                                                                                     
  - Identity selection by Azure resource ID using `msi_res_id`:                                                                                                                                               
    https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-managed-identities-work-vm

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
@@ -634,10 +634,10 @@ impl GenevaConfigClient {
             return Ok(None);
         }
 
-        // azure_identity 0.29 detects IDENTITY_ENDPOINT + IDENTITY_HEADER as App Service
-        // and rejects UserAssignedId::ResourceId. Local MSI endpoints used by some
-        // environments, such as Azure Arc extensions, support resource ID selection
-        // through msi_res_id.
+        // `azure_identity` 0.29 treats IDENTITY_ENDPOINT + IDENTITY_HEADER as the App Service
+        // local endpoint and does not forward `UserAssignedId::ResourceId` as `msi_res_id`.
+        // Selecting a user-assigned identity by Azure resource ID against this local endpoint
+        // requires sending `msi_res_id` directly, so this branch issues the request explicitly.
         let msi_resource = resource.trim_end_matches("/.default");
         let response = self
             .http_client
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
@@ -1,4 +1,4 @@
-// Geneva Config Client with TLS (PKCS#12) and Azure Workload Identity support TODO: Azure Arc support
+// Geneva Config Client with TLS (PKCS#12), Azure Workload Identity, and Managed Identity support.
 
 use base64::{engine::general_purpose, Engine as _};
 use reqwest::{
@@ -203,6 +203,11 @@ struct GenevaResponse {
     // Keep tag_id as it might be used for validation
     #[serde(rename = "TagId")]
     tag_id: String,
+}
+
+#[derive(Debug, Deserialize)]
+struct MsiTokenResponse {
+    access_token: String,
 }
 
 #[allow(dead_code)]
@@ -537,6 +542,17 @@ impl GenevaConfigClient {
             scope_candidates.push(format!("{base}/"));
         }
 
+        if let AuthMethod::UserManagedIdentityByResourceId { resource_id } =
+            &self.config.auth_method
+        {
+            if let Some(token) = self
+                .try_get_local_managed_identity_token_by_resource_id(resource, resource_id)
+                .await?
+            {
+                return Ok(token);
+            }
+        }
+
         let user_assigned_id = match &self.config.auth_method {
             AuthMethod::SystemManagedIdentity => None,
             AuthMethod::UserManagedIdentity { client_id } => {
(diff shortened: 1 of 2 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -537,6 +542,17 @@ impl GenevaConfigClient {
...
+            &self.config.auth_method
+        {
+            if let Some(token) = self
+                .try_get_local_managed_identity_token_by_resource_id(resource, resource_id)
+                .await?
+            {
+                return Ok(token);
+            }
+        }
+
```
utpilla: Could we move this to the match statement below?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -600,6 +616,91 @@ impl GenevaConfigClient {
         )))
     }
 
+    async fn try_get_local_managed_identity_token_by_resource_id(
+        &self,
+        resource: &str,
+        resource_id: &str,
+    ) -> Result<Option<String>> {
+        let Ok(identity_endpoint) = std::env::var("IDENTITY_ENDPOINT") else {
```
utpilla, line 619: Minor: Could we validate that endpoint is a valid URL? `reqwest` will return an error for malformed URLs, which would surface as `GenevaConfigClientError::Http` rather than a more descriptive error.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -1,4 +1,4 @@
-// Geneva Config Client with TLS (PKCS#12) and Azure Workload Identity support TODO: Azure Arc support
+// Geneva Config Client with TLS (PKCS#12), Azure Workload Identity, and Managed Identity support.
```
utpilla: We could be more specific here.

```suggestion
// Geneva Config Client with TLS (PKCS#12), Azure Workload Identity, and Azure Arc/local Managed Identity support.
```

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -600,6 +616,91 @@ impl GenevaConfigClient {
...
+        if !response.status().is_success() {
+            let status = response.status();
+            debug!(
+                name: "config_client.get_local_msi_token.failed",
+                target: "geneva-uploader",
+                status = %status.as_u16(),
+                "Local Managed Identity token request returned non-success status"
+            );
+            return Err(GenevaConfigClientError::MsiAuth(format!(
+                "Local Managed Identity token request failed with status {status}"
```
utpilla: Log the body as well to retain any useful error message from the MSI endpoint.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -600,6 +616,91 @@ impl GenevaConfigClient {
...
+        };
+        if identity_header.is_empty() {
+            return Ok(None);
+        }
+
+        // `azure_identity` 0.29 treats IDENTITY_ENDPOINT + IDENTITY_HEADER as the App Service
+        // local endpoint and does not forward `UserAssignedId::ResourceId` as `msi_res_id`.
+        // Selecting a user-assigned identity by Azure resource ID against this local endpoint
+        // requires sending `msi_res_id` directly, so this branch issues the request explicitly.
+        let msi_resource = resource.trim_end_matches("/.default");
```
utpilla: The caller (`get_msi_token`) already computed base with the same trim. Passing `base` instead would be cleaner.

utpilla (approved):

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -600,6 +616,91 @@ impl GenevaConfigClient {
...
+            return Ok(None);
+        }
+        let Ok(identity_header) = std::env::var("IDENTITY_HEADER") else {
+            return Ok(None);
+        };
+        if identity_header.is_empty() {
+            return Ok(None);
+        }
+
+        // `azure_identity` 0.29 treats IDENTITY_ENDPOINT + IDENTITY_HEADER as the App Service
```
utpilla, line 635: Is this issue related? https://github.com/Azure/azure-sdk-for-rust/issues/2407

Should we add a note in the comment to prompt re-evaluation on upgrade:

```rust
// `azure_identity` 0.29 treats IDENTITY_ENDPOINT + IDENTITY_HEADER as the App Service
// local endpoint and does not forward `UserAssignedId::ResourceId` as `msi_res_id`.
// See https://github.com/Azure/azure-sdk-for-rust/issues/2407
// TODO: Re-evaluate this workaround when upgrading azure_identity beyond 0.29.
```

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Made it as - 
// Geneva Config Client with TLS (PKCS#12), Azure Workload Identity, and Azure Managed Identity support, including local endpoint resource-ID flow.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Yes - good point. Added now.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Yes, this is related. I added the issue link and a TODO to re-evaluate the workaround when upgrading azure_identity beyond 0.29.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Good point. done now.

PR #632, opened 2026-06-04
feat(user-events-trace): Add links, statusMessage, rpc attrs and fix time format

cijothomas:
Improve Common Schema v4.0 compliance for the user-events-trace exporter.

- Add `links` field to Part B (serialized as JSON array of `{toTraceId, toSpanId}` per CS ToSpanLink type)
- Add `statusMessage` field to Part B for error spans (suppressed when `httpStatusCode` is present, per CS spec)
- **Breaking**: Update well-known attribute mappings to use stable OTel semantic conventions. The CS output field names are unchanged, but the OTel attribute keys that trigger the mapping have been updated:
  - `db.system` → `db.system.name`, `db.name` → `db.namespace`, `db.statement` → `db.query.text`, `messaging.destination` → `messaging.destination.name`
  - Added `rpc.system` → `rpcSystem`, `rpc.grpc.status_code` → `rpcGrpcStatusCode`
- Fix `PartA.time` and `PartB.startTime` to use trailing `Z` instead of `+00:00` (per CS spec: UTC ISO 8601 with trailing Z)
- Optimize `links_to_json` to avoid intermediate String allocations
- Expand integration test coverage (8 tests) and add unit tests

### DIFF
--- a/opentelemetry-user-events-trace/src/trace/exporter.rs
+++ b/opentelemetry-user-events-trace/src/trace/exporter.rs
@@ -270,7 +270,9 @@ impl UserEventsSpanExporter {
                 .iter()
                 .any(|kv| kv.key.as_str() == "http.response.status_code");
             let status_message = match &span.status {
-                Status::Error { description } if !description.is_empty() && !has_http_status_code => {
+                Status::Error { description }
+                    if !description.is_empty() && !has_http_status_code =>
+                {
                     Some(description.to_string())
                 }
                 _ => None,
--- a/opentelemetry-user-events-trace/src/trace/exporter.rs
+++ b/opentelemetry-user-events-trace/src/trace/exporter.rs
@@ -11,7 +11,7 @@ use opentelemetry_sdk::trace::SpanData;
 use opentelemetry_sdk::Resource;
 use std::{
     collections::HashMap,
-    fmt::{self, Debug, Write},
+    fmt::{Debug, Write},
     sync::{Arc, Mutex, OnceLock},
 };
 
@@ -41,8 +41,8 @@ fn get_well_known_attributes() -> &'static HashMap<&'static str, &'static str> {
         map.insert("messaging.url", "messagingUrl");
 
         // RPC attributes
-        map.insert("rpc.system", "rpcSystem");
-        map.insert("rpc.grpc.status_code", "rpcGrpcStatusCode");
+        map.insert("rpc.system.name", "rpcSystem");
+        map.insert("rpc.response.status_code", "rpcGrpcStatusCode");
 
         map
     })
@@ -50,8 +50,10 @@ fn get_well_known_attributes() -> &'static HashMap<&'static str, &'static str> {
 
 /// Serializes span links to a JSON string: `[{"toTraceId":"...","toSpanId":"..."},...]`
 ///
-/// Writes TraceId/SpanId hex directly into the buffer via `Display`,
-/// avoiding intermediate `String` allocations.
+/// Writes TraceId/SpanId via Display directly into the output buffer,
+/// avoiding temporary string allocations.
+/// TODO(perf): Revisit link encoding to reduce allocations further and avoid
+/// JSON-text serialization overhead where possible.
 fn links_to_json(links: &[opentelemetry::trace::Link]) -> String {
     // Pre-allocate: each link is ~80 chars: {"toTraceId":"<32>","toSpanId":"<16>"}
     let mut json = String::with_capacity(2 + links.len() * 80);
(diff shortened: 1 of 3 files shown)


opentelemetry-user-events-trace/src/trace/exporter.rs
```diff
@@ -40,10 +40,31 @@ fn get_well_known_attributes() -> &'static HashMap<&'static str, &'static str> {
         map.insert("messaging.destination", "messagingDestination");
         map.insert("messaging.url", "messagingUrl");
 
+        // RPC attributes
+        map.insert("rpc.system", "rpcSystem");
+        map.insert("rpc.grpc.status_code", "rpcGrpcStatusCode");
```
utpilla: Could we use the newer key names as specified in the semantic conventions?

It looks like even the db and messaging attribute keys should be updated.

opentelemetry-user-events-trace/src/trace/exporter.rs
```diff
@@ -40,10 +40,31 @@ fn get_well_known_attributes() -> &'static HashMap<&'static str, &'static str> {
...
+        }
+        json.push_str("{\"toTraceId\":\"");
+        json.push_str(&link.span_context.trace_id().to_string());
+        json.push_str("\",\"toSpanId\":\"");
+        json.push_str(&link.span_context.span_id().to_string());
+        json.push_str("\"}");
+    }
+    json.push(']');
+    json
+}
```
utpilla, line 73: This method could be optimized.  It currently allocates two throwaway `Strings` per link (one each for `trace_id().to_string()` / `span_id().to_string()`) and starts `json` string at capacity `1`, so it reallocates as it grows.

Since `TraceId` and `SpanId `both implement `Display` as fixed-width lowercase hex (32 / 16 chars), we can format directly into a pre-sized buffer and skip the intermediate allocations.

opentelemetry-user-events-trace/src/trace/exporter.rs
cijothomas: update to use the stable conventions only. Given alpha nature, we don't need to worry about keeping old conventions at all. PTAL.

opentelemetry-user-events-trace/src/trace/exporter.rs
cijothomas: ya I noticed this in eta-trace PR too. Given alpha stage, didn't spend much time on perf, but this is an easy fix.

opentelemetry-user-events-trace/src/trace/exporter.rs
utpilla: It looks rpc attributes were not updated.

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/rpc-spans.md

opentelemetry-user-events-trace/src/trace/exporter.rs
```diff
@@ -231,6 +267,36 @@ impl UserEventsSpanExporter {
...
+            // statusMessage should not be reported."
+            let has_http_status_code = span
+                .attributes
+                .iter()
+                .any(|kv| kv.key.as_str() == "http.response.status_code");
+            let status_message = match &span.status {
+                Status::Error { description }
+                    if !description.is_empty() && !has_http_status_code =>
+                {
+                    Some(description.to_string())
```
utpilla: `description.to_string()` causes an allocation per error span. `eb.add_str` only needs `&str` so we should be able to remove the allocation.

opentelemetry-user-events-trace/src/trace/exporter.rs
```diff
@@ -231,6 +267,36 @@ impl UserEventsSpanExporter {
                 );
             }
 
+            // statusMessage: emit error description if present.
+            // Per Common Schema spec: "If you report httpStatusCode,
+            // statusMessage should not be reported."
+            let has_http_status_code = span
+                .attributes
+                .iter()
+                .any(|kv| kv.key.as_str() == "http.response.status_code");
```
utpilla: There is an unconditional pre-scan for `has_http_status_code`. Every span pays an extra full pass over attributes. We could fold the detection into the existing well-known loop (set a flag when `*well_known_key == "httpStatusCode"`), then decide whether to emit `statusMessage` after the loop.

opentelemetry-user-events-trace/src/trace/exporter.rs
utpilla: One thing worth keeping in mind: alpha lowers the cost of changing behavior, but not really the cost of re-reviewing the same code later. The context is loaded right now (serialization technique, common schema etc.), but when the fixes are done later, it would again incur additional review time to build that context.

Not saying that every suggested perf optimization has to land here. If you'd rather keep this PR scoped, a `// TODO(perf)` is fine too.

utpilla (approved):
Left some comments.

- The RPC attribute keys should be updated to use the latest names defined in semantic conventions. 
- Let's create TODOs for perf suggestions if not fixing in this PR.

It looks good to merge with the above two points addressed.

opentelemetry-user-events-trace/src/trace/exporter.rs
cijothomas: Good point, fixing.

opentelemetry-user-events-trace/src/trace/exporter.rs
cijothomas: fixed.

opentelemetry-user-events-trace/src/trace/exporter.rs
cijothomas: done now.

PR #633, opened 2026-06-05
perf(tower): Eliminate allocs from protocol and version

francoposa:
Fixes #

## Changes

~Adds `alloc_tracker` crate and patches into benchmarks to allow us to track improvements there.~
Updated to remove alloc tracker per feedback

Converts `split_and_format_protocol_version` to use `&'static str` rather than `String`.

Alloc difference:
```console
Before:

Allocation statistics:

| Operation           | Mean bytes | Mean count |
|---------------------|------------|------------|
| baseline            |          0 |          0 |
| metrics             |        851 |         26 |
| noop                |        851 |         26 |
| tracing             |       1523 |         27 |
| tracing+metrics     |       1523 |         27 |
| tracing-sampled-out |       1075 |         26 |

---
After:

Allocation statistics:

| Operation           | Mean bytes | Mean count |
|---------------------|------------|------------|
| baseline            |          0 |          0 |
| metrics             |        837 |         22 |
| noop                |        837 |         22 |
| tracing             |       1509 |         23 |
| tracing+metrics     |       1509 |         23 |
| tracing-sampled-out |       1061 |         22 |
```

Criterion output here - percentages reported vs previous are 3-12% improvement (excluding baseline which is unchanged).
[criterion-output.txt](https://github.com/user-attachments/files/28654627/criterion-output.txt)

### DIFF
--- a/opentelemetry-instrumentation-tower/benches/middleware.rs
+++ b/opentelemetry-instrumentation-tower/benches/middleware.rs
@@ -71,6 +71,7 @@
 //! macOS 26.4.1, rustc 1.95.0, OpenTelemetry 0.32.
 //!
 
+use alloc_tracker::{Allocator, Session};
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
 use opentelemetry::global;
 use opentelemetry::trace::noop::NoopTracerProvider;
@@ -83,6 +84,9 @@ use std::convert::Infallible;
 use std::hint::black_box;
 use std::time::Duration;
 use tower::{Service, ServiceBuilder, ServiceExt};
+
+#[global_allocator]
+static ALLOCATOR: Allocator<std::alloc::System> = Allocator::system();
 
 /// Minimal handler — returns an empty body to keep baseline noise as low as possible.
 async fn handler(_req: http::Request<String>) -> Result<http::Response<String>, Infallible> {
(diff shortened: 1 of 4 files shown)


francoposa:
@jan-xyz @cijothomas

jan-xyz (approved):

opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -832,7 +832,7 @@ fn finalize_request<ResBody, E, ResExt>(
     );
 }
 
-fn split_and_format_protocol_version(http_version: http::Version) -> (String, String) {
+fn split_and_format_protocol_version(http_version: http::Version) -> (&'static str, &'static str) {
```
cijothomas, line 835: nice!

opentelemetry-instrumentation-tower/benches/middleware.rs
```diff
@@ -84,6 +85,9 @@ use std::hint::black_box;
 use std::time::Duration;
 use tower::{Service, ServiceBuilder, ServiceExt};
 
+#[global_allocator]
```
cijothomas: I know this is just benchmarks, but not quite sure of replacing global allocator. Let's keep this as-before - we won't be able to measure allocations, but we'd still see latency improvement due to less allocation.

We need to find a way to see allocations, but I prefer to handle that separately, and also across the OTel rust repo - every benchmark in main repo also would benefit from that.

cijothomas (changes requested):
The elimination of alloc is good, thanks!

I'd like to avoid depending on another crate in this crate for the allocation measurements - it is okay to do it locally and share results. For repo, prefer to have something that we can use across entire benches in this/main repo. (I think there is an open issue tracking it.)

Lets remove the allocation tracker for now, and get this merged.

francoposa:
I can remove the alloc tracking here but:

> I know this is just benchmarks, but not quite sure of replacing global allocator

> I'd like to avoid depending on another crate in this crate for the allocation measurements

This is the only existing approach in Rust that I can find. I also could not find the tracking issue in the main or contrib Rust repos.

francoposa:
Updated to remove alloc tracker

cijothomas (approved):
Thanks. Will follow up how best to add allocation tracker OTel Rust wide.

PR #640, opened 2026-06-10
feat(resource-detectors): add host.name to HostResourceDetector

### LINKED ISSUE #639: Add host.name detection to HostResourceDetector
[The spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/host.md) recommends collecting the `host.name` attribute for a host resource. [Go](https://github.com/open-telemetry/opentelemetry-go/blob/35e8690c4bf0d060d6df097e2fa7d316d73373f6/sdk/resource/builtin.go#L61-L64), [Python](https://github.com/open-telemetry/opentelemetry-python/blob/39ff55d3093ec7f218dad2709645824ace971e2d/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L451-L463), and [JS](https://github.com/open-telemetry/opentelemetry-js/blob/71d195c508320295f1892aaed1ee2f1971ffb470/packages/opentelemetry-resources/src/detectors/platform/node/HostDetector.ts#L21-L31) all collect it. I propose adding the same to Rust HostResourceDetector.

mwear:
Fixes #639

## Changes

This PR updates the `HostResourceDetector` to collect the spec recommended `host.name` attribute. This matches implementations in [Go](https://github.com/open-telemetry/opentelemetry-go/blob/35e8690c4bf0d060d6df097e2fa7d316d73373f6/sdk/resource/builtin.go#L61-L64), [Python](https://github.com/open-telemetry/opentelemetry-python/blob/39ff55d3093ec7f218dad2709645824ace971e2d/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L451-L463), and [JS](https://github.com/open-telemetry/opentelemetry-js/blob/71d195c508320295f1892aaed1ee2f1971ffb470/packages/opentelemetry-resources/src/detectors/platform/node/HostDetector.ts#L21-L31)

Since there is not a standard library alternative, I used the [hostname crate](https://crates.io/crates/hostname) as its the most straightforward way to access the hostname on windows/linux/mac. If we're reluctant to take on the dependency we can discuss other options.

* ~~[ ] Changes in public API reviewed (if applicable)~~

### DIFF
--- a/opentelemetry-resource-detectors/Cargo.toml
+++ b/opentelemetry-resource-detectors/Cargo.toml
@@ -16,6 +16,7 @@ rust-version = "1.75.0"
 opentelemetry = { workspace = true }
 opentelemetry_sdk = { workspace = true }
 opentelemetry-semantic-conventions = { workspace = true }
+hostname = "0.4"
 
 [dev-dependencies]
 temp-env = "0.3.6"
--- a/opentelemetry-resource-detectors/Cargo.toml
+++ b/opentelemetry-resource-detectors/Cargo.toml
@@ -16,7 +16,6 @@ rust-version = "1.75.0"
 opentelemetry = { workspace = true }
 opentelemetry_sdk = { workspace = true }
 opentelemetry-semantic-conventions = { workspace = true }
-hostname = "0.4"
 
 [dev-dependencies]
 temp-env = "0.3.6"
--- a/opentelemetry-resource-detectors/src/host.rs
+++ b/opentelemetry-resource-detectors/src/host.rs
@@ -18,8 +18,10 @@ use std::process::Command;
 ///
 /// - [`host.id from non-containerized systems`](https://opentelemetry.io/docs/specs/semconv/resource/host/#collecting-hostid-from-non-containerized-systems)
 /// - Host architecture (host.arch).
+/// - Host name (host.name).
 pub struct HostResourceDetector {
     host_id_detect: fn() -> Option<String>,
+    host_name_detect: fn() -> Option<String>,
 }
 
 impl ResourceDetector for HostResourceDetector {
@@ -39,6 +41,13 @@ impl ResourceDetector for HostResourceDetector {
                         opentelemetry_semantic_conventions::attribute::HOST_ARCH,
                         ARCH,
                     )),
+                    // Get host.name
+                    (self.host_name_detect)().map(|host_name| {
+                        KeyValue::new(
+                            opentelemetry_semantic_conventions::attribute::HOST_NAME,
+                            host_name,
+                        )
+                    }),
                 ]
                 .into_iter()
                 .flatten(),
@@ -81,9 +90,20 @@ fn host_id_detect() -> Option<String> {
     None
 }
 
+fn host_name_detect() -> Option<String> {
+    hostname::get()
+        .ok()?
+        .into_string()
+        .ok()
+        .filter(|name| !name.is_empty())
+}
+
 impl Default for HostResourceDetector {
     fn default() -> Self {
-        Self { host_id_detect }
+        Self {
+            host_id_detect,
+            host_name_detect,
+        }
     }
 }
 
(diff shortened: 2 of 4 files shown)


opentelemetry-resource-detectors/Cargo.toml
```diff
@@ -16,6 +16,7 @@ rust-version = "1.75.0"
 opentelemetry = { workspace = true }
 opentelemetry_sdk = { workspace = true }
 opentelemetry-semantic-conventions = { workspace = true }
+hostname = "0.4"
```
cijothomas: can you check if we can avoid the dependency?

mwear: We can avoid the dependency, but I'll mention that the hostname crate is small and focused, and its implementation is better than the alternative. It calls the system hostname functions directly via FFI (`gethostname` on *nix, `GetComputerNameExW` on Windows).

If we drop it, the cleanest approach is shelling out to the `hostname` command, which works on *nix/Windows, but it spawns a subprocess and depends on the command being available.

I can go either way on this. Let me know what you prefer.

cijothomas: This crate currently avoids external dependencies. I can understand that the crate is small, and well maintained (maintainer djc is well known in Rust world and was one of early OTel Rust maintainers too!), but adding dependency to crates maintained in personal repo is a supply chain risk -If we can avoid it, and inline the code, that works better.

mwear: That makes sense. Are we ok shelling out to `hostname` for now. We can eventually migrate to this: https://doc.rust-lang.org/stable/std/net/fn.hostname.html once it's stable.

mwear: While we're deciding, I'll list out all the options I am aware of:

- **Option 1:** This is probably still a no, but we could use the `hostname` crate and migrate to [std::net::hostname](https://doc.rust-lang.org/stable/std/net/fn.hostname.html) when it's stable. Both of these get the hostname by using the underlying sys calls.
- **Option 2:** Shell out to `hostname`. This will work on all three platforms.
- **Option 3:** Platform specific implementations
  - For linux, we can read `/proc/sys/kernel/hostname`
  - For mac, shell out to `hostname`
  - For windows, read `COMPUTERNAME` env var
- **Option 4:** Implement our own FFI to make the underlying sys calls. This would be unsafe code that we would have to own and maintain. We would basically be writing our own version of the `hostname` crate.

cijothomas: Lets do option2 - we already do that in this crate, so natural to stick with that pattern for now.

mwear: @cijothomas, I went with option 2. Do you mind taking a look when you get a chance?

opentelemetry-resource-detectors/src/host.rs
```diff
@@ -107,14 +124,19 @@ mod tests {
             .get(&Key::from_static_str(
                 opentelemetry_semantic_conventions::attribute::HOST_ARCH
             ))
+            .is_some());
+        assert!(resource
```
lalitb: Could we skip checking `HOST_NAME` in this default detector test? Since hostname detection is best-effort, this test could fail just because the machine running it does not have a usable `hostname` command. The injected `host_name_detect` tests below already cover the hostname behavior without depending on the local machine.

lalitb:
Can we  update the [README.md](https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-resource-detectors) for `host.name`?

opentelemetry-resource-detectors/src/host.rs
mwear: I'll make make that change. In addition, since we assert resource.len() above, I'll have to make it more permissive and pass if it's 2 or 3.

mwear:
I updated the tests and the readme. Codecov is the failing check. I'm not sure what if anything we can do about it.

lalitb (approved):

PR #646, opened 2026-06-12
feat: [Geneva uploader] add optional rustls TLS backend

thperapp:
## Summary

Adds a `tls-rustls` Cargo feature to `geneva-uploader` (and forwards it through `geneva-uploader-ffi` and `opentelemetry-exporter-geneva`) that swaps `native-tls` for `rustls` + `p12-keystore`. The default feature set (`tls-native`) preserves existing behavior bit-for-bit.

The new backend honors any `rustls::crypto::CryptoProvider` installed at process startup, which lets consumers drop their OpenSSL runtime dependency entirely and use FIPS-validated cryptography without rebuilding the uploader.

## Changes

- **`geneva-uploader`** — new additive `tls-native` (default) and `tls-rustls` feature flags. When `tls-rustls` is selected, the config-service client builds its `reqwest::Client` with the rustls TLS backend and parses the PKCS#12 keystore via `p12-keystore`, installing the resulting certs/key directly into the rustls `ClientConfig`. If both flags are enabled (e.g. `--all-features`), `tls-rustls` takes precedence at runtime.
- **`azure_core` / `azure_identity` defaults disabled** — the 0.29 `azure_core` default feature set includes `reqwest_native_tls`, which is additive and would force `native-tls` onto the shared `reqwest` dependency even in a `tls-rustls`-only build. `reqwest_native_tls` is now re-enabled only when `tls-native` is on, so `--no-default-features --features tls-rustls` is genuinely `native-tls`-free.
- **`geneva-uploader-ffi`** and **`opentelemetry-exporter-geneva`** — expose matching `tls-native` (default) / `tls-rustls` features that forward to `geneva-uploader`.
- CHANGELOG entries and a README/inline-doc note describing the new feature and the `CryptoProvider` install requirement when using `tls-rustls`.

## Testing

- `cargo check -p geneva-uploader` (default `tls-native`) — passes.
- `cargo check -p geneva-uploader --no-default-features --features tls-rustls` — passes; resolves rustls/p12-keystore and no longer pulls `native-tls` into the build graph.

## Compatibility

- Default builds are unchanged; consumers who do nothing keep the existing `native-tls`/OpenSSL backend.
- Opting into `tls-rustls` requires consumers to install a `rustls::crypto::CryptoProvider` (e.g. `rustls::crypto::aws_lc_rs::default_provider().install_default()` or `rustls-symcrypt`) before constructing a `GenevaClient`; this is documented in the README and CHANGELOG.

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/Cargo.toml
+++ b/opentelemetry-exporter-geneva/geneva-uploader/Cargo.toml
@@ -32,9 +32,13 @@ url = "2.2"
 md-5 = "0.11.0"
 hex = "0.4"
 lz4_flex = { version = "0.11", features = ["safe-encode"], default-features = false }
-# Azure Identity dependencies - using public crates.io versions
-azure_identity = "0.29"
-azure_core = "0.29"
+# Azure Identity dependencies - using public crates.io versions.
+# Default features are disabled so the consumer-selected TLS backend (see
+# `tls-native` / `tls-rustls` below) chooses whether to pull in `native-tls`
+# through azure_core. Otherwise azure_core 0.29's defaults include
+# `reqwest_native_tls`, which links native-tls unconditionally.
+azure_identity = { version = "0.29", default-features = false }
+azure_core = { version = "0.29", default-features = false, features = ["reqwest", "reqwest_gzip", "reqwest_deflate"] }
 tracing = "0.1"
 # Pinned to >=0.3.47 to pick up the RUSTSEC-2026-0009 fix; pulled in transitively
 # via azure_identity / rcgen / yasna. Not used directly by this crate.
@@ -47,7 +51,12 @@ mock_auth = [] # Disabled by default. Not to be enabled in the prod release.
 # install a `rustls::crypto::CryptoProvider` (e.g. SymCrypt) at process start to satisfy
 # FIPS / OpenSSL-free deployments. The features are additive: enabling both (e.g. via
 # `--all-features`) compiles successfully and `tls-rustls` is used at runtime.
-tls-native = ["dep:native-tls", "reqwest/native-tls", "reqwest/native-tls-alpn"]
+tls-native = [
+    "dep:native-tls",
+    "reqwest/native-tls",
+    "reqwest/native-tls-alpn",
+    "azure_core/reqwest_native_tls",
+]
 tls-rustls = [
     "dep:rustls",
     "dep:rustls-pemfile",
--- a/opentelemetry-exporter-geneva/geneva-uploader/Cargo.toml
+++ b/opentelemetry-exporter-geneva/geneva-uploader/Cargo.toml
@@ -47,10 +47,11 @@ time = ">=0.3.47"
 [features]
 mock_auth = [] # Disabled by default. Not to be enabled in the prod release.
 # TLS backends. At least one must be enabled. `tls-native` is the default and matches the
-# pre-existing behavior. `tls-rustls` swaps to a pure-Rust TLS stack, which lets consumers
-# install a `rustls::crypto::CryptoProvider` (e.g. SymCrypt) at process start to satisfy
-# FIPS / OpenSSL-free deployments. The features are additive: enabling both (e.g. via
-# `--all-features`) compiles successfully and `tls-rustls` is used at runtime.
+# pre-existing behavior. `tls-rustls` swaps to a pure-Rust TLS stack without compiling in a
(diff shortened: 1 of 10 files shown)


opentelemetry-exporter-geneva/geneva-uploader/Cargo.toml
```diff
@@ -16,27 +16,55 @@ base64 = "0.22"
...
+    "reqwest/native-tls",
+    "reqwest/native-tls-alpn",
+    "azure_core/reqwest_native_tls",
+]
+tls-rustls = [
+    "dep:rustls",
+    "dep:rustls-pemfile",
+    "dep:rustls-native-certs",
+    "dep:p12-keystore",
+    "reqwest/rustls-tls-native-roots",
```
lalitb: This still pulls in reqwest’s built-in ring provider path, so the `rustls` build is not truly provider-neutral. `otel-arrow` handles this with reqwest `no-provider` TLS features plus explicit `crypto-*` provider selection at startup; can we align with that pattern so FIPS/SymCrypt users do not get ring compiled in as an implicit fallback?

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -1064,3 +1101,78 @@ fn configure_tls_connector(
...
+            roots
+                .add(cert)
+                .map_err(|e| GenevaConfigClientError::Certificate(e.to_string()))?;
+        }
+    }
+
+    // Pin to TLS 1.2 (matches the native-tls path). The CryptoProvider is selected from the
+    // process-wide default if one was installed (e.g. rustls-symcrypt::default_symcrypt_provider),
+    // otherwise from the provider compiled into rustls via reqwest's `rustls-tls-native-roots`.
+    let config = rustls::ClientConfig::builder_with_protocol_versions(&[&rustls::version::TLS12])
```
lalitb: I believe this can silently use a compiled-in fallback provider if the app did not install one first. For the FIPS/SymCrypt use case, it would be clearer to require an installed `CryptoProvider` and return a clear error if none is installed, similar to `otel-arrow`.

thperapp: done

opentelemetry-exporter-geneva/geneva-uploader/Cargo.toml
thperapp: done

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -1064,3 +1101,102 @@ fn configure_tls_connector(
...
+            roots
+                .add(cert)
+                .map_err(|e| GenevaConfigClientError::Certificate(e.to_string()))?;
+        }
+    }
+
+    // Require an explicitly installed CryptoProvider (e.g. rustls-symcrypt for FIPS).
+    // No built-in fallback provider is compiled in (`reqwest/rustls-tls-native-roots-no-provider`),
+    // so callers must install one at process startup via `CryptoProvider::install_default()`.
+    let provider = rustls::crypto::CryptoProvider::get_default().cloned().ok_or_else(|| {
```
lalitb: nit -   This now requires a provider to be installed, which is the right behavior for the no-provider rustls build. Can we also update the cert-auth test setup in `config_service/mod.rs` to install a test provider when running with `--all-features`?  Otherwise, existing tests that construct `GenevaConfigClient` through the `rustls` path can fail here before they reach the TLS handshake.

lalitb: ok, it actually failed [here](https://github.com/open-telemetry/opentelemetry-rust-contrib/actions/runs/27588617696/job/81567595895?pr=646).

lalitb (approved):
LGTM once failing CI is fixed as mentioned [here](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/646#discussion_r3417796931). Thanks for contributing.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
thperapp: are we okay with using ring as the provider for CI tests? using Symcrypt for the tests would require editing the CI pipeline to make symcrypt libraries available for linking

PR #662, opened 2026-06-19
test: Simplify user_events integrations tests setup for logs and traces by using `one_collect` and `tracepoint_decode`

utpilla:
## Changes
- Simplify user_events integration tests setup by using `one_collect` and `tracepoint_decode` (No subprocesses / temp files / sudo shell-outs)

### DIFF
--- a/opentelemetry-user-events-logs/src/lib.rs
+++ b/opentelemetry-user-events-logs/src/lib.rs
@@ -138,12 +138,15 @@ pub use logs::ProcessorBuilder;
 #[cfg(feature = "experimental_eventname_callback")]
 pub use logs::EventNameCallback;
 
-#[cfg(test)]
+#[cfg(all(test, target_os = "linux"))]
 mod tests {
 
     #[cfg(feature = "experimental_eventname_callback")]
     use crate::EventNameCallback;
     use crate::Processor;
+    use one_collect::perf_event::{RingBufBuilder, RingBufSessionBuilder};
+    use one_collect::tracefs::TraceFS;
+    use one_collect::Writable;
     use opentelemetry::trace::Tracer;
     use opentelemetry::trace::{TraceContextExt, TracerProvider};
     use opentelemetry::{Key, KeyValue};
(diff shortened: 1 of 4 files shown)


opentelemetry-user-events-logs/src/lib.rs
```diff
@@ -895,89 +899,125 @@ mod tests {
...
-        let chmod_status = Command::new("sudo")
-            .args(["chmod", "uog+r", "./perf.data"])
-            .status()?;
-
-        if !chmod_status.success() {
-            panic!("chmod failed with exit code: {:?}", chmod_status.code());
-        }
+            let payload = &full[flags_offset..];
+
+            if let Ok(mut enumerator) = ctx.enumerate_with_name_and_data(
```
lalitb, line 972: Could we avoid silently ignoring decode failures here? Capturing the decode error and returning it from the helper would make these tests more reliable to diagnose.

opentelemetry-user-events-logs/src/lib.rs
```diff
@@ -895,89 +899,125 @@ mod tests {
     }
 
     fn check_user_events_available() -> Result<String, String> {
```
lalitb, line 901: Note to self: these could be shared across logs and traces.

lalitb (approved):

opentelemetry-user-events-logs/src/lib.rs
```diff
@@ -895,89 +899,125 @@ mod tests {
...
+    /// `perf-decode` pipeline produced, so the test assertions are unchanged.
+    ///
+    /// `event` is the full tracepoint spec, e.g. "user_events:myprovider_L2K1".
+    /// Capture runs for `duration_secs`; the caller emits events on another thread
+    /// during this window, mirroring the old `perf record` duration.
+    ///
+    /// This replaces the previous `perf record` + `perf-decode` external-tool
+    /// pipeline with a self-contained, in-process consumer (no external tools, no
+    /// temp files, no `sudo` shell-outs).
+    fn run_perf_and_decode(duration_secs: u64, event: &str) -> std::io::Result<String> {
```
cijothomas, line 933: could be renamed as we no longer use `perf`
Also, the CI for this repo that run these tests were installing perf-decode. Could you cleanup that too in this PR itself as that is no longed needed and we can confirm it works without it.

PR #613, opened 2026-05-12
ci: add weaver live-check workflow for tower instrumentation (seed for #530)

cijothomas:
Seeds CI integration of [Weaver live-check](https://github.com/open-telemetry/weaver/blob/main/crates/weaver_live_check/README.md) for the tower instrumentation, per #530.

Runs a purpose-built `live-check-app` example (under `opentelemetry-instrumentation-tower/examples/live-check-app/`) against an OTLP-mode weaver listener and fails the job on any semantic-convention `violation`. The example exercises a comprehensive HTTP method × status-code × route-shape matrix so weaver validates as many attribute combinations as possible. JSON report uploaded as a workflow artifact.

Uses the reusable `weaver-live-check-start` / `weaver-live-check-stop` composite actions from open-telemetry/weaver, pinned to the `v0.24.0` release. Semconv registry pinned to `v1.36.0` (the version the `opentelemetry-semantic-conventions 0.32` crate is based on).

Intentionally minimal — tower only, default advisors. Easy for follow-ups to expand (actix, finding filters, more endpoints).

Findings worth tracking as follow-ups (not blocking this PR):
- `error.type` is missing on 5xx/4xx server spans — semconv calls for it on error responses.
- Two metrics (`http.server.response.body.size`, `http.server.active_requests`) are flagged `not stable; stability = development` — informational only.

Issue #530, opened 2026-01-23
User weaver live check in CI to ensure Instrumentations are compliant


cijothomas:
Instrumentation libraries (actix, tower in this repo) should leverage weaver live check in CI.

https://github.com/open-telemetry/weaver/blob/main/crates/weaver_live_check/README.md

Don't know if such checks are done in other OTel repos yet. This would be an awesome validation and step towards stability in the libraries. (stability in terms of telemetry stability!)

PR #669, opened 2026-06-24
ci: add weaver live-check workflow for actix-web instrumentation (seed for #530)

cijothomas:
Follow-up to #613. Extends the weaver live-check CI pattern from tower to the actix-web instrumentation, per #530.

Runs a purpose-built `live-check-app` example (under `opentelemetry-instrumentation-actix-web/examples/live-check-app/`) against an OTLP-mode weaver listener and fails the job on any semantic-convention `violation`. The example exercises the same HTTP method × status-code × route-shape matrix the tower example does so weaver validates as many attribute combinations as possible. JSON report uploaded as a workflow artifact.

Uses the reusable `weaver-live-check-start` / `weaver-live-check-stop` composite actions from open-telemetry/weaver, pinned to the `v0.24.0` release. Semconv registry pinned to `v1.36.0` (the version the `opentelemetry-semantic-conventions 0.32` crate is based on). Same workflow shape and pins as `weaver-live-check-tower.yml`.

Differences from the tower example, all of them properties of the instrumentation we're testing rather than the workflow:

- Spans are exported via a simple processor (synchronous per-span) instead of a batch processor. With batching under actix-rt, the exporter background task did not flush by the time the job tore the process down, so weaver saw zero entities. Simple-exporter is fine for an example whose purpose is "every request must produce a span weaver sees".
- `actix` middleware does not currently record a span or metric when a handler panics (the future panics before the response wrapper runs), so the panicking `/throw` route from the tower example is intentionally omitted. The `/error` route already exercises 5xx attribute coverage. Adding panic-recording to the actix middleware is a worthwhile follow-up, out of scope here.
- Server uses `.disable_signals()` and we own SIGTERM/SIGINT so the shutdown sequence flushes telemetry deterministically.

CI result on this branch: **0 violations**, 134 information advisories + 3 improvement advisories. The workflow passes with `fail-on: violation`, same policy as tower.

Findings worth tracking as follow-ups (not blocking this PR):

- `error.type` is set to raw status-code strings (`"400"`, `"401"`, `"404"`, `"500"`) on 4xx/5xx spans. The semconv `error.type` attribute is an enum and weaver flags these as `undefined_enum_variant`. Should be set to a canonical enum value or `_OTHER`.
- `error.type` is missing entirely on a number of error spans/data-points where semconv calls for it (31 occurrences).
- `network.protocol.name` is missing on spans/metric data-points (39 occurrences).
- `http.response.status_code` is missing on some metric data-points (13 occurrences).
- `user_agent.synthetic.type` opt-in attribute not present (39 occurrences) — informational, opt-in is fine to skip.
- Same three "not stable; stability = development" notes the tower run produces (`http.server.active_requests`, `http.server.request.body.size`, `http.server.response.body.size`) — informational only, also present on tower.
- `meter_provider.shutdown()` from inside an `actix_web::main` runtime times out (`InternalFailure("[Timeout(5s)]")`). Data still lands because the PeriodicReader exports during the workflow's pre-kill grace period. Worth a separate investigation into the actix-rt × OTel SDK interaction.

Now that two instrumentation libraries use the same workflow shape, the next natural step is to factor out the shared scaffolding (action invocations, traffic-driver helper, artifact upload) so adding the third instrumentation library doesn't mean copy-pasting another ~140-line YAML file. Tracking separately.

### DIFF
--- a/.github/workflows/weaver-live-check-actix.yml
+++ b/.github/workflows/weaver-live-check-actix.yml
@@ -118,6 +118,11 @@ jobs:
           c "http://127.0.0.1:5000/error"
           echo "::endgroup::"
 
+          # Give the periodic metric reader one final export interval to flush
+          # before we tear down. (Spans use a simple processor and are already
+          # flushed synchronously per request.)
+          sleep 3
+
           kill "$APP_PID" && wait "$APP_PID" || true
 
       - name: Stop weaver live-check
--- a/.github/workflows/weaver-live-check-actix.yml
+++ b/.github/workflows/weaver-live-check-actix.yml
@@ -1,0 +1,135 @@
+name: Weaver Live Check (actix)
+
+# Validates that telemetry emitted by the actix-web instrumentation's
+# live-check-app example complies with the published OpenTelemetry
+# semantic conventions. Follows the same pattern as
+# weaver-live-check-tower.yml; both are part of issue #530.
+
+env:
+  CI: true
+  WEAVER_VERSION: v0.24.0
+  # Pinned semantic-conventions release for reproducible runs. Bump
+  # deliberately when conventions change.
+  SEMCONV_REGISTRY: 'https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.36.0.tar.gz[model]'
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'opentelemetry-instrumentation-actix-web/**'
+      - '.github/workflows/weaver-live-check-actix.yml'
+  merge_group:
+  push:
+    branches:
+      - main
+    paths:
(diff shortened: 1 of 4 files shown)


TommyCpp (approved):
>  the next natural step is to factor out the shared scaffolding (action invocations, traffic-driver helper, artifact upload) so adding the third instrumentation library doesn't mean copy-pasting another ~140-line YAML file

I feel this is a must if we want to maintain a reasonable size of CI configs

cijothomas:
> > the next natural step is to factor out the shared scaffolding (action invocations, traffic-driver helper, artifact upload) so adding the third instrumentation library doesn't mean copy-pasting another ~140-line YAML file
> 
> I feel this is a must if we want to maintain a reasonable size of CI configs

Agree. I am also trying to see if we can improve upstream Weaver itself to make the boilerplate even lesser.

PR #671, opened 2026-06-25
perf(tower): use static-str KeyValue for common HTTP method/scheme

cijothomas:
Use `&'static str` for the `http.request.method` and `url.scheme` `KeyValue`s when the value is one of the 9 standard HTTP methods or `http`/`https`. Custom methods and other schemes fall back to allocating as before.

Each of these `KeyValue`s is cloned 2–3 times per request (span attributes, the unconditional `http.server.active_requests` `UpDownCounter::add()` call, and the metrics label set). With `OtelString::Static`, those clones are just pointer copies instead of allocating a new `Box<str>` each time.

Measured on Apple M4 Pro, OTel 0.32, criterion median of 30 samples:

| Scenario | Before | After | Delta |
|---|---|---|---|
| noop | 705 ns | 569 ns | −19% |
| tracing | 825 ns | 689 ns | −16% |
| tracing-sampled-out | 726 ns | 590 ns | −19% |
| metrics | 1024 ns | 890 ns | −13% |
| tracing + metrics | 1243 ns | 1013 ns | −18% |

First of a planned two-part split; a follow-up will reduce `Vec<KeyValue>` allocations and clones inside `finalize_request`.

### DIFF
--- a/opentelemetry-instrumentation-tower/benches/middleware.rs
+++ b/opentelemetry-instrumentation-tower/benches/middleware.rs
@@ -27,8 +27,13 @@
 //!
 //! ### Middleware-internal (optimizable by this crate)
 //!
-//! - **Heap allocations per request**: `scheme`, `method`, `path`, `URI`,
-//!   `protocol`, `version` each `to_string()` / `to_owned()` into a new `String`.
+//! - **Heap allocations per request**: `path`, `URI` and (for non-common
+//!   methods or schemes) the `method`/`scheme` strings still `to_string()`
+//!   / `to_owned()` into a new `String`. The `method` and `scheme` labels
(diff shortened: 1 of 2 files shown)


cijothomas:
@francoposa @jan-xyz Please review. 
I also realized that you are not getting automatically notified, as you are not members of OpenTelmetry organization. Can you follow https://github.com/open-telemetry/community/issues/new?template=membership.md and request to be added as a member?

Myself, and @TommyCpp can help with sponsorship.

francoposa (approved):
Looks great 👍

jan-xyz (approved):
🚀

lalitb (approved):

PR #675, opened 2026-06-27
fix(opentelemetry-aws): allow deprecated semconv constants from 0.32.1

cijothomas:
`opentelemetry-semantic-conventions 0.32.1` deprecates `RPC_SYSTEM`, `RPC_MESSAGE_TYPE`, `PEER_SERVICE`, and `RPC_SERVICE`. The workspace pins `"0.32"`, so the patch is picked up automatically and breaks `clippy -D warnings` on `opentelemetry-aws` (currently failing CI on every PR).

Adds `#[allow(deprecated)]` per usage, matching the existing pattern for `HTTP_STATUS_CODE`. Not switching to renamed constants — that would change which attribute names the X-Ray exporter recognizes and is a separate behavior change.


chinmaychahar:
LGTM, thanks for the fix! The doc CI is failing currently, wondering if it's a flaky one. Will take a detailed look later

utpilla (approved):

PR #678, opened 2026-06-30
perf(tower): consume RequestData in finalize_request, pre-size label vec

cijothomas:
Consumes `RequestData` by value in `finalize_request` so per-request `KeyValue`s and the two custom-attr `Vec`s can be moved into the metric label set instead of cloned. `label_superset` is pre-sized with `Vec::with_capacity(...)`, and the error-path label set uses a stack array.

No behavior change — same metric labels, same span attributes, same span status. `test_metrics_labels` covers the exact attribute sets and still passes.

Measured on Apple M4 Pro, OTel 0.32, criterion median of 30 samples, two runs averaged, against current `main` (already includes #671):

| Scenario             | Before  | After  | Delta             |
|----------------------|---------|--------|-------------------|
| noop                 | 586 ns  | 559 ns | −5%               |
| tracing              | 683 ns  | 680 ns | not significant   |
| tracing-sampled-out  | 590 ns  | 581 ns | −2% to −4%        |
| metrics              | 931 ns  | 855 ns | **−7% to −9%**    |
| tracing + metrics    | 1023 ns | 970 ns | **−5% to −8%**    |

Tracing-only is unchanged because the no-op meter never builds the label `Vec` to begin with.

### DIFF
--- a/opentelemetry-instrumentation-tower/benches/middleware.rs
+++ b/opentelemetry-instrumentation-tower/benches/middleware.rs
@@ -34,8 +34,11 @@
 //!   (GET/POST/PUT/DELETE/HEAD/OPTIONS/PATCH/CONNECT/TRACE) and the `http`
 //!   and `https` schemes, so the corresponding `KeyValue::clone()` is
 //!   allocation-free in the hot path.
-//! - **`Vec<KeyValue>` allocations**: `span_attributes` and `label_superset`
-//!   are built and populated on every request.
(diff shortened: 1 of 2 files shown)


jan-xyz (approved):
Nice! 🙌

cijothomas:
@lalitb could you signoff? Can't merge with approval from component owner alone, so need sign off from an approver/maintainer.

lalitb (approved):

PR #679, opened 2026-06-30
feat(tower): add with_tracing/with_metrics to disable signals per layer

### LINKED ISSUE #673: opentelemetry-instrumentation-tower: Allow trace generation to be turned off or generated though tracing
opentelemetry-instrumentation-tower starts recording a root trace now instead of metrics. Myself as well as I'm sure other people that use this with AXUM will already use tracing as the main source for how the traces are generated so doing this means that for every request now I end up with 2 distinct root traces (1 managed by the tracing crate and the other by this crate).

Ideally they would be able to merge but I can also understand if theres no desire to pull the tracing dependencies into this crate, even if via feature-flag.

jan-xyz:
## Summary

- Add `http::server::LayerBuilder::with_tracing(bool)` and `with_metrics(bool)` to enable or disable each signal for a layer (both default to enabled).
- When disabled, the layer uses no-op providers internally so no spans or metrics are emitted. Context propagation is unaffected: incoming trace headers are still extracted and the current context still flows to the inner service.
- The instrumentation scope now carries the OpenTelemetry semantic conventions schema URL (previously used a plain name string).
- Update tests and benchmarks to use provider injection instead of mutating layer internals or global providers.

Closes #673.


opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -553,18 +563,29 @@ impl<RouteExt, ReqExt, ResExt> HTTPLayerBuilder<RouteExt, ReqExt, ResExt> {
...
+    }
+
+    /// Override the meter provider used for metrics collection.
     ///
-    /// In production, the default behavior of using the global meter provider
-    /// (via `opentelemetry::global::meter()`) is recommended.
-    #[cfg(test)]
-    fn with_meter(mut self, meter: Meter) -> Self {
-        self.meter = Some(meter);
+    /// By default the layer uses the global meter provider. Passing a no-op
```
cijothomas: it is okay for an instrumentation library to accept an optional provider, and if not provided, it can fallback to the global.
But this should not be documented as a way to disable tracing (or metrics) - we can expose simpler boolean builders like `with_tracing_enabled(true/false)` for that. Asking user to construct a NoOp, and then passing it as the provider feels too much just to disable a signal.

jan-xyz: How about just a `with_disable_tracing`/`with_disable_metrics` which then set it to no-op tracer/meter? I don't think we need an `with_enable_*` option since that's just the default, wdyt?

francoposa: What about `with_noop_trac(er|ing)`/`with_noop_met(er|rics)`?
I think it describes both what's actually being done (applying a no-op instrument) while also being clear that it's disabling it.

cijothomas: `noop` is revealing the internal details to endusers.
`with_disable` style is discouraged to avoid double negation. 

How about with_tracing(bool) / with_metrics(bool), defaulting to true? Short, symmetric, and hides the no-op detail from users.

cijothomas: @jan-xyz if the primary goal can be met with the bool option, lets defer from support accepting custom providers until there is an actual need to keep public API slim.

Also - it's worth noting the Context propagation behavior when disabled. We should be pretty clear that disabling tracing also disabled propagation (i.e we won't passthrough the headers at all, so users should know they have to take remediation themselves to avoid broken traces).

cijothomas: I think the spec expects us to do the propagation irrespective of whether noop or real tracer being used. propagation should continue to work, so we may need another switch to turn that off so users can take matters fully into their control.

jan-xyz: That means that we have to create hard control flows to skip metrics or trace collection and tagging depending on the config, and that setting it to no-op is not enough, right?

cijothomas: @jan-xyz no. We can use no-op internally. My comment about noop was about *not* exposing it to end users.
ie. if with_tracing(false) is used, then we'll use noop tracers, and rest all continue as-is. The net effect is tracing won't be producing any real spans. Context-propagation works (as expected) - and if we need to turn that off, then that'd require another flag.

jan-xyz: Understood, I'll adjust that in the coming days.

jan-xyz:
This is not stale. @cijothomas do you think we can move this forward?

cijothomas:
> This is not stale. @cijothomas do you think we can move this forward?

Sorry, Let me re-review today and update.

opentelemetry-instrumentation-tower/src/lib.rs
```diff
@@ -553,18 +584,41 @@ impl<RouteExt, ReqExt, ResExt> HTTPLayerBuilder<RouteExt, ReqExt, ResExt> {
...
     ///
-    /// This method exists primarily for testing purposes, allowing tests to inject
-    /// a custom meter (e.g., backed by an in-memory exporter) without relying on
-    /// global state. Using global providers in tests can cause interference between
-    /// concurrent tests.
+    /// Tracing is enabled by default. When disabled, the layer records no
+    /// spans, but context propagation is unaffected: incoming trace headers
+    /// are still extracted and the current context still flows to the inner
+    /// service.
+    pub fn with_tracing(mut self, enabled: bool) -> Self {
```
cijothomas: lets add some targetted tests for this
1. It should assert no telemetry is collected.
2. That context propagation still works - this is probably more involved, and good for another PR to keep this focused.

cijothomas (changes requested):
PR is good. Requeiting changes to address couple of things:

1. Add test coverage for new code path: https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/679/changes#r3832749100
2. Update PR desc to match the current approach taken.

jan-xyz:
@cijothomas the PR was adjusted!

cijothomas (approved):

PR #680, opened 2026-06-30
feat(tower): add gRPC instrumentation layer

jan-xyz:
## Summary

Add a Tower gRPC server instrumentation layer with tracing and metrics.

This follows the existing Tower HTTP layer shape where it makes sense:

- `GRPCLayer` / `GRPCService` / `GRPCLayerBuilder` mirror the HTTP layer structure.
- Request and response custom attribute extractors are supported, like the HTTP layer.
- Provider-based configuration is supported through `with_tracer_provider` and `with_meter_provider`.
- The implementation avoids boxed response futures and uses concrete `pin-project-lite` futures, matching the current HTTP-layer performance direction.
- Request data is consumed by value during finalization, vectors are pre-sized, and small error label sets use stack arrays.

The layer records the current RPC semantic convention attributes/metric:

- `rpc.system.name = "grpc"`
- `rpc.method`
- `rpc.response.status_code`
- `error.type` for failed RPCs
- `rpc.server.call.duration`

## Design notes / caveats

- This is server-side instrumentation only.
- The layer derives `rpc.method` from gRPC request paths shaped like `/{service}/{method}` and uses that same value as the span name.
- There is intentionally no custom method extractor. Unlike HTTP paths, gRPC method cardinality is normally bounded by the proto service definition, so the route-extractor/cardinality problem from the HTTP layer does not apply in the same way.
- gRPC status is observed from `grpc-status` response headers and, for streaming responses, from trailers.
- Streaming/trailer support is implemented by wrapping the response body. This means the service response type changes to `Response<GRPCResponseBody<B>>`, but tonic accepts this because the wrapper preserves the inner body `Data` and `Error` types.
- Finalization happens when trailers are observed, when the body ends, on body error, or as a drop fallback.
- Duration therefore measures the response stream lifecycle, not just time-to-response-headers.
- The implementation currently records `rpc.server.call.duration` only. It does not attempt to mirror HTTP-only metrics like active requests or request/response body sizes.
- Error span status follows the RPC semantic-convention guidance for library-generated gRPC errors. Application-level gRPC status codes such as `NOT_FOUND` are not automatically treated as span errors.

## Tests / examples

- Unit tests cover path parsing, tracing attributes, error status, metrics labels, and trailers overriding header status.
- Unit fixtures are HTTP/2-shaped gRPC requests/responses (`POST`, `HTTP_2`, `content-type: application/grpc`, `te: trailers`) while still avoiding a full network stack.
- A real tonic integration test uses `opentelemetry-proto`'s generated OTLP `TraceService` client/server and applies `GRPCLayer` through `tonic::transport::Server::builder().layer(...)`.
- The tonic integration test verifies that the layer composes with tonic and records the expected span name and RPC semantic attributes from a real generated gRPC request.
- There is a small `examples/grpc.rs` showing layer usage with a Tower service.

Open question: I added the tonic integration test to validate compatibility with a real tonic server/client. Now that the lower-level unit tests are finalized, we may decide the integration test is not worth the additional dev dependency on `opentelemetry-proto`, `tonic`, and `tokio-stream`. I am happy to remove it if maintainers prefer to keep this crate's dev dependency surface smaller.

## Benchmarks

The existing `middleware` Criterion benchmark now includes a gRPC group parallel to the HTTP group.

Quick local run:

```sh
cargo bench -p opentelemetry-instrumentation-tower --bench middleware -- --quick
```

Approximate medians from that run:

| Scenario | HTTP total | gRPC total | HTTP overhead vs baseline | gRPC overhead vs baseline |
|---|---:|---:|---:|---:|
| baseline | 69 ns | 344 ns | - | - |
| no-op | 744 ns | 869 ns | +675 ns | +525 ns |
| tracing | 932 ns | 1.10 us | +863 ns | +752 ns |
| tracing sampled-out | 798 ns | 922 ns | +729 ns | +577 ns |
| metrics | 1.26 us | 982 ns | +1.19 us | +638 ns |
| tracing + metrics | 1.46 us | 1.25 us | +1.39 us | +907 ns |

Notes:

- Raw HTTP and gRPC totals are not apples-to-apples. The gRPC benchmark consumes the response body to trigger trailer/body finalization, so its baseline is higher.
- Overhead-over-baseline is a better comparison for the middleware cost.
- The gRPC metrics path currently does less work than the HTTP layer because it records only `rpc.server.call.duration`; the HTTP layer records additional HTTP metrics and labels.
- These are quick local numbers, intended as a regression sanity check rather than stable reference numbers.

## Testing

- `cargo fmt --package opentelemetry-instrumentation-tower`
- `cargo test -p opentelemetry-instrumentation-tower`
- `cargo check -p opentelemetry-instrumentation-tower --examples`
- `cargo check -p opentelemetry-instrumentation-tower --benches`
- `cargo bench -p opentelemetry-instrumentation-tower --bench middleware -- --quick`

### DIFF
--- a/opentelemetry-instrumentation-tower/benches/middleware.rs
+++ b/opentelemetry-instrumentation-tower/benches/middleware.rs
@@ -77,8 +77,11 @@
 //!
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
+use http_body_util::{BodyExt, Empty};
 use opentelemetry::global;
+use opentelemetry::metrics::noop::NoopMeterProvider;
 use opentelemetry::trace::noop::NoopTracerProvider;
+use opentelemetry_instrumentation_tower::grpc::GRPCLayerBuilder;
 use opentelemetry_instrumentation_tower::HTTPLayerBuilder;
 use opentelemetry_sdk::{
     metrics::{InMemoryMetricExporter, PeriodicReader, SdkMeterProvider},
@@ -94,6 +97,19 @@ async fn handler(_req: http::Request<String>) -> Result<http::Response<String>, 
     Ok(http::Response::new(String::new()))
 }
 
+/// Minimal gRPC handler — returns an empty HTTP/2 gRPC response.
+async fn grpc_handler(
+    _req: http::Request<Empty<&'static [u8]>>,
+) -> Result<http::Response<Empty<&'static [u8]>>, Infallible> {
+    Ok(http::Response::builder()
+        .status(http::StatusCode::OK)
+        .version(http::Version::HTTP_2)
+        .header("content-type", "application/grpc")
+        .header("grpc-status", "0")
+        .body(Empty::new())
+        .unwrap())
+}
+
 /// Build requests outside the timed loop so request construction cost does not
 /// inflate the middleware overhead measurement.
 fn build_requests(n: u64) -> Vec<http::Request<String>> {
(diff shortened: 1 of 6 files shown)


opentelemetry-instrumentation-tower/benches/middleware.rs
```diff
@@ -132,6 +165,10 @@ fn noop_tracer() {
     global::set_tracer_provider(NoopTracerProvider::new());
 }
 
+fn noop_meter() {
+    global::set_meter_provider(NoopMeterProvider::new());
+}
```
jan-xyz, line 170: This is required because #679 isn't merged yet, which makes it possible that the benchmarks don't interfere

PR #696, opened 2026-07-01
feat(aws): add MissingSampledBehavior config to XrayPropagator

### LINKED ISSUE #522: [Feature]: Allow configuring XrayPropagator behavior when header lacks sampling decision
### Related Problems?

_No response_

### What component are you working with?

opentelemetry-aws

### Describe the solution you'd like:

Currently, `XrayPropagator` sets the flag `TRACE_FLAG_DEFERRED` if no `Sampled` attribute is present in the `x-amzn-trace-id` header: https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/53788cd57b144393632634c9fd504006dbace36f/opentelemetry-aws/src/trace/xray_propagator.rs#L108

If `XRayPropagator` is used by a service directly behind an AWS Elastic Load Balancer, then the `Sampled` attribute will not be present (unless the incoming HTTP request already had it set): https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html. As a result, extracting and setting the parent span context (e.g. using crates like [auxm-tracing-opentelemetry](https://docs.rs/axum-tracing-opentelemetry/latest/axum_tracing_opentelemetry/middleware/struct.OtelAxumLayer.html)) will always result in the root span having sampling disabled.

It would be useful to be able to configure the behavior of `XrayPropagator` when the `Sampled` attribute is not present - e.g. always setting the sampler flag, or delegating to a provided [Sampler](https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/trace/enum.Sampler.html)

### Considered Alternatives

A `TextMapCompositePropagator` can be used to configure a custom propagator to run after `XrayPropagator`, which can adjust the sampled flag on the `Context`. However, it would be useful to be able to do this directly in `XrayPropagator`, considering that the default behavior can be incorrect when using an ELB.

### Additional Context

_No response_

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

chinmaychahar:
Fixes #522

## Changes

Adds `MissingSampledBehavior` enum and `XrayPropagator::with_missing_sampled_behavior()` to configure what happens when the `Sampled` field is absent from the `x-amzn-trace-id` header. This fixes the issue where services behind AWS ELB lose all traces because the propagator defaulted to `DEFERRED` (which downstream SDKs treat as "not sampled").

Usage:
```rust
let propagator = XrayPropagator::new()
    .with_missing_sampled_behavior(MissingSampledBehavior::Sampled);

### DIFF
--- a/opentelemetry-aws/src/trace/xray_propagator.rs
+++ b/opentelemetry-aws/src/trace/xray_propagator.rs
@@ -86,9 +86,19 @@ fn trace_context_header_fields() -> &'static [String; 1] {
 /// [otel-spec]: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#SpanContext
 /// [xray-trace-id]: https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids
 /// [xray-header]: https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader
+
+/// Configures how `XrayPropagator` handles a missing `Sampled` field in the trace header.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub enum MissingSampledBehavior {
+    #[default]
+    Deferred,
+    Sampled,
+    NotSampled,
+}
+
 #[derive(Clone, Debug, Default)]
 pub struct XrayPropagator {
-    _private: (),
+    missing_sampled_behavior: MissingSampledBehavior,
 }
 
 /// Extract `SpanContext` from AWS X-Ray format string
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: opentelemetry-aws/src/trace/xray_propagator.rs [hunk 1]
```diff
@@ -204,13 +165,78 @@ impl XrayPropagator {
         XrayPropagator::default()
     }
 
+    /// Sets the behavior when the `Sampled` field is missing from the trace header.
+    pub fn with_missing_sampled_behavior(mut self, behavior: MissingSampledBehavior) -> Self {
+        self.missing_sampled_behavior = behavior;
+        self
+    }
+
+    fn default_sampling_decision(&self) -> TraceFlags {
```
TommyCpp: the naming between getter and setter should generally be aligned

chinmaychahar: Right, thanks for pointing out. I'll update

chinmaychahar:
CI cargo-deny failure is pre-existing. I have #701 out to fix it (it ignores quick-xml advisories from a dev-only transitive dep, couldn't find an upstream fix yet)

chinmaychahar:
@TommyCpp could you take another look at this? Thanks a lot!

chinmaychahar:
hey hey @TommyCpp , gentle reminder ^

opentelemetry-aws/src/trace/xray_propagator.rs
utpilla: I don't follow the original suggestion being here. I think using `with_` prefix for the missing sampled behavior method is better.

`default_sampling_decision` is an internal method. Also, it doesn't seem like a getter.

@TommyCpp could you clarify what you meant [here](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/696#discussion_r3510552544)?

opentelemetry-aws/src/trace/xray_propagator.rs
```diff
@@ -204,13 +164,78 @@ impl XrayPropagator {
         XrayPropagator::default()
     }
 
+    /// Sets the behavior when the `Sampled` field is missing from the trace header.
+    pub fn missing_sampled_behavior(mut self, behavior: MissingSampledBehavior) -> Self {
```
utpilla:
```suggestion
    pub fn with_missing_sampled_behavior(mut self, behavior: MissingSampledBehavior) -> Self {
```

utpilla (approved):
Looks good! I have left a comment about using `with_` prefix for the `missing_sampled_behavior` method. I'm not sure why that change was reverted.

opentelemetry-aws/src/trace/xray_propagator.rs
TommyCpp: Sorry for the ambiguity. I was commenting on `default_sampling_decision`, not on the setter.

Two things I had in mind:

1. **The same concept has two names.** The field and setter call it *missing sampled behavior*, while the accessor calls it *default sampling decision*. Readers have to map one onto the other.
2. **`default` collides with the derived `Default`** (which is `Deferred`). Once a user configures `Sampled`, a method named `default_*` returns the *non*-default value.

So my suggestion is:

```suggestion
    fn missing_sampling_decision(&self) -> TraceFlags {
```

and keep the setter as `with_missing_sampled_behavior` — agreed with @utpilla that `with_` is the right prefix for a consuming builder method, and that this method isn't really a getter. Sorry for the detour, @chinmaychahar.

chinmaychahar:
I restored the with_ prefix on the setter and renamed the internal default_sampling_decision as missing_sampling_decision. Thanks both!

utpilla (approved):

PR #698, opened 2026-07-04
 feat(opentelemetry-c): add Rust-backed C trace API and SDK

lalitb:
## Summary

  This PR adds an experimental `opentelemetry-c` component: a Rust-backed C API/SDK for OpenTelemetry traces.

  The goal is to expose the Rust OpenTelemetry implementation through a C ABI. C applications, C/C++ instrumentation libraries, and other runtimes/languages can consume this ABI without binding directly to Rust or C++.

This is not intended to replace the native C++ SDK. C++ can be one downstream consumer through a thin wrapper if needed.

  ## Why Rust Contrib

This is proposed in `opentelemetry-rust-contrib` because the implementation is Rust-backed.

The C layer is only the ABI surface. The implementation depends on Rust SDK types, feature flags, exporters/processors, lifecycle behavior, and future Rust SDK evolution, so ownership should stay close to the Rust ecosystem.

  ## What Is Included

  - `opentelemetry-c/abi`
    - Internal Rust-only ABI crate.
    - Shared `#[repr(C)]` types and internal vtable.
    - Not exposed through public C headers.

  - `opentelemetry-c/api`
    - C API artifact for tracer provider, tracer, and span handles.
    - Owns the process-global provider slot.
    - Supports API-only no-op behavior when no SDK is installed.
    - Intended for instrumentation libraries to link API-only.

  - `opentelemetry-c/sdk`
    - C SDK artifact backed by the Rust SDK.
    - Supports OTLP HTTP/protobuf trace export and batch span processing.
    - Registers into the API-owned global provider slot.
    - Keeps OTLP/reqwest/TLS optional behind features.

  ## Usage Model

  - Instrumentation libraries link only `libopentelemetry_c_api`.
  - Applications link both `libopentelemetry_c_api` and `libopentelemetry_c_sdk`.
  - The application configures the SDK/exporter/processor and installs it globally.
  - API-only instrumentation then dispatches through the installed SDK.

  ## Scope

This initial PR is trace-only and alpha/experimental.

  Included:

  - Spans, attributes, events, and status
  - Explicit parent span support
  - OTLP HTTP/protobuf exporter
  - Batch span processor
  - Force flush / shutdown
  - C example and header compile tests

  Deferred:

  - Metrics and logs
  - Propagation, span links, samplers, span limits, ID generator
  - Simple span processor
  - Custom/user-provided exporter
  - Array-valued attributes

  ## Important Limitations

  - ABI is alpha and not stable yet.
  - Shared global provider model is supported for Unix-like dynamic linking with one loaded `libopentelemetry_c_api`.
  - Windows dynamic split is documented as unsupported/unverified for now.
  - Static linking multiple copies of the API can create separate global slots.
  - `otel_sdk_shutdown` / `otel_sdk_destroy` do not clear the API global provider slot; SDK library lifetime is documented.

  ## Validation

  The PR includes tests/coverage for:

  - C header compilation
  - API-only no-op behavior
  - SDK-backed export path
  - Cross-artifact dynamic linking
  - Provider retain/free race safety
  - Backed failures returning `NULL`
  - Parent/child span semantics
  - No-default SDK core without OTLP/reqwest
  - Batch processor/exporter ownership transfer


opentelemetry-c/sdk/README.md
```diff
@@ -0,0 +1,163 @@
+# opentelemetry-c-sdk
```
cijothomas, line 1: we need to have an example that shows how this can be consumed from a C application.

opentelemetry-c/abi/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+## vNext
+
+### Added
+
+- Initial release of `opentelemetry-c-abi` as part of the split of `opentelemetry-c` into
```
cijothomas: we don't publish right? so changelog itself is not required.

opentelemetry-c/sdk/README.md
lalitb: Yes there is an example in example directory. But yes, i got your point - more full fledged example for a standalone application.

lalitb: Added a minimal README section with two snippets: one showing how an instrumentation library uses the API-only path, and one showing how an application configures the SDK/exporter/processor and installs it globally. The full buildable version with error handling remains linked from `examples/c-basic-traces`.

opentelemetry-c/abi/CHANGELOG.md
lalitb: good point. Removed.

lalitb:
Making it ready for review - with reasoning added in PR desc on why this belongs to otel-rust-contrib repo.

PR #699, opened 2026-07-06
chore: Migrate geneva-uploader off unmaintained rustls-pemfile

thperapp:
rustls-pemfile is unmaintained (RUSTSEC-2025-0134). Replace it with the PEM parsing now exposed directly by rustls-pki-types (>=1.9) via the PemObject trait, using CertificateDer::pem_slice_iter in the tls-rustls code path. This removes the direct rustls-pemfile dependency for downstream consumers such as otap-dataflow.


lalitb (approved):
Thanks.

PR #700, opened 2026-07-07
feat(tower): instrument outgoing HTTP requests with a client layer

jan-xyz:
Fixes #
Design discussion issue (if applicable) #713

## Changes

- HTTP client instrumentation layer (`http::client::Layer`) producing a
  `SpanKind::Client` span and the standard `http.client.*` metrics, and
  injecting the current trace context into outgoing request headers so the
  callee can continue the trace.
- `http-server` and `http-client` Cargo features (both enabled by default) so
  a server-only or client-only deployment does not compile instrumentation it
  will never call.
- The client layer defaults to method-only span names because an outbound URL
  carries no route template to normalize against; using the raw path would push
  request-specific values into span names and metric labels.
- Tracing and metrics are individually toggleable per layer via
  `with_tracing(bool)` and `with_metrics(bool)` (both enabled by default).
- `hyper-http-client` and `reqwest-http-client` examples demonstrating the
  client layer with Hyper and reqwest clients via `tower-reqwest`, with OTLP
  export.
- Test verifying that the default `NoRouteExtractor` produces method-only
  span names and omits the `http.route` attribute, preventing cardinality
  explosion from dynamic path segments.
- Client benchmarks aligned with #755 — switched from `iter_custom` to
  `iter_batched` with `BatchSize::SmallInput` and added guidance for
  `--sample-size` and `taskset` to reduce measurement noise.


lalitb:
@francoposa  - wondering if you would be the right person to review this PR ?

francoposa:
Could we do a clean refactor PR first to apply the namespacing to the existing code, then add the client code and all the benchmarking as a separate PR after?

I think the refactor may have a bit of discussion but be relatively quick to get through, while the client code could take a bit longer.

jan-xyz:
@francoposa good idea, I created a separate PR (#717) for moving the server to the new structure, let's discuss that one first, and once it's merged we can come back to this one.

francoposa:
Previous comment deleted, a rogue PR review agent decided to ignore all its restrictions

francoposa:
I would like there to be a client code example added as part of this - it could even supplant the k6 load-gen script in the docker compose setup.

IMO think it's especially important as the "Tower Service as client middleware" is not as obvious or well-trodden of a path in other library documentation.

jan-xyz:
not stale, just waiting for review.

opentelemetry-instrumentation-tower/examples/reqwest-http-client/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    let otel_layer = LayerBuilder::builder().build().unwrap();
+
+    let mut client = ServiceBuilder::new()
+        .layer(otel_layer)
+        .service(service_fn(move |req: http::Request<reqwest::Body>| {
+            let client = reqwest_client.clone();
+            async move {
+                let req: reqwest::Request = req.try_into()?;
+                let res: reqwest::Response = client.execute(req).await?;
```
jan-xyz: This is a bit ugly, but I couldn't find a better way to convert the `reqwest::{Request, Response}` to the corresponding `http` structs in a different way. I'm wondering if there is maybe a good way to adapt the layer itself?

francoposa: Does something like this help? https://github.com/alekseysidorov/tower-http-client

jan-xyz: For this I had to remove the `Send` requirement, which is a justifiable trade-off based on my research. Tower _client_ layers usually don't have this, and clients get `.await`-ed and not async spawned.

jan-xyz:
Not stale

PR #703, opened 2026-07-09
feat(geneva-uploader): route logs and spans by table name

ragumarimuthu-git:
This PR updates Geneva uploader routing so logs and spans use explicit table names from GenevaClientConfig via LogMethod and SpanMethod.

If default_event_name is not set, logs fall back to Log and spans fall back to Span. The change is applied through the uploader client, encoder, examples, FFI layer, stress harness, and integration tests. I also added unit coverage for the fallback and override behavior.

Validation:
- cargo build --workspace
- cargo test -p geneva-uploader --lib
- cargo test --workspace --all-targets

Note: this change also updates Geneva OBO lookup behavior to follow the fixed routing table name.

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
@@ -11,7 +11,7 @@ use std::sync::OnceLock;
 use tokio::runtime::Runtime;
 
 use geneva_uploader::client::{EncodedBatch, GenevaClient, GenevaClientConfig, UploadError};
-use geneva_uploader::{AuthMethod, LogMethod, SpanMethod};
+use geneva_uploader::{AuthMethod, LogsConfig, TracesConfig};
 use opentelemetry_proto::tonic::collector::trace::v1::ExportTraceServiceRequest;
 use prost::Message;
 use std::path::PathBuf;
@@ -596,10 +596,10 @@ pub unsafe extern "C" fn geneva_client_new(
         role_name,
         role_instance,
         msi_resource,
-        logs: LogMethod {
+        logs: LogsConfig {
             default_event_name: None,
         },
-        spans: SpanMethod {
+        spans: TracesConfig {
             default_event_name: None,
         },
         obo_event_map,
@@ -1645,10 +1645,10 @@ mod tests {
             role_name: "testrole".to_string(),
             role_instance: "testinstance".to_string(),
             msi_resource: None,
-            logs: LogMethod {
+            logs: LogsConfig {
                 default_event_name: None,
             },
-            spans: SpanMethod {
+            spans: TracesConfig {
                 default_event_name: None,
             },
             obo_event_map: None,
@@ -2177,10 +2177,10 @@ mod tests {
             role_name: "testrole".to_string(),
             role_instance: "testinstance".to_string(),
             msi_resource: None,
-            logs: LogMethod {
+            logs: LogsConfig {
                 default_event_name: None,
             },
-            spans: SpanMethod {
+            spans: TracesConfig {
                 default_event_name: None,
             },
             obo_event_map: None,
@@ -2306,10 +2306,10 @@ mod tests {
             role_name: "testrole".to_string(),
             role_instance: "testinstance".to_string(),
             msi_resource: None,
-            logs: LogMethod {
+            logs: LogsConfig {
                 default_event_name: None,
             },
-            spans: SpanMethod {
+            spans: TracesConfig {
                 default_event_name: None,
             },
             obo_event_map: None,
@@ -2473,10 +2473,10 @@ mod tests {
             role_name: "testrole".to_string(),
             role_instance: "testinstance".to_string(),
             msi_resource: None,
-            logs: LogMethod {
+            logs: LogsConfig {
                 default_event_name: None,
             },
-            spans: SpanMethod {
+            spans: TracesConfig {
                 default_event_name: None,
             },
             obo_event_map: None,
@@ -2600,10 +2600,10 @@ mod tests {
             role_name: "testrole".to_string(),
             role_instance: "testinstance".to_string(),
             msi_resource: None,
-            logs: LogMethod {
+            logs: LogsConfig {
                 default_event_name: None,
             },
-            spans: SpanMethod {
+            spans: TracesConfig {
                 default_event_name: None,
             },
             obo_event_map: None,
@@ -2773,10 +2773,10 @@ mod tests {
             role_name: "testrole".to_string(),
             role_instance: "testinstance".to_string(),
             msi_resource: None,
-            logs: LogMethod {
+            logs: LogsConfig {
                 default_event_name: None,
             },
-            spans: SpanMethod {
+            spans: TracesConfig {
                 default_event_name: None,
             },
             obo_event_map: None,
--- a/opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
@@ -11,7 +11,7 @@ use std::sync::OnceLock;
 use tokio::runtime::Runtime;
 
 use geneva_uploader::client::{EncodedBatch, GenevaClient, GenevaClientConfig, UploadError};
-use geneva_uploader::AuthMethod;
+use geneva_uploader::{AuthMethod, LogMethod, SpanMethod};
 use opentelemetry_proto::tonic::collector::trace::v1::ExportTraceServiceRequest;
 use prost::Message;
 use std::path::PathBuf;
@@ -596,6 +596,12 @@ pub unsafe extern "C" fn geneva_client_new(
         role_name,
         role_instance,
         msi_resource,
+        logs: LogMethod {
+            default_event_name: None,
+        },
+        spans: SpanMethod {
+            default_event_name: None,
+        },
         obo_event_map,
     };
 
@@ -1639,6 +1645,12 @@ mod tests {
             role_name: "testrole".to_string(),
             role_instance: "testinstance".to_string(),
             msi_resource: None,
+            logs: LogMethod {
+                default_event_name: None,
+            },
+            spans: SpanMethod {
+                default_event_name: None,
+            },
             obo_event_map: None,
         };
         let client = GenevaClient::new(cfg).expect("failed to create GenevaClient with MockAuth");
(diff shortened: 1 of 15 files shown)


opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
```diff
@@ -596,6 +596,12 @@ pub unsafe extern "C" fn geneva_client_new(
         role_name,
         role_instance,
         msi_resource,
+        logs: LogMethod {
+            default_event_name: None,
+        },
+        spans: SpanMethod {
+            default_event_name: None,
+        },
```
lalitb, line 604: Could we expose the log and span table names through GenevaConfig instead of hard-coding `None` here, so FFI consumers can pass the table names?

ragumarimuthu-git: By default, the table name is set as Log and Span, but this option provides the ability to override the hardcoded table name.

lalitb: Yes, the defaults make sense. My concern is that FFI consumers cannot use the override because `geneva_client_new` always sets both values to None. Could we add the log and span table-name fields to the C `GenevaConfig` struct and pass them through here? A breaking FFI change is fine at this stage.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -47,9 +47,21 @@ pub struct GenevaClientConfig {
     pub role_name: String,
     pub role_instance: String,
     pub msi_resource: Option<String>, // Required for Managed Identity variants
+    pub logs: LogMethod,
```
lalitb: Why do we need `LogMethod` and `SpanMethod` wrappers - can we directly use the `Option<String>` - unless the intent is to have these wrapper type support more settings later? Otherwise, they add an unnecessary level of indirection.

ragumarimuthu-git: Yes, this method will get extended based on this requirement: https://github.com/open-telemetry/otel-arrow/issues/3375. So I just created a shell in this PR but any extension related to Log/span will be added to this method

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -95,11 +107,25 @@ pub struct GenevaClient {
...
+    let Some(default_event_name) = default_event_name else {
+        return fallback;
+    };
+
+    default_event_name
+}
+
 impl GenevaClient {
     pub fn new(cfg: GenevaClientConfig) -> Result<Self, String> {
+        let client_config = cfg.clone();
```
lalitb: Could we avoid storing the full `GenevaClientConfig` just to retain the table names? `GenevaClient` is cloned for each encoded batch, so this change makes every clone copy the authentication configuration and another copy of the OBO map. Can we store only the resolved log and span table names instead?

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -3484,8 +3486,8 @@ mod tests {
         let batches = result.unwrap();
         assert_eq!(
             batches.len(),
-            2,
-            "Should have separate batches for different event names"
+            1,
+            "Routing table is fixed; events share one batch"
```
lalitb, line 3506: Does this test still exercise OBO? The encoder uses the fixed Log table, but the OBO map is keyed by `OboEvent`, so the OBO configuration will not be applied. Could we update or replace this test to verify the new table-based OBO behavior?

ragumarimuthu-git: Batching should be done with batch processor and not here in the uploader as the payload size will not be utilized to the maximum if it is done in the uploader.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -215,12 +216,11 @@ impl<'a> LogRecordParts<'a> {
...
         role: Cow<'a, str>,
         role_instance: Cow<'a, str>,
+        table_name: &'a str,
     ) -> Self {
         let event_name = normalized_event_name(record);
         let name = event_name.map(Cow::Borrowed);
-        let routing_event_name = event_name
-            .map(Cow::Borrowed)
-            .unwrap_or_else(|| Cow::Borrowed(CS_LOG_TYPENAME));
+        let routing_event_name = Cow::Borrowed(table_name);
```
lalitb, line 223: Since fixed-table routing can combine more schema shapes into one batch, could we add a benchmark covering high schema diversity to ensure schema lookup still scales well? Unless the plan is to eventually remove this grouping from the uploader and rely on an upstream otel-arrow processor to partition the records.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
lalitb: Agree, but my comment above is OBO test coverage, not batching.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
ragumarimuthu-git: Yes, plan is to remove grouping from uploader and rely on otel-arrow batch processor to partition the records.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -47,9 +47,21 @@ pub struct GenevaClientConfig {
...
+#[derive(Clone, Debug)]
+pub struct LogMethod {
+    pub default_event_name: Option<String>,
+}
+
+#[derive(Clone, Debug)]
+pub struct SpanMethod {
+    pub default_event_name: Option<String>,
+}
+
```
utpilla: Please rename these to `LogsConfig` and `TracesConfig`. That name is a better fit with what the issue describes: https://github.com/open-telemetry/otel-arrow/issues/3375 

```suggestion
#[derive(Clone, Debug)]
pub struct LogsConfig {
    pub default_event_name: Option<String>,
}

#[derive(Clone, Debug)]
pub struct TracesConfig {
    pub default_event_name: Option<String>,
}
```

utpilla: When you add the mapping logic in the follow-up PR, I expect the structs to look like ths:

```rust
#[derive(Clone, Debug, Default)]
pub struct LogsConfig {
    pub default_event_name: Option<String>,          // None => "Log"
    pub event_name_mapping: Option<EventNameMapping>, // None => all to default
}

pub struct EventNameMapping {
    pub routing_key: RoutingKey,
    pub events: HashMap<String, String>, // value == key => passthrough
}

pub enum RoutingKey {
    EventName,
    ResourceAttribute(String),
    ScopeAttribute(String),
    LogRecordAttribute(String),
}
```

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -95,11 +107,29 @@ pub struct GenevaClient {
...
+fn effective_table_name<'a>(
+    default_event_name: Option<&'a str>,
+    fallback: &'static str,
+) -> &'a str {
+    let Some(default_event_name) = default_event_name else {
+        return fallback;
+    };
+
+    default_event_name
+}
```
utpilla: Could we remove this method? This seems unnecessary. A simple `unwrap_or` statement is much more readable and immediately conveys the intent.

```rust
let log_table_name: Arc<str> =
    cfg.logs.default_event_name.as_deref().unwrap_or("Log").into();
```

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
utpilla, line 1: Please add a CHANGELOG entry for the behavioral change.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
```diff
@@ -344,6 +344,12 @@ mod tests {
             }
 
             let request = String::from_utf8_lossy(&request);
+            if request.trim().is_empty() {
+                // Some failure-path tests (for example, untrusted CA) can
+                // establish/tear down the socket without sending an HTTP
+                // request body. Do not fail those tests on an empty request.
+                return;
+            }
```
utpilla, line 352: This looks unrelated to the config changes. Why is this being added in this PR?

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -349,3 +395,212 @@ impl GenevaClient {
             })
     }
 }
+
+#[cfg(test)]
+mod tests {
```
utpilla, line 396: These 14 tests all assert the same thing — that `effective_table_name(Option, fallback)` returns the override or the fallback — routed through different builders (`build_config`, `build_config_from_optional`, `build_config_with_logs_none`, `build_config_with_spans_none`, `resolve_log_table_name`, `resolve_span_table_name`). For example `config_logs_default_event_name_none_uses_log_table_name`, `config_from_optional_logs_missing_uses_log_fallback`, and `config_build_helper_none_and_override_works` are the same assertion via three different helpers. 

They mostly exercise the test scaffolding, not production behavior.

**The gap:** none of them verify the resolved name actually reaches the wire. There's no test asserting that default_event_name: `Some("AppLog")` produces batches whose `event_name == "AppLog"`. That's the behavior a consumer cares about, and it's exactly what the encoder tests already do for the "Log"/"Span" defaults.

Could we simplify things here? We could drop these 14 plus the builder helpers, and cover the behavior end-to-end (2-3 tests):

If you want to keep a unit-level check on the unwrap_or itself, one test on default_name() (per the shape change) is plenty. The rest add maintenance cost without adding coverage.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/mod.rs
ragumarimuthu-git: This is to fix a failing test

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
ragumarimuthu-git: This will get changed in the next PR as this is the part1 of the big change

utpilla (approved):

lalitb (approved):

PR #708, opened 2026-07-14
ci: add cargo-semver-checks workflow

chinmaychahar:
## Changes

Adds a `cargo-semver-checks` CI workflow that runs on every PR to detect accidental semver-breaking API changes before merge. Uses `continue-on-error: true` so it's informational (non-blocking) while the team gets familiar with it.

Follows the same pattern as the upstream [opentelemetry-rust semver workflow](https://github.com/open-telemetry/opentelemetry-rust/blob/main/.github/workflows/semver.yml).

### DIFF
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -1,0 +1,30 @@
+name: Semver Checks
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  semver:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    timeout-minutes: 10
+    steps:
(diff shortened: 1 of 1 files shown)


cijothomas:
most components in this repo are alpga/beta with breaking changes still occurring. do we need this? or its more like a FYI that there is breaking changes across releases?

chinmaychahar:
> or its more like a FYI that there is breaking changes across releases?

yes, I was thinking if it can be a non-blocking workflow. Lemme know if that doesn't work

cijothomas:
> > or its more like a FYI that there is breaking changes across releases?
> 
> yes, I was thinking if it can be a non-blocking workflow. Lemme know if that doesn't work

any way to mark the CI as still green? Its not friendly to contributors to see a red CI that is not actionable for them.

chinmaychahar:
> any way to mark the CI as still green? Its not friendly to contributors to see a red CI that is not actionable for them.

I'll take a look. if not, I'll close this.

PR #709, opened 2026-07-15
feat(geneva-uploader): add agent-fed credential source

manishgoel3:
Fixes #
Design discussion issue (if applicable) # open-telemetry/otel-arrow#3275

## Changes

Adds an agent-fed credential source to `geneva-uploader` so a host agent can supply the GIG token and routing (endpoint/moniker) directly, letting the
uploader skip its GCS config-service handshake.

- `AgentFedCredentialSource` trait - async `current()` returning an
  `AgentFedCredential` snapshot (token + endpoint + moniker). `AgentFedCredential` has a `Debug` impl that redacts the token.
- `GenevaClient::with_agent_fed_source(...)` / `GenevaUploader::from_agent_fed(...)` build path: queries the source per upload (so host token rotation is observed without rebuilding the client) and creates no `GenevaConfigClient`.
- The `endpoint=` query parameter is derived from the token's `Endpoint` claim (same as the GCS path), so it can't drift from the token.

Consumed by the otel-arrow Geneva exporter (separate PR).

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
@@ -168,7 +168,7 @@ struct ClientParts {
 impl GenevaClient {
     /// Derive the shared uploader config + stored parts; constructors differ
     /// only in how they build the `GenevaUploader` from the returned config.
-    fn derive_parts(cfg: &GenevaClientConfig) -> (GenevaUploaderConfig, ClientParts) {
+    fn derive_parts(cfg: GenevaClientConfig) -> (GenevaUploaderConfig, ClientParts) {
         let log_table_name: Arc<str> = cfg
             .logs
             .default_event_name
@@ -190,12 +190,12 @@ impl GenevaClient {
 
         // Metadata fields that will appear as Bond schema fields in Geneva.
         let metadata_fields = MetadataFields::new(
-            cfg.environment.clone(),
+            cfg.environment,
             config_version.clone(),
-            cfg.tenant.clone(),
-            cfg.role_name.clone(),
-            cfg.role_instance.clone(),
-            cfg.namespace.clone(),
+            cfg.tenant,
+            cfg.role_name,
+            cfg.role_instance,
+            cfg.namespace,
             config_version,
         );
 
@@ -212,7 +212,7 @@ impl GenevaClient {
                 log_table_name,
                 span_table_name,
                 metadata_fields,
-                obo_event_map: cfg.obo_event_map.clone(),
+                obo_event_map: cfg.obo_event_map,
             },
         )
     }
@@ -270,17 +270,15 @@ impl GenevaClient {
             AuthMethod::MockAuth => {}
         }
 
-        let (uploader_config, parts) = Self::derive_parts(&cfg);
-
         let config_client_config = GenevaConfigClientConfig {
-            endpoint: cfg.endpoint,
-            environment: cfg.environment,
-            account: cfg.account,
-            namespace: cfg.namespace,
-            region: cfg.region,
+            endpoint: cfg.endpoint.clone(),
+            environment: cfg.environment.clone(),
+            account: cfg.account.clone(),
+            namespace: cfg.namespace.clone(),
+            region: cfg.region.clone(),
             config_major_version: cfg.config_major_version,
-            auth_method: cfg.auth_method,
-            msi_resource: cfg.msi_resource,
+            auth_method: cfg.auth_method.clone(),
+            msi_resource: cfg.msi_resource.clone(),
             #[cfg(test)]
             test_root_ca_pem: None,
         };
@@ -294,6 +292,8 @@ impl GenevaClient {
                 );
                 format!("GenevaConfigClient init failed: {e}")
             })?);
+
+        let (uploader_config, parts) = Self::derive_parts(cfg);
 
         let uploader =
             GenevaUploader::from_config_client(config_client, uploader_config).map_err(|e| {
@@ -332,7 +332,7 @@ impl GenevaClient {
             "Initializing GenevaClient (agent-fed credential source)"
         );
 
-        let (uploader_config, parts) = Self::derive_parts(&cfg);
+        let (uploader_config, parts) = Self::derive_parts(cfg);
 
         let uploader = GenevaUploader::from_agent_fed(source, uploader_config).map_err(|e| {
             debug!(
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
@@ -156,8 +156,19 @@ pub struct GenevaClient {
     obo_event_map: Option<OboEventMap>,
 }
 
+/// Stored client pieces shared by every constructor (all of `Self` except the
+/// uploader and encoder), derived once so the constructors can't drift.
+struct ClientParts {
+    log_table_name: Arc<str>,
+    span_table_name: Arc<str>,
+    metadata_fields: MetadataFields,
+    obo_event_map: Option<OboEventMap>,
+}
+
 impl GenevaClient {
-    pub fn new(cfg: GenevaClientConfig) -> Result<Self, String> {
+    /// Derive the shared uploader config + stored parts; constructors differ
+    /// only in how they build the `GenevaUploader` from the returned config.
+    fn derive_parts(cfg: &GenevaClientConfig) -> (GenevaUploaderConfig, ClientParts) {
         let log_table_name: Arc<str> = cfg
             .logs
             .default_event_name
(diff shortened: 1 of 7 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
```diff
@@ -322,22 +390,24 @@ impl GenevaUploader {
...
-            );
-            Err(GenevaUploaderError::UploadFailed {
-                status: status.as_u16(),
-                retry_after,
-                message: body,
-            })
+            return Ok(ingest_response);
         }
+
+        tracing::warn!(
```
lalitb, line 412: Can we avoid logging the full URL here? At warn level it may expose sourceIdentity, moniker, namespace/event, and OBO params.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -61,6 +63,55 @@ pub struct LogsConfig {
...
+    pub token: String,
+    /// GIG ingestion endpoint (base URL the upload POSTs to).
+    pub endpoint: String,
+    /// Account moniker for the upload.
+    pub moniker: String,
+    /// Monitoring endpoint hint from the host. For agent-fed uploads the
+    /// `endpoint=` query parameter is normally derived from the token's
+    /// `Endpoint` claim; this value is only a fallback when that claim is
+    /// absent.
+    pub monitoring_endpoint: String,
```
lalitb: Can we rename this field or avoid it? In the upload path this is only used as a fallback for the `endpoint=` query param when the token has no `Endpoint` claim. If agent-fed tokens should always carry that claim, maybe we should fail instead. If fallback is needed, a name like `endpoint_query_fallback` would make the contract clearer.

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
```diff
@@ -252,45 +284,81 @@ impl GenevaUploader {
             "Starting upload"
         );
 
-        // Always get fresh auth info
-        let (auth_info, moniker_info, monitoring_endpoint) =
-            self.config_client.get_ingestion_info().await?;
+        let data = Bytes::from(data);
```
lalitb: Do we need `Bytes` here? The body is only sent once, so we can keep the existing `Vec<u8>` flow: take `data.len()` before the request and pass `.body(data)`. That also avoids the new dependency and the `data.clone()`.

lalitb:
Do we need to expose this through `geneva-uploader-ffi` too, or is agent-fed only for the Rust `otel-arrow` consumer for now? The current FFI path still always builds through `geneva_client_new()` / `GenevaClient::new()`, so C/C++ callers would not be able to use this mode without a separate constructor or callback-style credential source.

lalitb (approved):
LGTM overall. I left a couple of comments, but I think we should avoid logging the full URL at warn level before merging.

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
manishgoel3: Donee.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
manishgoel3: Addressed in the latest iteration.

manishgoel3:
> Do we need to expose this through `geneva-uploader-ffi` too, or is agent-fed only for the Rust `otel-arrow` consumer for now? The current FFI path still always builds through `geneva_client_new()` / `GenevaClient::new()`, so C/C++ callers would not be able to use this mode without a separate constructor or callback-style credential source.

Thanks for the review. Yes, these changes are Rust-only for now. I'm waiting for the engine-side changes to land first, then will create the FFI/consumer-side wiring.

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
```diff
@@ -113,10 +115,19 @@ pub(crate) struct GenevaUploaderConfig {
...
 }
 
+/// Where the uploader gets the GIG credential per upload.
+#[derive(Debug, Clone)]
+pub(crate) enum IngestionSource {
+    /// Fetch from the Geneva Config Service (the default cert/MSI path).
+    ConfigClient(Arc<GenevaConfigClient>),
+    /// Use an agent-fed credential; no GCS handshake.
+    AgentFed(Arc<dyn crate::client::AgentFedCredentialSource>),
+}
```
utpilla, line 127: Right now there are two abstractions describing the same axis "where does the credential come from":

- the public `AgentFedCredentialSource` trait, and
- the internal `IngestionSource` enum that `upload()` matches on.

The two `match` arms converge on the same 4-tuple `(token, endpoint, moniker, endpoint_query_param)`, and the `extract_endpoint_from_token` fallback is duplicated inline in the agent-fed arm.

We could simplify things here by folding the `AgentFedCredentialSource` trait + `IngestionSource` enum into one internal trait. We could then express both paths as impls of a single internal trait, so `upload()` never branches.

```rust
pub(crate) struct ResolvedIngestion {
    token: String,
    gig_endpoint: String,
    moniker: String,
    endpoint_query_param: String,
}

pub(crate) trait IngestionCredentialSource: Send + Sync + Debug {
    fn resolve(&self) -> Pin<Box<dyn Future<Output = Result<ResolvedIngestion>> + Send + '_>>;
}

// GenevaConfigClient impls it (wraps get_ingestion_info).
// A small AgentFedIngestionSource adapter impls it (wraps
// Arc<dyn AgentFedCredentialSource> and owns the extract_endpoint_from_token fallback).
```

`upload()` then becomes:

```rust
let ResolvedIngestion { token, gig_endpoint, moniker, endpoint_query_param } =
    self.source.resolve().await?;
```

Why it is a bit cleaner:

- One abstraction, not two. Removes the trait+enum overlap for a single concern.
- Branch-free upload path. The endpoint-query-param derivation lives in exactly one place (the agent-fed adapter) instead of inline in a match arm.
- Extensible. A future source (env var, IMDS, sidecar) is a new impl with zero edits to `upload()`. The `#[ignore]` integration test also stops needing to know the concrete variant.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -240,6 +285,77 @@ impl GenevaClient {
         })
     }
 
+    /// Agent-fed construction: the host supplies the GIG credential via
+    /// `source`, so the uploader skips the GCS config-service handshake entirely.
+    /// `cfg.auth_method` / `cfg.msi_resource` are ignored (no `GenevaConfigClient`
+    /// is created). The source is queried per upload, so host token rotation is
+    /// observed without rebuilding the client.
+    pub fn with_agent_fed_source(
```
utpilla, line 323: `GenevaClient::new` and `with_agent_fed_source` share ~40 lines of identical setup: `log_table_name`/`span_table_name`, `source_identity`, `config_version`, `MetadataFields::new(...)`, the `GenevaUploaderConfig { ... }`, and the final `Self { ... }`. The only real difference is how the `GenevaUploader` is built (config-client vs agent-fed source).

Extracting a private helper that takes `cfg` plus the already-built uploader would collapse both constructors to "build the source, then assemble":

```rust
fn assemble(cfg: GenevaClientConfig, uploader: GenevaUploader) -> Self {
    // build tables + MetadataFields from cfg, then Self { ... }
}

pub fn new(cfg: GenevaClientConfig) -> Result<Self, String> {
    let uploader = /* config-client path */;
    Ok(Self::assemble(cfg, uploader))
}

pub fn with_agent_fed_source(cfg: GenevaClientConfig, source: Arc<dyn AgentFedCredentialSource>) -> Result<Self, String> {
    let uploader = /* agent-fed path */;
    Ok(Self::assemble(cfg, uploader))
}
```

This keeps the metadata/table derivation in one place, so the two entry points can't drift apart.

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
manishgoel3: Added IngestionSource::resolve() so upload() no longer branches. Kept it on the enum instead of a new trait + adapter, since AgentFedCredentialSource stays public anyway; same result, one less abstraction.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
manishgoel3: Done; extracted derive_parts() + assemble(), so both constructors now differ only in how they build the uploader.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -111,8 +156,19 @@ pub struct GenevaClient {
...
+    span_table_name: Arc<str>,
+    metadata_fields: MetadataFields,
+    obo_event_map: Option<OboEventMap>,
+}
+
 impl GenevaClient {
-    pub fn new(cfg: GenevaClientConfig) -> Result<Self, String> {
+    /// Derive the shared uploader config + stored parts; constructors differ
+    /// only in how they build the `GenevaUploader` from the returned config.
+    fn derive_parts(cfg: &GenevaClientConfig) -> (GenevaUploaderConfig, ClientParts) {
```
utpilla: Could we update `derive_parts` to take the metadata fields by value instead of `&cfg`? That way it can move the metadata fields rather than cloning them.

```suggestion
    fn derive_parts(cfg: GenevaClientConfig) -> (GenevaUploaderConfig, ClientParts) {
```

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
```diff
@@ -113,10 +115,71 @@ pub(crate) struct GenevaUploaderConfig {
...
+                    gig_endpoint: auth_info.endpoint,
+                    moniker: moniker_info.name,
+                    endpoint_query_param,
+                })
+            }
+            IngestionSource::AgentFed(source) => {
+                let cred = source.current().await.ok_or_else(|| {
+                    GenevaUploaderError::ConfigClient(
+                        "agent-fed credential not yet provisioned by host".to_string(),
+                    )
```
utpilla: When the host hasn't provisioned a credential, `IngestionSource::resolve()` returns `GenevaUploaderError::ConfigClient(...)`, whose Display is `"Config service error: {0}"`. So, the surfaced message becomes `"Config service error: agent-fed credential not yet provisioned by host"` - contradictory, since the agent-fed path never talks to the config service.

We could make things better by adding a dedicated variant for it:

```rust
#[derive(Debug, Error)]
pub(crate) enum GenevaUploaderError {
    // ...existing variants...
    #[error("Agent-fed credential not provisioned: {0}")]
    CredentialNotProvisioned(String),
}
```
```rust
// in IngestionSource::resolve(), AgentFed arm:
let cred = source.current().await.ok_or_else(|| {
    GenevaUploaderError::CredentialNotProvisioned(
        "host has not provisioned a credential yet".to_string(),
    )
})?;
```

That yields "Agent-fed credential not provisioned: host has not provisioned a credential yet", which is accurate and lets callers distinguish this case from a genuine config-service failure. Trivial change; non-blocking.

utpilla (approved):
Thanks @manishgoel3 for the contribution! Left some minor comments.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
manishgoel3: Done.

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
manishgoel3: Added a CredentialNotProvisioned variant and switched the agent-fed arm to it, so the message is accurate. Also tightened the not-provisioned test to assert the variant.

Kept it on the internal GenevaUploaderError to stay non-breaking (UploadError shipped exhaustive in 0.5.0), so it surfaces via UploadError::Other. Can promote it to a public variant if we are OK with a breaking 0.6.0.

utpilla (approved):

PR #712, opened 2026-07-20
feat(geneva): add attribute-based event/table-name routing for logs and spans

ragumarimuthu-git:
## Summary

Adds **attribute-based event/table-name routing** for logs and spans in the Geneva exporter, exposed in both the Rust API and the C FFI. This lets a single exporter fan records out to multiple destination events/tables based on the value of a chosen attribute (or the log's event name / span attributes), instead of everything landing in one fixed table.

## Motivation

Previously all logs went to a single `Log` table and all spans to a single `Span` table. Users need to route records to different destination tables based on record content (e.g. a `service.name` resource attribute or a per-record attribute) without running multiple exporters.

### Rust API (`geneva-uploader`)
- `LogsConfig` and `TracesConfig` gain an optional `event_name_mapping` field:
  - `LogsEventNameMapping` — routing key is one of: event name, resource attribute, scope attribute, log-record attribute.
  - `SpanEventNameMapping` — routing key is one of: resource attribute, scope attribute, span attribute.
  - Both support the reserved keys `scope.name` / `scope.version` to route by instrumentation scope.
- An `events: HashMap<String, Option<String>>` map defines `source_value → destination`:
  - Non-empty destination → route to that destination event/table.
  - `None` / empty destination → pass the source value through unchanged.
  - Source value absent from the map → fall back to the configured default event name (`Log` / `Span` if unset).
- Records/spans are split into **one encoded batch per resolved destination**.
- Invalid mappings (empty `events`, blank source keys, blank attribute routing-key name) are rejected up front in `GenevaClient::new`.
- Non-string attribute values (`Int64` / `Double` / `Bool`) are stringified before lookup; blank/whitespace strings are treated as absent.

### C FFI (`geneva-uploader-ffi`)
- New `GenevaConfig` fields (appended after `obo_map` to preserve field-append convention) plus `FfiLogsEventNameMapping` / `FfiSpansEventNameMapping` structs and entry arrays in `geneva_ffi.h`.
- A **null** mapping pointer means "routing not configured" (`Ok(None)`); a **non-null** mapping is fully validated (routing kind ↔ key-name consistency, non-null entries with `count > 0`, non-blank names/values).

### Breaking change
- `GenevaClientConfig.logs` and `.spans` changed from `LogsConfig` / `TracesConfig` to `Option<LogsConfig>` / `Option<TracesConfig>`. Pass `None` to use the default `Log` / `Span` table names. Examples, benches, stress harness, and integration tests updated accordingly.

## Implementation notes
- Logs routing lives in `payload_encoder/otlp_encoder.rs` (operates on `otap-df-pdata-views` view traits); span routing lives in `client.rs` (operates on `opentelemetry-proto` tonic types). Lookup semantics are unified via a shared `resolve_mapped_destination` helper and shared `validate_event_mapping`.
- Span grouping uses a `HashMap` index into an insertion-ordered `Vec` (O(n), preserves first-seen order).

## Testing
- `cargo fmt` clean, `cargo clippy` clean, `cargo build --all-targets` OK.
- **98** `geneva-uploader` tests + **36** FFI tests pass (FFI via `--features geneva-uploader-ffi/mock_auth`).
- Coverage (`cargo-llvm-cov`) of changed files: `otlp_encoder.rs` 95.5%, `client.rs` 87.7%, `ffi/lib.rs` 82.1%.
- Added tests cover attribute/scope/event-name routing, passthrough vs default fallback, non-string values, blank-value fallback, and FFI validation (null → none, zero-count rejection, invalid-kind rejection).

## Changelog
- Entries added under `[Unreleased]` in `geneva-uploader/CHANGELOG.md` (feature + breaking `Option<...>` note) and `geneva-uploader-ffi/CHANGELOG.md` (FFI config fields).


ragumarimuthu-git:
### Note on the `codecov/project` check

`codecov/patch` (coverage of the lines changed in this PR) **passes at ~94%**. The failing check is `codecov/project`, and it is failing due to a **fork-PR coverage-session artifact**, not because of uncovered code in this change:

| | Base (`main`) | This PR (head) |
|---|---|---|
| Coverage | 77.87% | 71.1% |
| Codecov upload **sessions** | **6** | **1** |
| Total hits | 20,221 | 19,456 |

The head has **765 fewer total hits than base despite this PR *adding* ~1,900 covered lines**. That is only possible because whole *unrelated* crates (`opentelemetry-user-events-*`, `opentelemetry-etw-*`) flip from "hit" to "missed": their coverage comes from the `--ignored` integration-coverage jobs (`user-events-integration`, `etw-integration`), which are path-gated to those crates and **do not run on this PR**. On this PR only the `coverage` job uploads (1 session), so those crates report as uncovered and drag the project total down ~6.8% — past the 0.5% `auto` threshold.

For comparison, a same-repo (non-fork) PR such as #711 merged with `codecov/project` = success and 6 sessions.

**This cannot be fixed by adding tests to the geneva crates** — the "lost" coverage is in other crates whose integration jobs don't execute here.

Coverage of the code actually changed in this PR is high and was deliberately strengthened (patch ~94%; new unit tests added for logs/spans routing value-type branches, FFI mapping validation for all routing kinds, and batch-grouping). Full suite: **105 geneva-uploader + 43 FFI tests pass**, clippy clean, formatted.

**Requested action** (maintainers): please either (a) merge past the advisory `codecov/project` status, (b) re-run CI from within the repo so all coverage sessions upload for the head SHA, or (c) if desirable, make the project status `informational`/enable flag `carryforward` in `.github/codecov.yaml` so fork PRs aren't penalized for integration sessions they can't run.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -998,24 +1377,31 @@ impl OtlpEncoder {
...
-                        &resource_role,
+                        resource_role: &resource_role,
+                        resource: resource.as_ref(),
+                        scope: scope.as_ref(),
+                        event_name_mapping,
                         obo_event_map,
-                    );
+                    },
+                };
+                for log_record in scope_logs.log_records() {
```
utpilla, line 1045: For `ResourceAttribute` / `ScopeAttribute` (and `scope.name`/`scope.version`) routing, the routing value is constant for every record in a scope, but `resolve_log_routing_event_name` re-scans the resource/scope attributes and allocates a fresh `String` for each record. In a batch of N records that's N redundant scans + allocations.

Could we resolve the resource-level value once per resource and the scope-level value once per scope, then pass it down to each record? Only `EventName` and `LogRecordAttribute` genuinely need per-record resolution. The loop already has the resource/scope in hand at the right level, so it's a localized change.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -419,27 +626,58 @@ impl GenevaClient {
...
-                    "Span compression failed"
-                );
-                format!("Compression failed: {e}")
-            })
+        for resource_span in spans {
+            let resource = resource_span.resource.as_ref();
+            for scope_span in &resource_span.scope_spans {
+                let scope = scope_span.scope.as_ref();
+                for span in &scope_span.spans {
+                    let event_name = resolve_span_event_name(
```
utpilla: Same issue as the logs path: `resolve_span_event_name` re-scans the resource/scope attributes for every span even though the value is constant within a scope. Since this loop is already structured `resource -> scope -> span`, we can compute the resource/scope routing value once at the scope level and reuse it, resolving per span only for `SpanAttribute`.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -152,50 +164,244 @@ pub struct GenevaClient {
...
+            ProtoAnyValue::IntValue(value) => Some(value.to_string()),
+            ProtoAnyValue::DoubleValue(value) => Some(value.to_string()),
+            ProtoAnyValue::BoolValue(value) => Some(value.to_string()),
+            _ => None,
+        };
+    }
+    None
+}
+
+fn span_routing_value_from_scope(
```
utpilla: The span routing helpers here (`span_routing_value_from_attributes` / `_from_scope` / `_from_resource`) duplicate the logic in the encoder's `routing_value_from_attributes` / `routing_value_from_scope` — same attribute-value stringification (String/Int64/Double/Bool) and the same `scope.name`/`scope.version` handling, just over proto types vs view traits. Two copies of the same rules can drift.

Could we extract common logic to a shared helper method?

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
```diff
@@ -342,6 +422,237 @@ unsafe fn convert_obo_event_map(
...
+/// Converts an FFI logs event-name mapping to the Rust-side `LogsEventNameMapping`.
+///
+/// Returns `Ok(None)` only when `ffi_mapping` is null (routing not configured).
+/// A non-null mapping is fully validated: the routing kind and key name must be
+/// consistent, and `entries` must be non-null with `count > 0`.
+///
+/// # Safety
+/// - `ffi_mapping` must be null or point to a valid `FfiLogsEventNameMapping`
+/// - All non-null string pointers inside the mapping must be valid null-terminated C strings
+unsafe fn convert_logs_event_name_mapping(
```
utpilla, line 540: These two converters are ~150 lines each and structurally identical apart from the routing-kind enum and which kinds are valid. A small, shared helper (or macro) for the entries loop + routing-kind/name validation would remove most of the duplication and keep the two in sync.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
ragumarimuthu-git: Done in 8ff6ac88. Log routing is now precomputed once per instrumentation scope via a new `LogScopeRouting { None, Fixed, PerRecord }` enum. `ResourceAttribute`/`ScopeAttribute` (and `scope.name`/`scope.version`) resolve to a `Fixed` event name once per scope and are reused for every record; only `EventName`/`LogRecordAttribute` remain per-record. Added a multi-record test asserting a resource-attribute scope yields a single batch.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
ragumarimuthu-git: Done in 8ff6ac88. `encode_and_compress_spans` now computes a `SpanScopeRouting { Fixed, PerSpan }` once per scope via `resolve_span_scope_routing`. Resource-/scope-level keys (and the no-mapping/table-name case) resolve to a `Fixed` name once and are reused across spans; only `SpanAttribute` resolves per span. `resolve_span_event_name` is retained as a `#[cfg(test)]` delegator over the same helpers. Added a multi-span test guarding the shared resolution.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
ragumarimuthu-git: Done in 8ff6ac88. The primitive-stringification rules are now shared via `RoutingPrimitive` + `stringify_routing_primitive` in `otlp_encoder.rs`, and both the logs (pdata views) and spans (proto `ProtoAnyValue`) paths funnel string/int/double/bool and `scope.name`/`scope.version` values through it, so the two paths can't drift on value handling. Full unification isn't possible since the two operate over different type families (proto vs pdata-view traits), but the value semantics and lookup (`resolve_mapped_destination`) are now shared.

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
ragumarimuthu-git: Done in 8ff6ac88. Extracted the shared logic: a `FfiMapEntry` trait (implemented for both `FfiLogsEventNameMapEntry` and `FfiSpansEventNameMapEntry`) plus `required_routing_key_name` (null-check + non-blank validation) and `convert_mapping_entries` (the identical entries loop). Both converters now just map their kind discriminants to routing keys and delegate the rest, cutting the duplicated ~150-line bodies down substantially. Existing FFI tests (43) still pass and guard behavior.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -152,50 +164,294 @@ pub struct GenevaClient {
...
+}
+
+/// Resolve the routed event name for a single span given its scope's routing.
+fn resolve_span_event_name_in_scope(
+    scope_routing: &SpanScopeRouting<'_>,
+    span: &Span,
+    table_name: &str,
+) -> String {
+    match scope_routing {
+        SpanScopeRouting::Fixed(name) => name.clone(),
```
utpilla: Could we avoid the per-span/per-record allocation? `scope_routing` outlives the inner loop, so the scope-invariant name can be borrowed instead of cloned per span. Right now the common path (and notably the no-mapping default path, where every span resolves to the table name) allocates a String per span and the new-group branch clones it a second time.

We could return `Cow` and only own when a new group is actually created.

```rust
fn resolve_span_event_name_in_scope<'a>(
    scope_routing: &'a SpanScopeRouting<'a>,
    span: &Span,
    table_name: &'a str,
) -> Cow<'a, str> {
    match scope_routing {
        SpanScopeRouting::Fixed(name) => Cow::Borrowed(name.as_str()),
        SpanScopeRouting::PerSpan { mapping, key } => {
            match span_routing_value_from_attributes(&span.attributes, key)
                .and_then(|value| resolve_mapped_destination(&mapping.events, &value))
            {
                Some(name) => Cow::Owned(name),
                None => Cow::Borrowed(table_name),
            }
        }
    }
}
```

and in the grouping loop, look up by borrow and only own on insert:

```rust
match group_index.get(event_name.as_ref()) {
    Some(&idx) => routed_groups[idx].1.push(span),
    None => {
        let event_name = event_name.into_owned();
        group_index.insert(event_name.clone(), routed_groups.len());
        routed_groups.push((event_name, vec![span]));
    }
}
```

The same applies on the logs side in `LogRecordParts::new` make `routing_event_name` an `Option<Cow<'a, str>>` and return `Cow::Borrowed(name.as_str())` for `Fixed`; only `PerRecord` owns.

utpilla (approved):
Left another suggestion for avoiding an allocation. Looks good to merge after that.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -419,27 +684,63 @@ impl GenevaClient {
             "Encoding and compressing resource spans"
         );
 
-        let span_iter = spans
-            .iter()
-            .flat_map(|resource_span| resource_span.scope_spans.iter())
-            .flat_map(|scope_span| scope_span.spans.iter());
+        let mut routed_groups: Vec<(String, Vec<&Span>)> = Vec::new();
```
lalitb, line 584: Is `span_event_name_mapping` expected to become mandatory? If not, I think we should keep a direct fast path for `None`. Routing is optional, so existing users should not need a dummy mapping or pay the grouping cost when every span still goes to the default table. Only the configured routing path needs `routed_groups` and `group_index`.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -89,6 +196,221 @@ fn normalized_event_name(record: &impl LogRecordView) -> Option<&str> {
...
+/// Resolves a routing `source_value` against a mapping's `events` table.
+///
+/// Returns:
+/// - `Some(destination)` when the source value has a non-empty destination,
+/// - `Some(source_value)` when the source value maps to `None`/empty (passthrough),
+/// - `None` when there is no entry for the source value (caller falls back to the
+///   configured default/table name).
+///
+/// Shared by logs and spans routing to keep the lookup semantics identical.
+pub(crate) fn resolve_mapped_destination(
```
lalitb: Could this method return a borrowed value instead of `Option<String>`? Both the mapped destination and string source value already live long enough for the lookup. We should only allocate for numeric formatting or when inserting a new group key, since this runs for every record or span using attribute routing.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -152,50 +165,301 @@ pub struct GenevaClient {
...
+                        format!("resource_attribute({})", attr)
+                    }
+                    LogsEventNameRoutingKey::ScopeAttribute(attr) => {
+                        format!("scope_attribute({})", attr)
+                    }
+                    LogsEventNameRoutingKey::LogRecordAttribute(attr) => {
+                        format!("log_record_attribute({})", attr)
+                    }
+                };
+                let events_desc = mapping
```
lalitb: nit - Do we need to log the full mapping at info level? These values may contain tenant, service, or deployment names, and a large mapping could create a very large log entry. Logging the routing kind and number of entries should be enough for normal startup logs.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -152,50 +165,301 @@ pub struct GenevaClient {
...
+                        format!("resource_attribute({})", attr)
+                    }
+                    SpanEventNameRoutingKey::ScopeAttribute(attr) => {
+                        format!("scope_attribute({})", attr)
+                    }
+                    SpanEventNameRoutingKey::SpanAttribute(attr) => {
+                        format!("span_attribute({})", attr)
+                    }
+                };
+                let events_desc = mapping
```
lalitb: same as [here](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/712/changes#r3642156867)

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -6,10 +6,21 @@ use crate::ingestion_service::uploader::{
...
 };
 use crate::payload_encoder::otlp_encoder::OtlpEncoder;
-use crate::payload_encoder::otlp_encoder::{lookup_obo_config, MetadataFields};
-pub use crate::payload_encoder::otlp_encoder::{OboEventConfig, OboEventMap};
-use opentelemetry_proto::tonic::trace::v1::ResourceSpans;
+use crate::payload_encoder::otlp_encoder::{
+    lookup_obo_config, resolve_mapped_destination, stringify_routing_primitive, MetadataFields,
+    RoutingPrimitive, SCOPE_NAME_ROUTING_KEY, SCOPE_VERSION_ROUTING_KEY,
+};
+pub use crate::payload_encoder::otlp_encoder::{
```
lalitb: Could we move the routing configuration and resolution logic into a separate routing module? These are public client configuration types, but they currently live inside the payload encoder and are re-exported from here. We also now have one routing implementation for log views and another in client.rs for proto spans. A shared routing component would make the ownership clearer  and reduce the chance of the two paths drifting later.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -89,6 +196,221 @@ fn normalized_event_name(record: &impl LogRecordView) -> Option<&str> {
...
+        };
+    }
+    None
+}
+
+fn routing_value_from_scope<SV>(scope: &SV, key: &str) -> Option<String>
+where
+    SV: InstrumentationScopeView,
+{
+    match key {
```
lalitb: Could we represent scope name and version as explicit routing-key variants instead of special attribute names? `ScopeAttribute("scope.name")` currently reads the scope name rather than an attribute, so a real scope attribute with that name cannot be selected. Variants such as `ScopeName` and `ScopeVersion` would make the API clearer and avoid this collision.

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
```diff
@@ -342,6 +422,209 @@ unsafe fn convert_obo_event_map(
...
+        } else {
+            Some(unsafe {
+                c_str_to_string(
+                    entry.destination_event_name(),
+                    &format!("{ctx}.entries[{i}].destination_event_name"),
+                )?
+            })
+        };
+
+        events.insert(source_value, destination_event_name);
```
lalitb: Should we reject duplicate source values here instead of silently keeping the last one? A duplicate is most likely a configuration mistake, and silently overwriting it can route data to an unexpected table. We can use the result of insert or the entry API and return an invalid-config error when the source already exists.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
ragumarimuthu-git: Done - extracted a shared `payload_encoder/routing` module that now owns the public routing config types, the shared routing primitives, and both resolution paths (the log-view resolvers and the proto-span resolvers that previously lived in client.rs). Public API paths are preserved via re-exports from client.rs. Done in fe4856dd.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
ragumarimuthu-git: Good catch - restored a direct None fast path in `encode_and_compress_spans`: when no span mapping is configured all spans are encoded once under the span table name, so existing users pay no per-span routing cost and need no dummy mapping. Done in a21d9890.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
ragumarimuthu-git: Agreed - no longer logging the full mapping at info. We now log only the routing kind and `event_mapping_count` (logs and spans). Done in a21d9890.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
ragumarimuthu-git: Done - `resolve_mapped_destination` now returns `Option<Cow<'a, str>>`. The common lookup hit borrows the destination from the `events` table (zero allocation); we only allocate on passthrough. Threaded `Cow` through the logs resolvers and the span per-span path. Done in c832c29d.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
ragumarimuthu-git: Same fix applied on the spans side - logging routing kind + count instead of the full mapping. Done in a21d9890.

opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs
ragumarimuthu-git: Done - `convert_mapping_entries` now rejects duplicate `source_value` entries (via the `entry` API) instead of silently overwriting, returning an error. Added a unit test covering the duplicate-source case. Done in a21d9890.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
lalitb: Thanks, the direct None path resolves my original concern. One small follow-up: do we need to collect the spans into a `Vec` here? `encode_span_batch` accepts an iterator, so we could pass the flattened iterator directly and avoid allocating one pointer for every span.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/routing.rs
```diff
@@ -0,0 +1,493 @@
...
+
+/// Stringify a routing primitive using the shared routing rules: string values
+/// are treated as absent when blank/whitespace and otherwise passed through;
+/// numeric/bool values are formatted via `to_string`.
+///
+/// Kept in one place so the logs path (over pdata-view traits) and the spans
+/// path (over proto types) can't drift on value handling.
+pub(crate) fn stringify_routing_primitive(value: RoutingPrimitive<'_>) -> Option<String> {
+    match value {
+        RoutingPrimitive::Str(s) => (!s.trim().is_empty()).then(|| s.to_string()),
```
lalitb: Could we avoid allocating for string routing values here? The string already lives long enough for the map lookup, so this could return a borrowed value for `RoutingPrimitive::Str` and allocate only when formatting numeric or bool values. This path runs for every record or span using attribute routing.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
lalitb: To clarify the issue with an example: suppose the instrumentation scope name is "my-library" and the scope also has an attribute scope.name = "custom-route". With `ScopeAttribute("scope.name")`, the current code returns "my-library" and never searches the attributes, so "custom-route" cannot be selected. Could we use separate `ScopeName` and `ScopeVersion` variants for the built-in fields, and let `ScopeAttribute(key)` always search the actual scope attributes?

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/routing.rs
ragumarimuthu-git: Done - `stringify_routing_primitive` now returns `Option<Cow<str>>`: string values are borrowed (zero allocation) and only numeric/bool values, which must be formatted, own a new `String`. The proto-span resolvers (`span_routing_value_from_attributes`/`_scope`/`_resource`) now return a borrowed `Cow` tied to the input spans/scope/resource, so the per-span and per-scope source-value `String` allocation on attribute routing is eliminated.

On the pdata log-view path the value must still be owned via `Cow::into_owned`: the view traits tie the borrowed string bytes to a per-iteration temporary that cannot outlive the resolver, so `routing_value_from_attributes`/`_scope` keep returning `Option<String>` (documented inline). Behavior is unchanged and existing routing tests pass. Done in 06c42d85.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
ragumarimuthu-git: Done - the no-mapping fast path no longer collects into a `Vec<&Span>`. `encode_span_batch` takes `impl IntoIterator<Item = &Span>` and consumes it in a single pass, so the flattened `scope_spans`/`spans` iterator is now passed directly, dropping the per-span pointer allocation on the default-table path. Behavior is unchanged. Done in cbe3b45a.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
ragumarimuthu-git: Done - the scope-invariant name is now borrowed instead of cloned per span/record:

- Spans: `resolve_span_event_name_in_scope` returns `Cow<'a, str>` and yields `Cow::Borrowed` for the resource-/scope-level `Fixed` name (resolved once per scope) and for the default table name; only a per-span mapping hit owns a `String`. The grouping loop looks up `group_index` by borrow and only owns (`into_owned`) when a new group is actually created.
- The no-mapping default path you specifically called out is now a dedicated fast path that streams the flattened span iterator straight into `encode_span_batch` (single pass), so it no longer allocates a `String` (or a `Vec`) per span at all.
- Logs: `LogRecordParts.routing_event_name` is an `Option<Cow<'a, str>>`; it is `Cow::Borrowed(table_name)` for the `Fixed`/no-mapping cases and only owns for `PerRecord`.

Done across a21d9890 (None span fast path), c832c29d (`Cow` return), f07f8852 (borrow scope-invariant name), and refined in cbe3b45a (stream spans on the no-routing path).

lalitb (approved):
LGTM. One comment be handled as follow-up - https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/712#discussion_r3642197221.

PR #714, opened 2026-07-28
chore(tower): prepare release 0.18.0

cijothomas:
Prepares opentelemetry-instrumentation-tower v0.18.0 for release.

### DIFF
--- a/opentelemetry-instrumentation-tower/CHANGELOG.md
+++ b/opentelemetry-instrumentation-tower/CHANGELOG.md
@@ -8,6 +8,7 @@ Released 2026-Jul-28
 
 ### Added
 
+* Added OpenTelemetry trace support
 * Configurable route extraction with built-in extractors:
   - `NoRouteExtractor` - No route, uses only HTTP method (e.g., `GET`), safest for cardinality
   - `PathExtractor` - Uses the URL path without query params (e.g., `/users/123`)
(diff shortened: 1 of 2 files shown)


opentelemetry-instrumentation-tower/CHANGELOG.md
```diff
@@ -2,8 +2,13 @@
...
 
+## v0.18.0
+
+Released 2026-Jul-28
+
 ### Added
 
+* Added OpenTelemetry trace support
 * Configurable route extraction with built-in extractors:
   - `NoRouteExtractor` - No route, uses only HTTP method (e.g., `GET`), safest for cardinality
```
francoposa: This is conflating two things
1. `NoRouteExtractor` returns nothing
2. The trace span name will only be the HTTP method when *any* route extractor returns nothing

```rust
        // Extract route using the configured extractor
        let route = self.route_extractor.extract_route(&req);
        let route_kv_opt = route
            .as_ref()
            .map(|r| KeyValue::new(HTTP_ROUTE_LABEL, r.clone()));

        // Build span name: "{method} {route}" or just "{method}"
        let span_name = match &route {
            Some(r) => format!("{} {}", method, r),
            None => method.clone(),
        };
```

For brevity we could avoid mentioning specifics about the span using only the method, as I think that is just the expected behavior based on the semconv.

francoposa: Possibly we could add a little extra clarification w.r.t span name in the next bullet point which currently says:

> Route extraction now provides both span names and http.route metric attribute from the same source.

e.g.

> If a route is extracted, the same value is utilized for the `http.route` attribute, and the tracing span name.

Refs:
* [HTTP Server metrics attributes](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-server)
*  [Tracing `http.route` attribute](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name)
* [Tracing span name](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name)

francoposa (approved):

jan-xyz (approved):

lalitb (approved):

PR #716, opened 2026-07-28
fix(geneva-uploader): reduce bearer-token memory remanence

manishgoel3:
Fixes # N/A
Design discussion issue (if applicable): open-telemetry/otel-arrow#3275

## Changes

Reduce GIG ingestion bearer-token memory remanence:

- Store agent-fed and GCS ingestion credentials in zeroizing secret containers.
- Zeroize successful GCS response bodies after parsing.
- Back sensitive `Authorization` headers with zeroizing application-owned memory.
- Make the agent-fed token private behind explicit construction and access APIs.
- Preserve redacted debug output and token-rotation behavior.
- Require `bytes` 1.9 for `Bytes::from_owner`.

HTTP and TLS implementations may still create transport-owned copies outside the uploader's control.

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
@@ -10,6 +10,7 @@ use crate::payload_encoder::otlp_encoder::{lookup_obo_config, MetadataFields};
 pub use crate::payload_encoder::otlp_encoder::{OboEventConfig, OboEventMap};
 use opentelemetry_proto::tonic::trace::v1::ResourceSpans;
 use otap_df_pdata_views::views::logs::LogsDataView;
+use secrecy::{ExposeSecret, SecretString};
 use std::fmt;
 use std::future::Future;
 use std::pin::Pin;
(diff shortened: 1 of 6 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
```diff
@@ -128,12 +131,31 @@ pub(crate) enum IngestionSource {
...
+/// the last request/header clone is dropped. The HTTP/TLS stack may still copy
+/// the bytes into transport-owned buffers.
+fn bearer_authorization_header(token: &SecretString) -> Result<header::HeaderValue> {
+    let token = token.expose_secret().as_bytes();
+    let mut value = Zeroizing::new(Vec::with_capacity(b"Bearer ".len() + token.len()));
+    value.extend_from_slice(b"Bearer ");
+    value.extend_from_slice(token);
+
+    let mut header =
+        header::HeaderValue::from_maybe_shared(Bytes::from_owner(value)).map_err(|error| {
```
lalitb, line 148: This only keeps the zeroizing buffer when `from_maybe_shared` receives the same `Bytes` type used by `http`. Otherwise it copies the token into normal memory. Can we document that here and update the test to verify the header uses the original buffer? The current test would still pass if a future change accidentally introduced a copy.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -991,6 +1024,8 @@ mod tests {
         assert_eq!(batches[0].event_name, "AppTrace");
     }
 
+    /// Scenario: An agent-fed source is used to initialize a Geneva client.
```
lalitb: This comment does not match the test below. The test checks span fallback and grouping, not agent-fed credential construction. 

There are also several similar comment-only changes to unrelated tests in this PR. Can we remove those unrelated comments so the security changes are easier to review?

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
manishgoel3: Updated the comment and test to verify buffer reuse.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
manishgoel3: Agreed - have removed them for now. The readability cleanup can be handled separately.

manishgoel3: @lalitb please take a look when you get a chance.

utpilla (approved):

Issue #713, opened 2026-07-27
[Tower] Scope discussion


jan-xyz:
For now the layer supports metrics and tracing for HTTP servers. However, there are at least still the client side for pure HTTP as well as gRPC semantic convention setups to do. We should find a way to structure both the project as well as the next couple of steps. That way we hopefully reduce the amount of breaking changes while still using Tower's service semantics as efficiently as possible and support the different semantic conventions. One proposal for now that includes gRPC and HTTP looks like this:

```rust
  opentelemetry_instrumentation_tower::{
      http::{
          server::{Layer, LayerBuilder, Service, RouteExtractor, PathExtractor, NoRouteExtractor},
          client::{Layer, LayerBuilder, Service},
      },
      grpc::{
          server::{Layer, LayerBuilder, Service},
          client::{Layer, LayerBuilder, Service},
      },
  }
```

with modules like this:
```
  opentelemetry-instrumentation-tower/
    src/
      lib.rs

      common/
        mod.rs
        instrumentation.rs
        propagation.rs
        attributes.rs
        status.rs

      http/
        mod.rs
        server.rs
        client.rs
        extractors.rs

      grpc/
        mod.rs
        server.rs
        client.rs
        status.rs
        trailers.rs
```
Usage:

```rust
  let layer = otel_tower::http::server::Layer::new();
  let layer = otel_tower::http::client::Layer::new();
  let layer = otel_tower::grpc::server::Layer::new();
```
```rust
  tonic::transport::Server::builder()
      .layer(otel_tower::grpc::server::Layer::new())
      .add_service(service);
```

jan-xyz:
See #700, #680

jan-xyz:
@cijothomas I brought this out of our chat into the open by recommendation of @francoposa. Do you have any opinions on how we gonna structure this, anything we should keep in mind when doing the refactoring?

francoposa:
I generally agree on structure, but I think one thing we need to be careful of is that RPC/gRPC are two separate standards in OTel: https://opentelemetry.io/docs/specs/semconv/rpc/ vs. https://opentelemetry.io/docs/specs/semconv/rpc/grpc/.

> The Semantic Conventions for [gRPC](https://grpc.io/) extend and override the [RPC Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/rpc/).

I have not yet dug into how the "extends and overrides" might impact this code structure.

[...]

PR #717, opened 2026-07-30
refactor(tower): namespace the HTTP server middleware under http::server

jan-xyz:
Design discussion issue (if applicable) #713

## Changes

This is purely splitting out parts of #700 to separate the discussion around the shape of the modules from adding HTTP client support.


opentelemetry-instrumentation-tower/benches/http_server.rs
jan-xyz, line 1: I was running these to make sure the extractions and new modules don't add any performance regressions, and it does, however, it depends a lot on the compiler settings. When setting it to
```
[profile.bench]
   lto = "thin"
   codegen-units = 1
   # https://doc.rust-lang.org/cargo/reference/profiles.html#bench
   # See function names in profiling reports.
   # 2/true is too much, 0 is not enough, 1 is just right for back traces
   debug = 1
```

There was even a slight performance increase. I still think it makes sense to force certain benchmark compiler configs to have a more stable comparison over time, wdyt? I am not a benchmark/compiler expert but it does sound reasonable to me to make these things fixtures?

opentelemetry-instrumentation-tower/src/http/extractors.rs
jan-xyz, line 1: this was previously uncovered by tests, extracting it made this more obvious. I didn't add any to not inflate the diff, let me know if it makes sense to you?

francoposa: This makes sense to me, what I'm not totally sure about without looking at potential client usage is if it makes sense for some of the extractors to stay namespaced under `server` if the signatures or usage don't match anything that would be used in client code

francoposa (approved):
This LGTM, I would check with @cijothomas or someone else that may have more strict opinions on:
1. The benchmark profile to build/run with
2. The overall vision of the namespacing/module structure w.r.t the res of the repo

But this isn't going in the release yet so we have time to fiddle with it more if needed.

opentelemetry-instrumentation-tower/src/http/extractors.rs
jan-xyz: Most of it can be re-used, apart from the `AxumMatchedPathExtractor` because it's specific to `Axum`

jan-xyz:
@cijothomas can you review this as well?

opentelemetry-instrumentation-tower/CHANGELOG.md
```diff
@@ -2,6 +2,72 @@
 
 ## vNext
 
+### Changed
+
+* **BREAKING**: Reorganized the public API into an `http::server` module with
```
cijothomas, line 7: lets add link to the PR itself, as is done always in the repo.

The changelog/migration writeup is very well don.

cijothomas (approved):
Agree with the refactoring! Thank you.

Issue #707, opened 2026-07-14
[Feature]: AWS EKS resource detector


chinmaychahar:
### Related Problems?

Child of #94

### Description

Add an EKS resource detector that populates resource attributes for services running on EKS

### What component are you working with?

opentelemetry-aws

### Describe the solution you'd like:

Use the in-cluster k8s API to:
1. Confirm we're on k8s (check `/var/run/secrets/kubernetes.io/serviceaccount/token` and `ca.crt` exist)
2. Confirm we're on EKS specifically (read `aws-auth` configmap in `kube-system`)
3. Get cluster name (read `cluster-info` configmap in `amazon-cloudwatch` namespace)

Resource attributes:
- `cloud.provider` → `aws`
- `cloud.platform` → `aws_eks`
- `k8s.cluster.name`
- `container.id` (from `/proc/self/cgroup`)

### Design constraint

`ResourceDetector::detect()` is sync. This detector needs a k8s client that can make blocking API calls. More complex than EC2/ECS due to the k8s dependency (can discuss on the approach in #94)

#### Reference

- Go implementation: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/detectors/aws/eks/detector.go

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

chinmaychahar:
Done as part of #721

Issue #94, opened 2024-07-22
[Feature]: AWS resource detectors


thomasgouveia:
### Related Problems?

The crate `opentelemetry-resource-detectors` provides only basic resource detectors. The idea is to extend the crate `opentelemetry-resource-detectors` or add a new create to add AWS resource detectors (such as done in Go [here](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/detectors/aws))

### Describe the solution you'd like:

I'd like to be able to use detectors for the following AWS services:

- EKS
- EC2
- Lambda
- ECS

### Progress

- [x] Lambda (https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/93)
- [ ] EKS
- [ ] EC2
- [ ] Lambda

cijothomas:
Would love to [get 1 or more owners](https://github.com/open-telemetry/opentelemetry-rust-contrib/issues/103#issuecomment-2330384053) for the component before adding more functionalities. If anyone would like to take ownership and maintain this crate, please reply here.

jj22ee:
Hi @cijothomas

I and @bjrara from AWS are willing to be code owners for the `opentelemetry-aws` components

cijothomas:
> Hi [@cijothomas](https://github.com/cijothomas)
> 
> I and [@bjrara](https://github.com/bjrara) from AWS are willing to be code owners for the `opentelemetry-aws` components

Sorry for the delay. Could you self submit a PR like https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/163 to list you and @bjrara as owner for the crate?

jj22ee:
Created PR here, along with another owner: https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/189

bestgopher:
In some cases, an HTTP service needs to be called, but ResourceDetector::detect is not an async function. Is there a good way to handle this?

cijothomas:
@chinmaychahar PTAL.

chinmaychahar:
Yes

chinmaychahar:
Acknowledged. I can take ownership but if anyone wants to own or contribute a specific detector, happy to coordinate. I can open new issues per detector if needed

chinmaychahar:
#### Summary of what's needed:
AWS resource detectors that automatically populate resource attributes for services running on EC2, ECS, and EKS and can scope out more. Currently we only have a Lambda detector, which works purely from env vars

[...]

PR #720, opened 2026-07-31
chore(aws): Rename `subsegment-nesting` feature

JeremieRodon:
## Changes

- `subsegment-nesting` feature renamed to `xray-subsegment-nesting` so it is clearer that it is part of the xray-* features.
- removed the `internal-logging` feature from the xray module documentation as it only take advantage of the feature but does not own it
- taking the opportunity to correct my mistake in CHANGELOG on the previous PR and correctly placed the XRay exporter in the 'Added' section instead of 'Changed' 

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes

PR #721, opened 2026-07-31
feat(aws): add EC2, ECS and EKS resource detectors

### LINKED ISSUE #705: [Feature]: AWS EC2 resource detector
### Related Problems?

Child of #94

### Description

Add an EC2 resource detector that populates resource attributes for services running on EC2 instances.

### What component are you working with?

opentelemetry-aws

### Describe the solution you'd like:

Call the [Instance Metadata Service (IMDS)](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) at `169.254.169.254` to retrieve components like:

- `cloud.provider` → `aws`
- `cloud.platform` → `aws_ec2`
- `cloud.region`
- `cloud.availability_zone`
- `cloud.account.id`
- `host.id` (instance ID)
- `host.image.id` (AMI ID)
- `host.type` (instance type)
- `host.name` (hostname)

### Design constraint

`ResourceDetector::detect()` is sync. This detector will need a blocking HTTP client (can discuss on the approach in #94)

#### Reference

- Go implementation: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/detectors/aws/ec2/v2/ec2.go

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

### LINKED ISSUE #706: [Feature]:  AWS ECS resource detector
### Related Problems?

Child of #94

### Description

Add an ECS resource detector that populates resource attributes for services running in ECS containers.

### What component are you working with?

opentelemetry-aws

### Describe the solution you'd like:

Read `ECS_CONTAINER_METADATA_URI_V4` env var, then call that HTTP endpoint to retrieve components like:

- `cloud.provider` → `aws`
- `cloud.platform` → `aws_ecs`
- `cloud.region` (parsed from task ARN)
- `cloud.account.id` (parsed from task ARN)
- `cloud.availability_zone`
- `cloud.resource_id` (container ARN)
- `container.name`
- `container.id`
- `aws.ecs.container.arn`
- `aws.ecs.cluster.arn`
- `aws.ecs.launchtype`
- `aws.ecs.task.arn`
- `aws.ecs.task.family`
- `aws.ecs.task.revision`
- `aws.log.group.names` / `aws.log.group.arns` (if awslogs driver)
- `aws.log.stream.names` / `aws.log.stream.arns` (if awslogs driver)

Note: The metadata may return short names (e.g., cluster name instead of full ARN). The detector should construct full ARNs from the base ARN when needed (Go implementation does this from task/ container ARN parts)

Should return empty resource if neither `ECS_CONTAINER_METADATA_URI` nor `ECS_CONTAINER_METADATA_URI_V4` env vars are set

#### Design constraint

`ResourceDetector::detect()` is sync. This detector will need a blocking HTTP client (can discuss on the approach in #510 

#### Reference

- Go implementation: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/detectors/aws/ecs/ecs.go

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

JeremieRodon:
Fixes #705
Fixes #706
Part of #707
Part of #94

## Changes

Adds three AWS resource detectors to the `opentelemetry-aws` crate, each behind its own feature flag:

- `Ec2ResourceDetector` (`detector-aws-ec2`) — queries IMDSv2 for `cloud.*` and `host.*` attributes.
- `EcsResourceDetector` (`detector-aws-ecs`) — reads the ECS container metadata   endpoint v4 (`ECS_CONTAINER_METADATA_URI_V4`) and IMDSv2 on EC2 workers for `cloud.*`, `aws.ecs.*`,
  `container.*` and `aws.log.*` attributes.
- `EksResourceDetector` (`detector-aws-eks`) — detects a pod via the
  service-account namespace file and ties it to AWS via IMDSv2, reporting
  `cloud.*`, `k8s.*`, `aws.eks.cluster.arn`, `container.id` and node `host.*`.

Supporting changes:

- New internal feature groups `_detector` / `_detector-http` to share deps.
- New optional dependency `ureq` (blocking HTTP client — `ResourceDetector::detect`
  is synchronous). Empty attribute values are filtered out; errors are best-effort
  logged (gated by `internal-logs`) and the attribute is skipped.
- The existing `LambdaResourceDetector` is intentionally untouched.

### Usage

```rust
use opentelemetry_aws::detector::Ec2ResourceDetector;
use opentelemetry_sdk::Resource;

let resource = Resource::builder()
    .with_detector(Box::new(Ec2ResourceDetector))
    .build();
```

# Discussion / open questions
I'd like maintainer input on the following before finalizing:
1. Warning on wrong platform. A detector run on the wrong platform emits a
single warning log line (disableable by dropping the internal-logs
feature). Pro: surfaces likely misconfiguration cheaply. Con: an app that
legitimately runs on multiple platforms (e.g. EC2 and ECS) and enables
several detectors will always warn for the ones that don't match. Keep,
downgrade to debug, or drop?
2. IMDS use in ECS/EKS vs. Fargate. Both ECS and EKS detectors contact
IMDSv2 to enrich with node info, which does not work on Fargate. Should these
detectors instead focus only on ECS/EKS-specific sources and leave node
detection to the EC2 detector (user opts in by also enabling it)? If so, how
should cloud.platform be resolved when several detectors are chained —
relying on detector order is poor API design; is there a better contract?
3. IMDSv2 only. I only implement IMDSv2 (no v1 fallback). Is v1 support
wanted? Relatedly, endpoints use /latest/; should a pinned, dated IMDS
version be used instead?
4. Empty-value filtering. Every detector filters out empty/blank values to
avoid populating keys with empty strings. Is the added code worth it, or is
it acceptable to emit empty values?
5. EKS approach differs from #707 which proposes the k8s-API route
(read aws-auth / cluster-info configmaps, container.id from cgroup).
This PR instead uses the service-account namespace file + IMDSv2
aws:eks:cluster-name instance tag + downward-API env vars, which avoids a
Kubernetes client dependency but requires instance-tags-in-metadata to be
enabled (or AWS_CLUSTER_NAME to be set) for k8s.cluster.name. Is this
trade-off acceptable, or should it follow #707's configmap approach?
6. Refactor the Lambda detector? The existing lambda.rs populates
attributes with unwrap_or_default(), producing empty values on error —
inconsistent with the empty-filtering the new detectors do. Should a small
refactor be included in this PR, or kept separate?
 # Testing

Unit tests cover the pure logic (arch mapping, value filtering, ARN parsing, etc.).

End-to-end behavior was verified by running a probe program on real AWS (EC2, ECS, EKS). See: https://github.com/RustyServerless/opentelemetry-rust-contrib/tree/wip/aws-resource-detectors/otel-aws-probe-deploy

# Merge requirement checklist

* [x] CONTRIBUTING (https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Appropriate CHANGELOG.md files updated for non-trivial, user-facing changes

### DIFF
--- a/opentelemetry-aws/src/detector/ec2.rs
+++ b/opentelemetry-aws/src/detector/ec2.rs
@@ -29,28 +29,13 @@ const DETECTOR: &str = "aws_ec2";
 /// | `host.arch`               | identity document `architecture`, mapped to `host.arch` values |
 /// | `host.name`               | `/latest/meta-data/hostname`                                   |
 ///
-/// All but `host.name` come from the instance identity document
-/// (`/latest/dynamic/instance-identity/document`), so detection costs three
-/// requests in total: the IMDSv2 token, the document and the hostname.
-///
 /// Values that cannot be found are skipped.
 ///
-/// # Feature flag
-///
-/// This type is only available when the `detector-aws-ec2` Cargo feature is
-/// enabled.
-///
-/// # Behavior
-///
-/// Detection is best-effort. Any IMDSv2 call that fails (network error, HTTP
-/// error, or JSON parse error) is silently skipped and the corresponding
-/// attribute is omitted from the returned [`Resource`].
+/// # Probing
 ///
 /// The platform probe is the acquisition of an IMDSv2 session token *and* the
 /// successful parse of the instance identity document. Unless both succeed the
-/// environment is assumed not to be EC2 and an empty [`Resource`] is returned,
-/// so that `cloud.provider` and `cloud.platform` are never asserted
-/// off-platform.
+/// environment is assumed not to be EC2 and an empty [`Resource`] is returned.
 ///
 /// # Blocking
 ///
--- a/opentelemetry-aws/src/detector/ec2.rs
+++ b/opentelemetry-aws/src/detector/ec2.rs
@@ -1,0 +1,116 @@
+use opentelemetry::KeyValue;
+use opentelemetry_sdk::{resource::ResourceDetector, Resource};
+use opentelemetry_semantic_conventions::attribute as semco;
+
+use super::{
+    imds::ImdsClient,
+    utils::{opt_kv, warn_on_error},
+};
+
+/// Name reported in internal logs emitted by this detector.
+const DETECTOR: &str = "aws_ec2";
+
+/// EC2 resource detector (`detector-aws-ec2` feature).
+///
+/// Queries the EC2 Instance Metadata Service v2 (IMDSv2) at the link-local
+/// address `169.254.169.254` and returns an OTel [`Resource`] with the
+/// following attributes:
+///
+/// | OTel attribute            | Source                                                         |
+/// |---------------------------|----------------------------------------------------------------|
+/// | `cloud.provider`          | hardcoded `"aws"`                                              |
+/// | `cloud.platform`          | hardcoded `"aws_ec2"`                                          |
+/// | `cloud.account.id`        | identity document `accountId`                                  |
+/// | `cloud.region`            | identity document `region`                                     |
+/// | `cloud.availability_zone` | identity document `availabilityZone`                           |
+/// | `host.id`                 | identity document `instanceId`                                 |
+/// | `host.type`               | identity document `instanceType`                               |
+/// | `host.image.id`           | identity document `imageId`                                    |
+/// | `host.arch`               | identity document `architecture`, mapped to `host.arch` values |
+/// | `host.name`               | `/latest/meta-data/hostname`                                   |
+///
+/// All but `host.name` come from the instance identity document
+/// (`/latest/dynamic/instance-identity/document`), so detection costs three
+/// requests in total: the IMDSv2 token, the document and the hostname.
+///
+/// Values that cannot be found are skipped.
+///
+/// # Feature flag
+///
+/// This type is only available when the `detector-aws-ec2` Cargo feature is
+/// enabled.
+///
+/// # Behavior
+///
+/// Detection is best-effort. Any IMDSv2 call that fails (network error, HTTP
+/// error, or JSON parse error) is silently skipped and the corresponding
+/// attribute is omitted from the returned [`Resource`].
+///
+/// The platform probe is the acquisition of an IMDSv2 session token *and* the
+/// successful parse of the instance identity document. Unless both succeed the
+/// environment is assumed not to be EC2 and an empty [`Resource`] is returned,
+/// so that `cloud.provider` and `cloud.platform` are never asserted
+/// off-platform.
+///
+/// # Blocking
+///
+/// Detection performs blocking HTTP requests, each capped by a one second
+/// timeout, and will therefore stall the calling thread.
+///
+/// # Examples
+///
+/// Register the detector with the OpenTelemetry SDK so that EC2 attributes are
+/// automatically merged into the global resource:
+///
+/// ```no_run
+/// // Requires a running EC2 instance with IMDSv2 enabled.
+/// use opentelemetry_aws::detector::Ec2ResourceDetector;
+/// use opentelemetry_sdk::Resource;
+///
+/// let resource = Resource::builder()
+///     .with_detector(Box::new(Ec2ResourceDetector))
+///     .build();
+/// ```
+///
+/// [`Resource`]: opentelemetry_sdk::Resource
+pub struct Ec2ResourceDetector;
+
+impl ResourceDetector for Ec2ResourceDetector {
+    fn detect(&self) -> Resource {
+        // Platform probe. A session token only proves that *something* answers
+        // at the link-local address; a parsable identity document is what
+        // proves it is IMDSv2. Both failures are routine off-platform.
+        let Some(imds) = warn_on_error(DETECTOR, ImdsClient::new()) else {
+            // Not EC2, return empty resource
+            return Resource::builder_empty().build();
+        };
+        let Some(document) = warn_on_error(DETECTOR, imds.get_identity_document()) else {
+            // Not EC2, return empty resource
+            return Resource::builder_empty().build();
+        };
+
+        // Past this point the environment is believed to be EC2.
+        let attribute_options = [
+            Some(KeyValue::new(semco::CLOUD_PROVIDER, "aws")),
+            Some(KeyValue::new(semco::CLOUD_PLATFORM, "aws_ec2")),
+            document
+                .host_arch()
+                .map(|arch| KeyValue::new(semco::HOST_ARCH, arch)),
+            opt_kv(semco::CLOUD_ACCOUNT_ID, document.account_id),
+            opt_kv(semco::CLOUD_REGION, document.region),
+            opt_kv(semco::CLOUD_AVAILABILITY_ZONE, document.availability_zone),
+            opt_kv(semco::HOST_ID, document.instance_id),
+            opt_kv(semco::HOST_TYPE, document.instance_type),
+            opt_kv(semco::HOST_IMAGE_ID, document.image_id),
+            // Host name — the only attribute absent from the identity document
+            opt_kv(
+                semco::HOST_NAME,
+                warn_on_error(DETECTOR, imds.get("hostname")),
+            ),
+        ];
+
+        Resource::builder_empty()
+            .with_attributes(attribute_options.into_iter().flatten())
+            .build()
+    }
+}
--- a/opentelemetry-aws/src/detector/ec2.rs
+++ b/opentelemetry-aws/src/detector/ec2.rs
@@ -4,7 +4,7 @@ use opentelemetry_semantic_conventions::attribute as semco;
 
 use super::{
     imds::{ImdsClient, ImdsError, ImdsProvider},
-    utils::{opt_kv, warn_on_error},
+    utils::{info_on_error, opt_kv, warn_on_error},
 };
 
 /// Name reported in internal logs emitted by this detector.
@@ -83,14 +83,12 @@ impl ResourceDetector for Ec2ResourceDetector {
 
 impl Ec2ResourceDetector {
     fn detect_from<P: ImdsProvider>(imds: Result<P, ImdsError>) -> Resource {
-        // Platform probe. A session token only proves that *something* answers
-        // at the link-local address; a parsable identity document is what
-        // proves it is IMDSv2. Both failures are routine off-platform.
-        let Some(imds) = warn_on_error(DETECTOR, imds) else {
-            // Not EC2, return empty resource
+        // Platform probe.
+        let Some(imds) = info_on_error(DETECTOR, imds) else {
+            // Not EC2/IMDSv2, return empty resource
             return Resource::builder_empty().build();
         };
-        let Some(document) = warn_on_error(DETECTOR, imds.get_identity_document()) else {
+        let Some(document) = info_on_error(DETECTOR, imds.get_identity_document()) else {
             // Not EC2, return empty resource
             return Resource::builder_empty().build();
         };
--- a/opentelemetry-aws/src/detector/ec2.rs
+++ b/opentelemetry-aws/src/detector/ec2.rs
@@ -3,7 +3,7 @@ use opentelemetry_sdk::{resource::ResourceDetector, Resource};
 use opentelemetry_semantic_conventions::attribute as semco;
 
 use super::{
-    imds::ImdsClient,
+    imds::{ImdsClient, ImdsError, ImdsProvider},
     utils::{opt_kv, warn_on_error},
 };
 
@@ -77,10 +77,16 @@ pub struct Ec2ResourceDetector;
 
 impl ResourceDetector for Ec2ResourceDetector {
     fn detect(&self) -> Resource {
+        Self::detect_from(ImdsClient::new())
+    }
+}
+
+impl Ec2ResourceDetector {
+    fn detect_from<P: ImdsProvider>(imds: Result<P, ImdsError>) -> Resource {
         // Platform probe. A session token only proves that *something* answers
         // at the link-local address; a parsable identity document is what
         // proves it is IMDSv2. Both failures are routine off-platform.
-        let Some(imds) = warn_on_error(DETECTOR, ImdsClient::new()) else {
+        let Some(imds) = warn_on_error(DETECTOR, imds) else {
             // Not EC2, return empty resource
             return Resource::builder_empty().build();
         };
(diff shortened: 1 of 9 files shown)


chinmaychahar:
It's well-built code at this scale. Thanks for that! Thorough docs, unit tests and live testing

chinmaychahar:
> Warning on wrong platform. A detector run on the wrong platform emits a
single warning log line (disableable by dropping the internal-logs
feature). Pro: surfaces likely misconfiguration cheaply. Con: an app that
legitimately runs on multiple platforms (e.g. EC2 and ECS) and enables
several detectors will always warn for the ones that don't match. Keep,
downgrade to debug, or drop?

My leaning is to downgrade to debug, what do you think? But if we are on the platform and a call that should've worked fails, that still feels worth a warn

> IMDS use in ECS/EKS vs. Fargate. Both ECS and EKS detectors contact
IMDSv2 to enrich with node info, which does not work on Fargate. Should these
detectors instead focus only on ECS/EKS-specific sources and leave node
detection to the EC2 detector (user opts in by also enabling it)? If so, how
should cloud.platform be resolved when several detectors are chained —
relying on detector order is poor API design; is there a better contract?


We can keep IMDS in the ECS/EKS detectors, seems safe without having to add another detector, but I'll need to analyze this more

> IMDSv2 only. I only implement IMDSv2 (no v1 fallback). Is v1 support
wanted? Relatedly, endpoints use /latest/; should a pinned, dated IMDS
version be used instead?

v2-only should be fine. AWS is pushing v2 as the default. @cijothomas could answer more if there are any OTel standards. Also, I'd keep /latest/. Is there a specific field you're worried about?

chinmaychahar:
> Empty-value filtering. Every detector filters out empty/blank values to
avoid populating keys with empty strings. Is the added code worth it, or is
it acceptable to emit empty values?

I think an empty data can look like real data and break things downstream, unlike null so since these are utils and a few small re-used helpers (with tests) - we should be good to keep 

> EKS approach differs from https://github.com/open-telemetry/opentelemetry-rust-contrib/issues/707 which proposes the k8s-API route
(read aws-auth / cluster-info configmaps, container.id from cgroup).
This PR instead uses the service-account namespace file + IMDSv2
aws:eks:cluster-name instance tag + downward-API env vars, which avoids a
Kubernetes client dependency but requires instance-tags-in-metadata to be
enabled (or AWS_CLUSTER_NAME to be set) for k8s.cluster.name. Is this
trade-off acceptable, or should it follow https://github.com/open-telemetry/opentelemetry-rust-contrib/issues/707's configmap approach?

I looked at how the other SDKs do it and Go uses the full client-go library, but Python and JS both just hit the k8s API directly over plain HTTPS with the service-account toke. So, curious to know how quantitative are the benefits of avoiding the kubernetes client dependency? I'm good with your lighter-weight approach, just wanna understand more. 

Also, right now the platform tie is AWS_REGION ([eks.rs:170](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-aws/src/detector/eks.rs#L170), [eks.rs:185](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-aws/src/detector/eks.rs#L185)) and AWS_REGION is set by pretty much any app which will use a AWS SDK, so what if a GKE/AKS pod reads from S3 and gets labelled as cloud.platform=aws_eks? Let's think about both happy cases and edge cases

> Refactor the Lambda detector? The existing lambda.rs populates
attributes with unwrap_or_default(), producing empty values on error —
inconsistent with the empty-filtering the new detectors do. Should a small
refactor be included in this PR, or kept separate?

Let's  keep it separate, what do you think? This PR is already big

chinmaychahar:
As you work on the re-factor, the only red check is codecov currently - can we add more mockable scenarios?

JeremieRodon:
@chinmaychahar

> > Warning on wrong platform. A detector run on the wrong platform emits a
> > single warning log line (disableable by dropping the internal-logs
> > feature). Pro: surfaces likely misconfiguration cheaply. Con: an app that
> > legitimately runs on multiple platforms (e.g. EC2 and ECS) and enables
> > several detectors will always warn for the ones that don't match. Keep,
> > downgrade to debug, or drop?
> 
> My leaning is to downgrade to debug, what do you think? But if we are on the platform and a call that should've worked fails, that still feels worth a warn

I felt that a single warn at startup for an enabled detector that cannot find its platform was not a big deal, which is why I took that option; but I will happily change it to info (or debug) if you prefer. I agree that the warn would be useful in situations where the dev made a genuine mistake or the env behavior is unexpected somehow... But I don't see how we could know the difference.

> We can keep IMDS in the ECS/EKS detectors, seems safe without having to add another detector, but I'll need to analyze this more

Ok I keep it as-is for now.

> > IMDSv2 only. I only implement IMDSv2 (no v1 fallback). Is v1 support
> > wanted? Relatedly, endpoints use /latest/; should a pinned, dated IMDS
> > version be used instead?
> 
> v2-only should be fine. AWS is pushing v2 as the default. @cijothomas could answer more if there are any OTel standards. Also, I'd keep /latest/. Is there a specific field you're worried about?

Ok for v2-only.
Regarding /latest/ I'm not worried about anything in particular. I raised the point mainly out of design/philosophical concerns: using /latest/ *do* create the potential for outage if AWS drop the ball on retro-compatibility and I'm not sure of OTel general policy on the subject.

> > Empty-value filtering. Every detector filters out empty/blank values to
> > avoid populating keys with empty strings. Is the added code worth it, or is
> > it acceptable to emit empty values?
> 
> I think an empty data can look like real data and break things downstream, unlike null so since these are utils and a few small re-used helpers (with tests) - we should be good to keep

Ok.

> > Refactor the Lambda detector? The existing lambda.rs populates
> > attributes with unwrap_or_default(), producing empty values on error —
> > inconsistent with the empty-filtering the new detectors do. Should a small
> > refactor be included in this PR, or kept separate?
> 
> Let's keep it separate, what do you think? This PR is already big

Sure. I may work on that after this one is merged.

JeremieRodon:
> > EKS approach differs from #707 which proposes the k8s-API route
> > (read aws-auth / cluster-info configmaps, container.id from cgroup).
> > This PR instead uses the service-account namespace file + IMDSv2
> > aws:eks:cluster-name instance tag + downward-API env vars, which avoids a
> > Kubernetes client dependency but requires instance-tags-in-metadata to be
> > enabled (or AWS_CLUSTER_NAME to be set) for k8s.cluster.name. Is this
> > trade-off acceptable, or should it follow #707 configmap approach?
> 
> I looked at how the other SDKs do it and Go uses the full client-go library, but Python and JS both just hit the k8s API directly over plain HTTPS with the service-account toke. So, curious to know how quantitative are the benefits of avoiding the kubernetes client dependency? I'm good with your lighter-weight approach, just wanna understand more.

To be honest, I'm not a k8s person. I was able to review and correct the code as it is because it's env vars and IMDS, but I don't feel like tackling the k8s API (I explicitly prevented the agent from taking this route for that reason). I can scrap the k8s detector from the PR, or we can keep it simple and let some k8s enthusiast upgrade it later. What do you prefer? 

> Also, right now the platform tie is AWS_REGION ([eks.rs:170](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-aws/src/detector/eks.rs#L170), [eks.rs:185](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-aws/src/detector/eks.rs#L185)) and AWS_REGION is set by pretty much any app which will use a AWS SDK, so what if a GKE/AKS pod reads from S3 and gets labelled as cloud.platform=aws_eks? Let's think about both happy cases and edge cases

Ok I can use only `cluster_name` then. It is retrieved either from tags in IMDS or from the `AWS_CLUSTER_NAME` env-var so it is unlikely to show-up outside of AWS. But should we warn when it is absent then? It seems very likely that the user forgot to set the `AWS_CLUSTER_NAME` or enable the tags in IMDS. If they decided to enable this detector, should we assume that they are indeed on AWS and log a hint as to why the detector failed?

> As you work on the re-factor, the only red check is codecov currently - can we add more mockable scenarios?

Ok I will work on that.

JeremieRodon:
@chinmaychahar I increased the test coverage.

I chose a path similar to what was already done for the Lambda detector for local files and env var. For the IMDS and ECS Metadata Service, I implemented fake clients storing raw strings (rather than fake local servers) so the tests cover deserialization as well as detectors logic.

I let the agent pick the scenarii, after review it looks ok to me. Maybe a little overkill for ECS which make it look verbose, but that's mainly because the ECS detector outputs a lot of attributes.

JeremieRodon:
Give me your final thoughts on:
- the warn vs info for detector failure (I think debug would really be going too low);
- k8s detector keep simple or drop entirely from this PR;

and I will be able to wrap it up.

chinmaychahar:
> the warn vs info for detector failure (I think debug would really be going too low)

info works for me for the case where detector is enabled but not on the current platform. I'd only keep warn for when one has already confirmed they are on the right platform but a call that should've worked fails. Like the EKS case, where we know we're in a pod but cluster_name isn't set. That might be a misconfig and worth a warn

> k8s detector keep simple or drop entirely from this PR

let's keep it. It's well documented and I'm happy to pick up that enhancement myself. No need to drop it

chinmaychahar:
> Ok I can use only cluster_name then. It is retrieved either from tags in IMDS or from the AWS_CLUSTER_NAME env-var so it is unlikely to show-up outside of AWS. But should we warn when it is absent then? It seems very likely that the user forgot to set the AWS_CLUSTER_NAME or enable the tags in IMDS. If they decided to enable this detector, should we assume that they are indeed on AWS and log a hint as to why the detector failed?

yes, let's keep a warn

JeremieRodon:
Done. I also rebased on top of main, hence the force-push

JeremieRodon:
Rebased following the merge of PR #720 into main

chinmaychahar:
I'm reviewing it

opentelemetry-aws/src/detector/ec2.rs
```diff
@@ -0,0 +1,255 @@
...
+/// | `host.image.id`           | identity document `imageId`                                    |
+/// | `host.arch`               | identity document `architecture`, mapped to `host.arch` values |
+/// | `host.name`               | `/latest/meta-data/hostname`                                   |
+///
+/// All but `host.name` come from the instance identity document
+/// (`/latest/dynamic/instance-identity/document`), so detection costs three
+/// requests in total: the IMDSv2 token, the document and the hostname.
+///
+/// Values that cannot be found are skipped.
+///
```
chinmaychahar, line 33: Should we standardize the documentation for all four detectors we have currently - Lambda, EC2, ECS, EKS? It can be a good mix of in-line code documentations or even a local readme for these detectors

chinmaychahar: Put simply, I think giant documentation blocks for these detectors inside the code might not be the best way to put it. What do you think?

Of course, if there's such precedence in OTel and they recommend it, I'm unaware

opentelemetry-aws/src/detector/eks.rs
```diff
@@ -0,0 +1,795 @@
...
+/// network error, missing environment variable) is silently skipped and the
+/// corresponding attribute is omitted from the returned [`Resource`].
+///
+/// Detection is gated on two probes, both of which must pass before
+/// `cloud.provider` and `cloud.platform` are asserted:
+///
+/// 1. the service-account namespace file must be readable, which proves the
+///    process runs in a Kubernetes pod;
+/// 2. something must tie that pod to AWS — either a parsable instance identity
+///    document from IMDSv2, or one of the `AWS_CLUSTER_NAME` and `AWS_REGION`
```
chinmaychahar: This doc looks stale, we gate only on cluster_name now. Could we update it?

opentelemetry-aws/src/detector/ecs.rs
```diff
@@ -0,0 +1,1214 @@
...
+    }
+}
+
+impl EcsResourceDetector {
+    fn detect_from<E: EcsMetadataProvider, I: ImdsProvider>(
+        ecs: Result<E, EcsMetadataError>,
+        imds: impl FnOnce() -> Result<I, ImdsError>,
+    ) -> Resource {
+        // Platform probe: the metadata URI is injected by the ECS agent, so its
+        // absence is the normal case off-platform and only worth a debug event.
```
chinmaychahar: Nit: says "debug" but we use info_on_error now

opentelemetry-aws/src/detector/eks.rs
```diff
@@ -0,0 +1,795 @@
...
+impl EksResourceDetector {
+    fn detect_from<P: ImdsProvider>(
+        imds: Result<P, ImdsError>,
+        namespace_path: &Path,
+        cgroup_path: &Path,
+        mountinfo_path: &Path,
+    ) -> Resource {
+        // Kubernetes probe: without the service-account mount, this is not a
+        // pod at all. An unreadable file is the normal case off-Kubernetes and
+        // so is only worth a debug event.
```
chinmaychahar: same nit here - says "debug" but we use info_on_error

opentelemetry-aws/src/detector/ec2.rs
chinmaychahar: Thinking more on this and understanding the pattern, I think the big rustdoc blocks should be fine as I read that they render on docs.rs and in IDEs. 

Could we just trim the long caveats and keep the four detectors consistent? I think the lambda detector does not have the same doc styles

opentelemetry-aws/src/detector/eks.rs
```diff
@@ -0,0 +1,795 @@
...
+        // The cluster ARN needs a partition, which costs an extra IMDS request,
+        // so it is only fetched once the rest of the ARN is known.
+        // If partition is not retrievable from IMDS, assume "aws".
+        let cluster_arn = match (&region, &account_id) {
+            (Some(region), Some(account_id)) => {
+                let partition = imds
+                    .as_ref()
+                    .and_then(|imds| warn_on_error(DETECTOR, imds.get("services/partition")))
+                    .and_then(non_empty)
+                    .unwrap_or_else(|| "aws".to_owned());
```
chinmaychahar: The doc above says ` The partition is only available from IMDSv2, so the ARN is never reported when IMDS is unreachable, even if the cluster name is configured` but here we fall back to "aws" and still emit it on EC2/Fargate.

I think hardcoding aws will be a bug in GovCloud `aws-us-gov` and China `aws-cn` on Fargate and we'll get an incorrect ARN. On EC2 it's fine since the partition comes from IMDS. How do you think we fix this?

chinmaychahar:
Great work @JeremieRodon , just a couple of (majorly doc) fixes to look into

chinmaychahar:
I will take a look into the cargo-deny failure

JeremieRodon:
> I will take a look into the cargo-deny failure

It is unrelated to the PR, it just so happened that a vulnerability in the version of `h2` used by the project was found on August 17th. It should be more or less a workspace-wise issue caused by `opentelemetry-instrumentation-actix-web` from my understanding.

chinmaychahar:
@JeremieRodon the cardo-deny failure has been fixed for this repo. Please push another commit to re-run CI

JeremieRodon:
@chinmaychahar 
Now it's Rust 1.98 lints that fail on `opentelemetry-stackdriver` because of a clippy fix that previously prevented the result_large_err lint from detecting in async functions: https://github.com/rust-lang/rust-clippy/pull/17130

opentelemetry-aws/src/detector/eks.rs
chinmaychahar: @JeremieRodon could you check this?

chinmaychahar:
could you also rebase to mainline again? I believe the above errors should be fixed now :D

JeremieRodon:
> could you also rebase to mainline again? I believe the above errors should be fixed now :D

Done. And at last all relevant checks are passing!

chinmaychahar (approved):

chinmaychahar:
LGTM! 🚀 

Thanks @JeremieRodon! You can mark the comments resolved

cc: @cijothomas to help review/merge

opentelemetry-aws/src/detector/ec2.rs
JeremieRodon: Yes the doc-comments are Markdown documents and rendered as such by doc.rs and IDEs.
In this case, they will serve to render the IDE doc card when someone hovers over (e.g.) `Ec2ResourceDetector` in:
```rs
let resource = Resource::builder()
    .with_detector(Box::new(Ec2ResourceDetector))
    .build();
```
so I think we want it as descriptive/useful as possible.

I agree the Lambda detector does not have the same style, but that's because it virtually does not have any documentation. A user of `LambdaResourceDetector` will only see "Resource detector that collects resource information from AWS Lambda environment." when hovering it. I don't believe that can be considered useful (I personally consider `tokio` and the standard lib as references for "good doc"). I'm still planing to refactor lambda.rs (beyond the doc, the behavior in case of missing values is also different) but you (rightly) pointed out that this PR was already big enough. So I propose we keep out of lambda.rs for this PR, and I will do another one to update its behavior *and* doc.

To reduce the verbosity, I propose to trim:
- the "Feature flag" section, which is redundant with the first doc line and only useful in doc.rs;
- the entire "Behavior" section, I admit it is hard to maintain (as demonstrated by my own PR...) and users interested in this level of details might as well read the code.

I advise to keep the detailed table on the origin of each field, even if it is hard to maintain. People reading the doc card will do so as they write the code for their OTel Resource, which is all about the attributes that will be attached to every spans of their project, so I think it is very useful.

I also think the "Blocking" section is short and low maintenance effort and should be kept.

opentelemetry-aws/src/detector/eks.rs
JeremieRodon: Following your first remark, I reckon that's too much details and too hard to maintain. I proposed to simply drop the section and will proceed to do so.

opentelemetry-aws/src/detector/eks.rs
JeremieRodon: IMDS is always absent for EKS/Fargate, so I took the default option to avoid the ARN systematically missing on Fargate.
But I will check if there is another way to retrieve it.

If not, we have to choose:
- make it so the ARN is always absent on Fargate (which will be a fairly common situation);
- make it so the ARN is wrong on Fargate in China (Gov and Sovereign EU does not support Fargate for EKS to date) and document the caveat;
- designate an env var to fill the gap?

JeremieRodon: Given that the partition is used only to construct the ARN, and we must have a region for that, I propose to use the region to infer the partition. I'll proceed to write code in that direction and drop the IMDS request entirely.
If the region is absent, then we cannot produce an ARN anyway.

opentelemetry-aws/src/detector/ecs.rs
JeremieRodon: I removed all justifications for te log level from comments

opentelemetry-aws/src/detector/eks.rs
JeremieRodon: Done

chinmaychahar (approved):
Nice, thanks for addressing the threads. I hope no open questions are left

JeremieRodon:
> Nice, thanks for addressing the threads. I hope no open questions are left

I addressed everything apart from things related to `lambda.rs` which I propose to address in another PR once this one is merged.

chinmaychahar:
Hey hey @cijothomas could you help us close this and move to merge queue?

cijothomas (approved):

PR #736, opened 2026-08-03
fix(resource-detectors): disable opentelemetry default features

mwear:
## Changes

This addresses the feedback from https://github.com/open-telemetry/otel-arrow/pull/3555#discussion_r3692141624.

Currently, `opentelemetry-resource-detectors` inherits `opentelemetry` from the workspace with default features enabled. That pulls `futures-*` and `thiserror` into downstream builds that depend on just this crate.

This declares `opentelemetry` directly with `default-features = false` so the crate requests only what it needs. `opentelemetry_sdk` already uses `default-features = false` at the workspace level, so it's unchanged.

Note: `{ workspace = true, default-features = false }` does not work here. Cargo ignores `default-features` when the workspace dependency doesn't specify it.

* [ ] Unit tests added/updated (if applicable) — n/a, dependency-only change
* [ ] Changes in public API reviewed (if applicable) — no public API change


mwear:
I just wanted to check in on this. I'm not sure why the codecov check is failing. This PR doesn't modify any code. Is there anything you need from me?

lalitb (approved):

PR #737, opened 2026-08-04
fix: Handle ETW event write failures non-fatally

utpilla:
## Summary
The ETW logs and traces exporters previously called `debug_assert!(false, ...)` when `EventBuilder::write()` returned a nonzero Win32 error. This results in a process-terminating panic in debug and `debug-assertions`-enabled builds. This PR replaces the assertion with a non-fatal internal diagnostic that logs the Win32 error code and continues.

### DIFF
--- a/opentelemetry-etw-logs/CHANGELOG.md
+++ b/opentelemetry-etw-logs/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## vNext
 
+- A failed ETW event write (`EventBuilder::write` returning a
+  nonzero Win32 error) no longer triggers a `debug_assert!` panic. ETW write
+  failures are expected, transient runtime conditions (e.g. buffer pressure from
+  an active session) that Microsoft documents as diagnostic-only, so the
+  exporter now logs the Win32 error code via `otel_debug!` and continues instead
+  of terminating the process in debug or `debug-assertions`-enabled builds.
+
 ## v0.11.0
 
 - Bump opentelemetry and opentelemetry_sdk versions to 0.32
--- a/opentelemetry-etw-logs/CHANGELOG.md
+++ b/opentelemetry-etw-logs/CHANGELOG.md
@@ -2,13 +2,6 @@
 
 ## vNext
 
-- A failed ETW event write (`EventBuilder::write` returning a
-  nonzero Win32 error) no longer triggers a `debug_assert!` panic. ETW write
-  failures are expected, transient runtime conditions (e.g. buffer pressure from
-  an active session) that Microsoft documents as diagnostic-only, so the failed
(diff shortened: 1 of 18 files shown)


opentelemetry-etw-logs/src/exporter/mod.rs
```diff
@@ -125,10 +125,12 @@ impl ETWExporter {
...
-            // The return value is for diagnostic purposes only and should generally be ignored in retail builds.
-            match result {
-                0 => (),
-                _ => debug_assert!(false, "Failed to write event to ETW. ETW reason: {result}"),
+            // A nonzero result is an expected, transient runtime condition (e.g. buffer pressure from an
+            // active session), not a programming error. Microsoft documents the code as diagnostic-only:
+            // production code should keep running even if an event cannot be written. Log it and continue
+            // rather than panicking.
+            if result != 0 {
+                otel_debug!(name: "EtwLogs.WriteFailed", error_code = result);
```
cijothomas: we shouldn't log from inside exporter due to live-loop and stack overflow crash.

utpilla: Ah I'm not sure how I missed this 😅

Thanks for catching that! I have now updated the code to just discard the result. I think that's the right thing to do since it doesn't seem reasonable to panic for transient errors. Long-term, we should look at adding metrics for improving diagnosability.

lalitb (approved):
LGTM - Good to add a tracking issue for adding the metrics for error.

opentelemetry-etw-logs/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## vNext
 
+- A failed ETW event write (`EventBuilder::write` returning a
```
cijothomas: too big of a changelog for a debug-build only fix. I think we don't even need a changelog. if keeping, trim it to one line.

opentelemetry-etw-logs/src/exporter/mod.rs
```diff
@@ -121,15 +121,18 @@ impl ETWExporter {
...
-
-            // event.write() above returns 0 for success or a Win32 error from EventWrite for failure.
-            // The return value is for diagnostic purposes only and should generally be ignored in retail builds.
-            match result {
-                0 => (),
-                _ => debug_assert!(false, "Failed to write event to ETW. ETW reason: {result}"),
-            }
+            // Write event to ETW.
+            //
+            // event.write() returns 0 for success or a Win32 error from EventWrite for failure.
```
cijothomas: current, minimal comment is good as-is.
lets also add that data losses due to this is better tracked via *future* metrics.

cijothomas (approved):

PR #744, opened 2026-08-11
feat(geneva): gate certificate auth behind feature

lalitb:
## Summary

- move PKCS#12 client certificate authentication behind a new opt-in `certificate-auth` feature
- keep `tls-rustls` independent of `p12-keystore` for builds using managed identity, workload identity, or other non-certificate authentication
- forward `certificate-auth` through the Geneva exporter and FFI crates
- disable `p12-keystore`'s legacy PBES1 support so opt-in builds do not include RC2 or DES

## Why

The rustls backend currently enables `p12-keystore` for every build because rustls cannot parse PKCS#12 identities itself. This pulls PKCS#12 parsing and cryptographic dependencies into applications that never configure certificate authentication.

Certificate authentication is now an explicit capability. Existing users of `AuthMethod::Certificate` must enable `certificate-auth`; otherwise client construction returns a clear error directing them to enable the feature or use another authentication method.

## Validation

- `cargo check -p geneva-uploader`
- `cargo check -p geneva-uploader --features certificate-auth`
- `cargo check -p geneva-uploader --no-default-features --features tls-rustls`
- `cargo check -p geneva-uploader --no-default-features --features tls-rustls,certificate-auth`
- focused tests for disabled and enabled certificate authentication
- clippy with `-D warnings` for rustls with and without `certificate-auth`
- rustls compile checks for `opentelemetry-exporter-geneva` and `geneva-uploader-ffi`
- verified the default rustls dependency graph does not contain `p12-keystore`
- verified the opt-in certificate dependency graph does not contain `rc2`


opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -292,77 +298,90 @@ impl GenevaConfigClient {
...
-                let p12_bytes = fs::read(path).map_err(|e| {
-                    debug!(
-                        name: "config_client.new.certificate_read_error",
-                        target: "geneva-uploader",
-                        error = %e,
-                        "Failed to read certificate file"
-                    );
-                    GenevaConfigClientError::Certificate(e.to_string())
-                })?;
+            AuthMethod::Certificate { .. } => {
```
utpilla: The arm matches `{ .. }` and then re-destructures the same value with a `let-else` + `unreachable!`, which is a bit hard to follow. You can bind the fields directly in the arm and silence the unused-binding warning for the disabled build with an arm-level attribute, dropping the `unreachable!` branch entirely:

```rust
#[cfg_attr(not(feature = "certificate-auth"), allow(unused_variables))]
AuthMethod::Certificate { path, password } => {
    #[cfg(not(feature = "certificate-auth"))]
    return Err(GenevaConfigClientError::Certificate("...".to_string()));
    #[cfg(feature = "certificate-auth")]
    {
        // use `path` / `password` directly
    }
}
```

opentelemetry-exporter-geneva/geneva-uploader/Cargo.toml
```diff
@@ -65,9 +65,13 @@ tls-rustls = [
...
     "dep:rustls-pki-types",
     "dep:rustls-native-certs",
-    "dep:p12-keystore",
     "reqwest/rustls-tls-native-roots-no-provider",
 ]
+# Enables password-protected PKCS#12 client certificate authentication. This is
+# intentionally disabled by default so builds that use managed identity,
+# workload identity, or another non-certificate method do not compile PKCS#12
+# parsing and cryptographic dependencies.
+certificate-auth = ["dep:p12-keystore"]
```
utpilla, line 74: `p12-keystore` is only referenced from `build_rustls_client_config` (`#[cfg(all(certificate-auth, tls-rustls))]`); the native-tls path uses `native_tls::Identity::from_pkcs12` instead. So a `certificate-auth` + `tls-native` build compiles `p12-keystore` without ever using it.

I don't think there's a clean Cargo fix — features can't express a "both flags" intersection, and moving the dep onto `tls-rustls` would pull it into rustls-only builds that don't use cert auth (a broader waste). So attaching it to certificate-auth is probably the least-bad option. Flagging only so the coupling is a deliberate, documented choice — a one-line comment here noting "only used by the rustls cert path" would help future readers.

utpilla (approved):

PR #747, opened 2026-08-12
feat(geneva): support multi-moniker routing

lalitb:
## Summary

- add required account routing with one default logical group and optional final-event-name overrides
- retain all logical-group-to-primary-moniker mappings returned by the configuration service
- route each encoded batch by its final event name, then resolve that group to its physical ingestion moniker
- enforce exact, case-sensitive group matching and exactly one primary moniker per group
- expose the same routing model through the C configuration API and agent-fed credentials

## Why

A configuration can contain several logical account groups. Physical moniker names are routing results and are not reliable selectors, so inspecting their names or depending on response ordering can send telemetry to the wrong destination.

The new flow is deterministic:

1. event-name processing produces the final event name and encoded batch
2. account routing maps that event name to a logical group, falling back to the configured default
3. the current configuration snapshot resolves the logical group to its primary physical moniker
4. the upload uses that moniker while sharing the snapshot's endpoint and token

## Breaking changes

- Rust configuration now requires `account_routing: AccountRouting`
- C configuration now requires `account_group` as the default and optionally accepts `account_group_mapping`
- agent-fed credentials now provide a logical-group-to-primary-moniker map

## Validation

- `cargo test -p geneva-uploader --all-features`
- `cargo test -p geneva-uploader-ffi --all-features`
- `cargo check -p stress --all-targets --all-features`
- `cargo clippy -p geneva-uploader -p opentelemetry-exporter-geneva -p geneva-uploader-ffi -p stress --all-targets --all-features -- -D warnings`
- `cargo fmt --all -- --check`

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader-ffi/examples/log_records_example.c
+++ b/opentelemetry-exporter-geneva/geneva-uploader-ffi/examples/log_records_example.c
@@ -10,6 +10,7 @@
  *   GENEVA_ENVIRONMENT           e.g. Test
  *   GENEVA_ACCOUNT               e.g. myaccount
  *   GENEVA_NAMESPACE             e.g. mynamespace
+ *   GENEVA_ACCOUNT_GROUP         e.g. mynamespacediag (optional for a single-group response)
  *   GENEVA_REGION                e.g. eastus
  *   GENEVA_CONFIG_MAJOR_VERSION  e.g. 2
  *
@@ -57,6 +58,7 @@ int main(void) {
     const char* environment   = getenv("GENEVA_ENVIRONMENT");
     const char* account       = getenv("GENEVA_ACCOUNT");
     const char* namespaceName = getenv("GENEVA_NAMESPACE");
+    const char* accountGroup  = getenv("GENEVA_ACCOUNT_GROUP");
     const char* region        = getenv("GENEVA_REGION");
     const char* cfg_ver_str   = getenv("GENEVA_CONFIG_MAJOR_VERSION");
 
@@ -103,6 +105,7 @@ int main(void) {
         .environment          = environment,
         .account              = account,
         .namespace_name       = namespaceName,
+        .account_group        = accountGroup,
         .region               = region,
         .config_major_version = (uint32_t)cfg_ver,
         .auth_method          = auth_method,
--- a/opentelemetry-exporter-geneva/geneva-uploader-ffi/examples/log_records_example.c
+++ b/opentelemetry-exporter-geneva/geneva-uploader-ffi/examples/log_records_example.c
@@ -10,7 +10,7 @@
  *   GENEVA_ENVIRONMENT           e.g. Test
  *   GENEVA_ACCOUNT               e.g. myaccount
  *   GENEVA_NAMESPACE             e.g. mynamespace
- *   GENEVA_ACCOUNT_GROUP         e.g. mynamespacediag (optional for a single-group response)
+ *   GENEVA_ACCOUNT_GROUP         e.g. mynamespacediag (default logical account group)
  *   GENEVA_REGION                e.g. eastus
  *   GENEVA_CONFIG_MAJOR_VERSION  e.g. 2
  *
@@ -62,12 +62,13 @@ int main(void) {
     const char* region        = getenv("GENEVA_REGION");
     const char* cfg_ver_str   = getenv("GENEVA_CONFIG_MAJOR_VERSION");
 
-    if (!endpoint || !environment || !account || !namespaceName || !region || !cfg_ver_str) {
+    if (!endpoint || !environment || !account || !namespaceName || !accountGroup || !region || !cfg_ver_str) {
         fprintf(stderr, "Missing required environment variables:\n"
                         "  GENEVA_ENDPOINT\n"
                         "  GENEVA_ENVIRONMENT\n"
                         "  GENEVA_ACCOUNT\n"
                         "  GENEVA_NAMESPACE\n"
+                        "  GENEVA_ACCOUNT_GROUP\n"
                         "  GENEVA_REGION\n"
                         "  GENEVA_CONFIG_MAJOR_VERSION\n");
         return 1;
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
@@ -26,6 +26,7 @@ use opentelemetry_proto::tonic::resource::v1::Resource;
 use opentelemetry_proto::tonic::trace::v1::{ResourceSpans, Span};
 use otap_df_pdata_views::views::logs::LogsDataView;
 use secrecy::{ExposeSecret, SecretString};
+use std::collections::HashMap;
 use std::fmt;
 use std::future::Future;
 use std::pin::Pin;
@@ -38,6 +39,7 @@ use tracing::{debug, info};
 #[derive(Debug, Clone, PartialEq)]
 pub struct EncodedBatch {
     pub event_name: String,
+    pub account_group: String,
     pub(crate) data: Vec<u8>,
     pub(crate) metadata: crate::payload_encoder::central_blob::BatchMetadata,
     pub row_count: usize,
@@ -58,6 +60,7 @@ pub struct GenevaClientConfig {
     pub environment: String,
     pub account: String,
     pub namespace: String,
+    pub account_routing: AccountRouting,
     pub region: String,
     pub config_major_version: u32,
     pub auth_method: AuthMethod,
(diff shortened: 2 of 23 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -70,6 +73,50 @@ pub struct GenevaClientConfig {
     pub obo_event_map: Option<OboEventMap>, // Per-event OBO config (None = no OBO)
 }
 
+/// Maps each final Geneva event name to one logical GCS account group.
+#[derive(Clone, Debug)]
+pub struct AccountRouting {
+    pub default_group: String,
+    pub groups_by_event: HashMap<String, String>,
```
utpilla: Make  AccountRouting  fields private so the config stays additive.\

Right now the fields are `pub` and the struct is built by literal (here, plus the FFI shim and tests). If we later add a tier for e.g. routing by scope/resource attribute for the passthrough case , adding a field becomes a source-breaking change for everyone constructing the literal, even though the behavior is purely additive.

It also lets the builder enforce the non-empty invariants at construction. The only call site that needs a tweak is the FFI, which builds from a whole `HashMap`.

Add a bulk builder method for it:
```rust
pub fn with_event_groups(mut self, map: HashMap<String, String>) -> Self {
    self.groups_by_event.extend(map);
    self
}
// FFI:
account_routing: AccountRouting::new(account_group).with_event_groups(account_group_mapping),
```

Net: same behavior today but adding the scope tier or pass-through later is a clean, non-breaking add.

```suggestion
    default_group: String,
    groups_by_event: HashMap<String, String>,
```

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
```diff
@@ -372,13 +373,18 @@ impl GenevaUploader {
         let ResolvedIngestion {
             token: auth_token,
             gig_endpoint,
-            moniker,
+            primary_monikers,
             endpoint_query_param,
         } = self.source.resolve().await?;
+        let moniker = primary_monikers.get(account_group).ok_or_else(|| {
+            GenevaUploaderError::InternalError(format!(
```
utpilla: Unknown account group fails per-upload as `InternalError`  and doesn't say what's valid.

If `default_group` or a mapped group isn't in the GCS response, this only surfaces here, per batch, after encoding as `InternalError`, which collapses to `UploadError::Other` and looks retriable for a permanent config error.

Two suggestions:
1. Include the available groups in the message (like `select_primary_monikers` does) —casing/typo mismatches are the likely cause and this makes them debuggable.
2. Use a dedicated, non-retriable error variant (e.g.  `AccountGroupNotResolved` ) instead of `InternalError`.

Optional but ideal: validate the configured groups against the resolved set on first successful fetch, so misconfig is caught before data flows

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
utpilla: This doc comment still references `MonikerInfo` (`Result<IngestionGatewayInfo, MonikerInfo>`), which was removed in this PR. It now returns `HashMap<String, String>` of account group → primary moniker.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
```diff
@@ -664,9 +722,19 @@ impl GenevaClient {
             batches.extend(encoded);
         }
 
+        self.assign_account_groups(&mut batches);
         Ok(batches)
     }
 
+    fn assign_account_groups(&self, batches: &mut [EncodedBatch]) {
```
utpilla, line 758: Please document that routing keys on the final (post- event_name_mapping ) event name, and that unmapped events fall to  default_group  silently.

Two things worth a doc comment note:
•  groups_by_event  keys must be the destination event names, not source names — if  event_name_mapping  renames events, this map must use the renamed names.
• Resolution is fail-open: any event not in the map goes to  default_group . For a security→default misroute that's silent. Calling this out (and noting the mitigation: give differently-routed streams distinct event names) helps operators avoid it

opentelemetry-exporter-geneva/opentelemetry-exporter-geneva/examples/README.md
```diff
@@ -236,6 +237,7 @@ Expected log output should show:
 | `GENEVA_ENVIRONMENT` | Environment name in Geneva | `Test` |
 | `GENEVA_ACCOUNT` | Geneva monitoring account | `PipelineAgent2Demo` |
 | `GENEVA_NAMESPACE` | Geneva namespace | `PAdemo2` |
+| `GENEVA_ACCOUNT_GROUP` | Logical GCS account group; may be omitted only when GCS returns one group | `pademo2diag` |
```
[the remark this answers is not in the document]
lalitb: Fixed. Updated the environment-variable table to document `GENEVA_ACCOUNT_GROUP` as required, consistent with the Rust examples and `AccountRouting` validation

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/README.md
```diff
@@ -1,6 +1,6 @@
 ### GenevaConfigClient Flow (Certificate-Based Authentication)
 
-The diagram below illustrates how the `GenevaConfigClient` is initialized with a client certificate (in PKCS#12 format) and then used to fetch ingestion information from the Geneva Config Service using mutual TLS (mTLS). It includes the flow for loading the certificate, handling cached tokens, making authenticated requests, and parsing the response (including primary diagnostic monikers).
+The diagram below illustrates how the `GenevaConfigClient` is initialized with a client certificate (in PKCS#12 format) and then used to fetch ingestion information from the Geneva Config Service using mutual TLS (mTLS). It includes the flow for loading the certificate, handling cached tokens, making authenticated requests, and parsing the response. The client resolves the ingestion moniker from the primary `StorageAccountKeys` entry whose `AccountGroupName` exactly and case-sensitively matches the configured logical account group. When no group is configured, selection succeeds only if GCS returns one distinct group. Physical moniker names are never used to infer the account group.
```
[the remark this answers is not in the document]
lalitb: Fixed. The README now describes the config client returning the complete account-group-to-primary-moniker map and validating exactly one primary per distinct account group. I also updated the diagram’s success and error paths.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
lalitb: Fixed. The fields are now private, and I added `with_event_groups` for bulk construction. The FFI shim now uses `AccountRouting::new(account_group).with_event_groups(account_group_mapping)`.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
```diff
@@ -811,7 +841,7 @@ impl GenevaConfigClient {
     #[allow(dead_code)]
     pub(crate) async fn get_ingestion_info(
         &self,
-    ) -> Result<(IngestionGatewayInfo, MonikerInfo, String)> {
+    ) -> Result<(IngestionGatewayInfo, HashMap<String, String>, String)> {
```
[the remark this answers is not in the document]
lalitb: Fixed. Updated the `# Returns` section to document the current `(IngestionGatewayInfo, HashMap<String, String>, String)` return value, including the account-group-to-primary-moniker map and endpoint query parameter.

opentelemetry-exporter-geneva/geneva-uploader-ffi/examples/log_records_example.c
```diff
@@ -10,6 +10,7 @@
  *   GENEVA_ENVIRONMENT           e.g. Test
  *   GENEVA_ACCOUNT               e.g. myaccount
  *   GENEVA_NAMESPACE             e.g. mynamespace
+ *   GENEVA_ACCOUNT_GROUP         e.g. mynamespacediag (optional for a single-group response)
```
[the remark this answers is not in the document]
lalitb: Fixed. `GENEVA_ACCOUNT_GROUP` is now documented as required, included in the runtime required-variable check, and listed in the missing-variable error output.

opentelemetry-exporter-geneva/geneva-uploader/src/ingestion_service/uploader.rs
lalitb: Fixed. An unresolved group now returns the dedicated non-retriable `UploadError::AccountGroupNotResolved` variant. The error includes the requested group and a deterministically sorted list of available groups. I also added focused coverage for this path.

I kept eager first-fetch validation out of this change because credentials can be refreshed dynamically. The upload-time error now classifies the permanent configuration problem explicitly.

opentelemetry-exporter-geneva/geneva-uploader/src/config_service/client.rs
lalitb: Fixed. Removed the stale `MonikerInfo` reference and documented the current account-group-to-primary-moniker map return value.

opentelemetry-exporter-geneva/geneva-uploader/src/client.rs
lalitb: Fixed. The routing documentation now states that overrides use the final post-`event_name_mapping` destination name and that unmatched events silently fall back to `default_group`. It also recommends distinct destination event names for streams that require different routing.

utpilla (approved):

PR #754, opened 2026-08-21
test(tower): assert no telemetry and context propagation when signals disabled

jan-xyz:
## Summary

Targeted tests for `with_tracing(false)` / `with_metrics(false)` on `http::server::LayerBuilder`, as requested by @cijothomas in [#679](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/679).

### Tests added

- **`test_with_tracing_false_still_propagates_context`** — sets a global `TraceContextPropagator`, injects a parent span's context into request headers, calls `with_tracing(false)`, asserts the handler receives a valid span context (propagation works) while no spans are recorded.

All three confirm the boolean flags take precedence even when a real provider is supplied — no-op providers are used internally and no telemetry is emitted, but context propagation is unaffected.


jan-xyz:
Not stale

cijothomas (approved):

PR #755, opened 2026-08-22
chore(tower): Update tower http server benchmarks to minimize measurement noise

francoposa:
The intent here is minimize the wild variance I experienced between benchmark runs even when there are no code changes.
There are very few resources on this, so I mostly conversed with Claude and iterated through this until I could get much more stable results run after run.

Prior to these changes, I would regularly get 9-15% fluctuations between runs.

Fixes #

## Changes
**`iter_batched`, not `iter_custom`**
Primary change is swapping `iter_custom` setup where we start & stop the timer ourselves to Criterion's more standard `iter_batched`.
The actual effect that this has is that Criterion calls the setup funcs in small batches and then drops the memory before the next batch.
Claude's hypothesis was that the `iter_custom` setup built _all_ N requests for the benchmark before each run - for any non-small N, this gives the objects varying levels of cache locality through the run, causing larger timing fluctations.

In any case, the `iter_batched` with `BatchSize::SmallInput` seems to be the recommended approach for Criterion unless you have a very special case.

**Benchmark run config  suggestions**
Added notes to run with `--sample-size 1000` and `taskset` to help minimize fluctuation further.

### DIFF
--- a/opentelemetry-instrumentation-tower/benches/http_server.rs
+++ b/opentelemetry-instrumentation-tower/benches/http_server.rs
@@ -168,7 +168,7 @@ fn benchmark_http_server(c: &mut Criterion) {
 
     // Scenario 1: Baseline - no middleware
     group.bench_function(BenchmarkId::new("request", "baseline"), |b| {
-        let mut service = tower::service_fn(handler);
+        let service = tower::service_fn(handler);
         b.to_async(&rt).iter_batched(
             || (service, build_request()),
             |(mut service, req)| async move {
--- a/opentelemetry-instrumentation-tower/benches/http_server.rs
+++ b/opentelemetry-instrumentation-tower/benches/http_server.rs
@@ -58,9 +58,27 @@
 //!
 //! ## Run
 //!
-//! ```sh
+//! ```shell
 //! cargo bench --bench http_server -p opentelemetry-instrumentation-tower
 //! ```
+//!
+//! ### Minimize Benchmark Fluctuation
+//! As these benchmarks cover a very small codepath,
+//! measurements can fluctuate significantly between runs without any code changes.
+//!
+//! Minimize noise by increasing samples collected to 1000 (from the default 100):
+//! ```shell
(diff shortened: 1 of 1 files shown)


jan-xyz:
Nice! Benchmarking this also across different machines is quite the pain as well. I had multiple projects where we used to get dedicated EC2 instances for benchmarks but that's quite overkill, so I think resorting to something like this and counting allocations like you used in the past is at least a local optimization.

jan-xyz (approved):

jan-xyz:
Not stale

francoposa:
@cijothomas If we could get a quick look here, thanks

opentelemetry-instrumentation-tower/benches/http_server.rs
```diff
@@ -176,112 +185,92 @@ fn benchmark_http_server(c: &mut Criterion) {
...
                        // meter is not set, so meter instruments are already no-op
         let layer = LayerBuilder::builder().build().unwrap();
-        b.to_async(&rt).iter_custom(|iters| {
-            let layer = layer.clone();
-            async move {
-                let mut service = ServiceBuilder::new()
-                    .layer(layer)
+        b.to_async(&rt).iter_batched(
+            || {
+                let service = ServiceBuilder::new()
```
cijothomas, line 190: Currently we create one service and reuses it across requests, while the proposed one creates a *new* service per request. Is that intentional?

francoposa: Fixed this in the latest commit - it also further improves & stabilizes the benchmarks.

francoposa:
Fixed all the merge conflicts & updated benchmark numbers.
Each of the non-baseline benchmarks is faster in absolute terms with the batched approach, though baseline is higher on my machine, so the diffs shrunk.

But all in all everything is much more stable and reproducible with less variance across runs which is the point of this.

jan-xyz (approved):

cijothomas (approved):

PR #785, opened 2026-09-04
fix: [geneva-uploader] Fix Common Schema related fields

utpilla:
## Summary

Correct the Geneva uploader's OTLP log and span encoding to produce compatible Common Schema fields and values.

| Field | Expected | Observed |
|---|---|---|
| `env_name` | `Log` | `DiagnosticsProd` |
| `env_ver` | `4.0` | `Ver1v0` |
| `TIMESTAMP` | Current event time | Missing/default year 1 |
| `name` | `Log` | Missing when OTLP `event_name` was absent |
| Severity fields | `severityNumber`, `severityText` | `SeverityNumber`, `SeverityText` |

The underlying issue was that Geneva transport configuration was reused as Common Schema payload metadata. `DiagnosticsProd` and `Ver1v0` identify the GCS transport environment and event configuration version; they are not the
Common Schema `env_name` and `env_ver` values.

## Changes

- Keep Geneva transport metadata separate from Common Schema payload metadata.
- Preserve `DiagnosticsProd` and `Ver1v0` for GCS routing and upload metadata.
- Emit Common Schema version `4.0` as `env_ver`.
- Populate `env_name` from explicit `PartA.name` when present, otherwise from
  the final outgoing Geneva event/table name.
- Emit uppercase `TIMESTAMP` and give it the same event time as `env_time`.
- Emit canonical `severityNumber` and `severityText` field names.
- Populate Part B `name` from:
  1. explicit `PartB.name` for Common Schema records;
  2. OTLP `event_name` for canonical OTLP records;
  3. the final routed event/table name when no explicit name is supplied.
- Apply equivalent Part A metadata corrections to spans.
- Update the decoded-payload integration assertions and add regression coverage
  for canonical logs, routed logs, and spans.

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/CHANGELOG.md
+++ b/opentelemetry-exporter-geneva/geneva-uploader/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### Fixed
+- Emit compatible Common Schema metadata for OTLP logs and spans, including `env_ver=4.0`, event-based `env_name`, uppercase `TIMESTAMP`, canonical log severity field casing, and a default Part B log name.
+
 ## [0.7.0] - 2026-09-03
 
 ### Changed
@@ -15,7 +18,6 @@
 - Native multi-moniker routing. `AccountRouting` defines one required default logical account group and optional exact final-event-name overrides. Encoded batches carry their resolved logical group, and uploads resolve that group through the current GCS snapshot to the corresponding primary physical moniker.
 
 ### Fixed
-- Emit AMACA-compatible Common Schema metadata for OTLP logs and spans, including `env_ver=4.0`, event-based `env_name`, uppercase `TIMESTAMP`, canonical log severity field casing, and a default Part B log name.
 - Corrected a misleading startup log message: when `logs.default_event_name` is set without an `event_name_mapping`, `GenevaClient::new` now logs `Configured logs event name routing [default_event_name=...]` instead of `Logs config not initialized; using default values for log events`. The default was always applied correctly; only the log line was wrong. This makes the logs message symmetric with the equivalent spans message.
 - Resolve every GCS logical account group to exactly one primary physical
   moniker instead of inspecting physical moniker names for `diag` or depending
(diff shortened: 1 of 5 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -225,6 +227,12 @@ impl<'a> LogRecordParts<'a> {
         if let Some(routing_event_name) = routing_event_name {
             parts.routing_event_name = routing_event_name;
         }
+        if parts.env_name.is_none() {
+            parts.env_name = Some(parts.routing_event_name.clone());
```
lalitb: This clones `routing_event_name` twice. The clones are free for `Fixed` and `None` routing, but each clone allocates on the per-record passthrough path where the Cow is owned.

Could we keep these fields unset and use `routing_event_name` as the write-time fallback? `env_name` already does that at line 1244. For `name`, we could always add the schema field and use the same fallback while encoding. That preserves the new Part B behavior without the extra per-record allocations.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
lalitb: nit - this TODO can be removed perhaps with explanation ?

lalitb (approved):
LGTM. Thanks.

opentelemetry-exporter-geneva/geneva-uploader/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+### Fixed
```
lalitb, line 5: nit - Add `vNext` above this entry.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
utpilla: Addressed in 9c8cfcb0. Missing `env_name` and Part B `name` now remain unset internally and use `routing_event_name` as the write-time fallback, avoiding the per-record `Cow` clones.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
utpilla: Addressed in 9c8cfcb0. Replaced the TODO with a comment explaining that the GCS transport event version is distinct from Common Schema `env_ver`.

opentelemetry-exporter-geneva/geneva-uploader/CHANGELOG.md
utpilla: Addressed in 9c8cfcb0 by adding the `vNext` heading.

PR #786, opened 2026-09-05
feat: [geneva-uploader] Include resource and scope attributes

utpilla:
## Summary

Include OTLP resource and instrumentation-scope attributes in Geneva Common Schema log payloads.

## Changes

- Include resource attributes in every associated log record.
- Include instrumentation-scope attributes in every associated log record.
- Resolve duplicate keys using precedence:
`log-record attributes > instrumentation-scope attributes > resource attributes`
- Prevent resource and scope attributes from replacing fixed Common Schema fields such as `name`, `body`, `severityNumber`, and `TIMESTAMP`.
- Ignore resource and scope attributes whose names conflict with fixed Common Schema Part A or Part B fields

### DIFF
--- a/opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
+++ b/opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
@@ -34,6 +34,8 @@ const CS_VERSION_4_DISPLAY: &str = "4.0";
 const KEY_CSVER: &str = "__csver__";
 const KEY_PARTB_TYPENAME: &str = "PartB._typeName";
 const CS_LOG_TYPENAME: &str = "Log";
+const ATTR_SERVICE_NAME: &str = "service.name";
+const ATTR_SERVICE_INSTANCE_ID: &str = "service.instance.id";
 
 const FIELD_ENV_NAME: &str = "env_name";
 const FIELD_ENV_VER: &str = "env_ver";
@@ -121,14 +123,14 @@ impl RoleOverrides {
                 continue;
             };
             match key {
-                "service.name" if overrides.role.is_none() => {
+                ATTR_SERVICE_NAME if overrides.role.is_none() => {
                     overrides.role = attr.value().and_then(|value| {
                         value_as_utf8(&value)
                             .filter(|value| !value.trim().is_empty())
                             .map(str::to_owned)
                     });
                 }
-                "service.instance.id" if overrides.role_instance.is_none() => {
+                ATTR_SERVICE_INSTANCE_ID if overrides.role_instance.is_none() => {
                     overrides.role_instance = attr.value().and_then(|value| {
                         value_as_utf8(&value)
                             .filter(|value| !value.trim().is_empty())
@@ -171,6 +173,12 @@ impl AttributeTemplate {
         let mut template = Self::default();
         if let Some(resource) = resource {
             for attr in resource.attributes() {
+                if matches!(
+                    std::str::from_utf8(attr.key()),
+                    Ok(ATTR_SERVICE_NAME | ATTR_SERVICE_INSTANCE_ID)
+                ) {
+                    continue;
+                }
                 template.upsert_attribute(attr, false);
             }
         }
@@ -3070,6 +3078,8 @@ mod tests {
                 resource: Some(Resource {
                     attributes: vec![
                         string_attr("microsoft.resourceId", "/subscriptions/test"),
+                        string_attr(ATTR_SERVICE_NAME, "checkout"),
+                        string_attr(ATTR_SERVICE_INSTANCE_ID, "instance-1"),
                         string_attr("shared", "resource"),
                         string_attr("scope-wins", "resource"),
                         string_attr(FIELD_BODY, "resource-body"),
@@ -3130,6 +3140,8 @@ mod tests {
         assert!(dynamic_names.contains(&"microsoft.resourceId"));
         assert!(dynamic_names.contains(&"scope.attribute"));
         assert!(dynamic_names.contains(&"record.attribute"));
+        assert!(!dynamic_names.contains(&ATTR_SERVICE_NAME));
+        assert!(!dynamic_names.contains(&ATTR_SERVICE_INSTANCE_ID));
         assert_eq!(
             dynamic_names
                 .iter()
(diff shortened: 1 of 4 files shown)


opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -2728,6 +3048,105 @@ mod tests {
         assert_eq!(dup_row.len() - base_row.len(), 16);
     }
 
+    #[test]
+    fn resource_and_scope_attributes_are_emitted_with_amaca_precedence() {
```
lalitb: nit - remove `amaca` internal jargon ?

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -1005,8 +1141,10 @@ impl OtlpEncoder {
                 .as_ref()
                 .map(RoleOverrides::from_resource)
                 .unwrap_or_default();
+            let resource_attributes = AttributeTemplate::from_resource(resource.as_ref());
```
lalitb, line 1147: nit - Add a TODO to make this resource and scope enrichment configurable, default as true.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -343,6 +439,29 @@ impl<'a> LogRecordParts<'a> {
...
+            if record_has_key {
+                continue;
+            }
+
+            let value_start = self.dynamic_values.len();
+            self.dynamic_values.extend_from_slice(
+                &enrichment.values[field.value_start..field.value_start + field.value_len],
+            );
+            self.dynamic_fields.push(DynamicField {
+                name: field.name.clone(),
```
lalitb, line 459: `field.name` is an owned `Cow`, so this clone allocates once for every inherited attribute on every log record. `finish_fields` clones the same name again at line 528, even when the schema already exists.

The template avoids decoding the resource and scope attributes again, but the keys still allocate twice per attribute per record. Could we share the names with something like `Arc<str>`, or borrow them from the scope template while the record is encoded?

This can be a followup, not blocking this PR.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -2728,6 +3048,105 @@ mod tests {
...
+        let ctx = LogEncodeContext {
+            metadata_fields: &metadata,
+            routing: LogRoutingContext {
+                table_name: "Log",
+                resource_role: &role_overrides,
+                scope_routing: &scope_routing,
+            },
+        };
+
+        let parts = LogRecordParts::new_with_enrichment(&record, &ctx, &enrichment);
```
lalitb, line 3135: This test checks the intermediate `dynamic_fields` and `dynamic_values`, then stops before schema construction and row encoding.

Could we add resource and scope attributes to `geneva_test_server_integration.rs` and assert the decoded values there, including record-over-scope-over-resource precedence? That would verify the payload users receive and cover the three changed lines Codecov currently reports as missed.

Not blocker - can be followup.

lalitb:
Can we also update this limitation - https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/ef15796c7a0bc9f3a5af439a52d253f9a6a15737/opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs#L1944-L1950

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
```diff
@@ -157,6 +160,67 @@ impl DynamicField {
...
+        };
+        if is_reserved_log_field(key, is_scope_attribute) {
+            return;
+        }
+        let Some(type_id) = bond_type_for_value(&value) else {
+            return;
+        };
+
+        let value_start = self.values.len();
+        write_view_value(&mut self.values, &value, type_id);
```
lalitb, line 215: nit - When a scope attribute replaces a resource attribute with the same key, the new value is appended and the old encoded bytes remain unused in `AttributeTemplate::values`.

The waste is bounded to the template and those bytes do not reach each row, so this is a nit. It may be cleaner to keep each encoded value with its field, which would make replacement drop the old value naturally.

Non-blocking, can be a followup.

lalitb (approved):
thanks. LGTM. Also note that this PR is on top of previous PR, so need to rebase once that is merged.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
utpilla: Updated.

utpilla:
> Can we also update this limitation -
> 
> https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/ef15796c7a0bc9f3a5af439a52d253f9a6a15737/opentelemetry-exporter-geneva/geneva-uploader-ffi/src/lib.rs#L1944-L1950

Updated in https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/786/changes/5302bcbc6d70a7dc47f37df68c0dee111ea99c39

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
utpilla: Added resource and scope attributes to the test-server integration test and asserted the decoded values, including record-over-scope-over-resource precedence.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
utpilla: The unused bytes are bounded to the scope template and are not copied into emitted rows. I’d prefer to address this in a follow-up by storing each encoded value with its template field so replacement naturally drops the previous value.

opentelemetry-exporter-geneva/geneva-uploader/src/payload_encoder/otlp_encoder.rs
utpilla: The inherited attribute names are currently cloned when added to each record and again during schema-field construction. A clean fix requires sharing the names or adjusting the relevant lifetimes, so I’d prefer to handle this as a follow-up as suggested.

PR #787, opened 2026-09-05
fix: allow manual_ok_err in detector log helpers when internal-logs is off

chinmaychahar:
## Issue

The three AWS detectors have helpers - `debug_on_error` / `info_on_error` / `warn_on_error` that take a result and log error if there is one. This logging line only exists when the **internal-logs** feature is on but when it's off, the helper is left doing nothing but return the value on success and None on error. Clippy is failing the build with the `manual_ok_err` lint because of this

## Change 

Allow `manual_ok_err` but only in the case where internal-logs is off. This reuses the `#[cfg_attr(...)]` line that's already there for the same reason


chinmaychahar:
cc: @cijothomas

cijothomas (approved):

PR #788, opened 2026-09-07
fix(tower): emit the missing HTTP server span attributes

### LINKED ISSUE #777: [instrumentation-tower] Emit missing HTTP server semantic convention attributes
### Description

The centralized semantic-conventions conformance work in
open-telemetry/semantic-conventions-conformance#140 exercises
`opentelemetry-instrumentation-tower` 0.18.0 through its Axum integration and
identifies several HTTP server semantic-convention gaps.

The instrumentation successfully emits server spans and all four standard HTTP
server metrics, but the conformance run reports the following missing span
attributes:

- `url.query` is absent for a request to
  `/users/456?fields=name&verbose=true`. This attribute is conditionally
  required when the request target contains a query component.
- `error.type` is absent from the failed server span whose status is `error`.
- `client.address` is absent. This is a recommended HTTP server span attribute.
- `server.address` is absent. This is a recommended HTTP server span attribute.
- `network.protocol.version` is absent from server spans. This is a recommended
  HTTP server span attribute. Some emitted metrics already contain it.

Central conformance PR and recorded findings:

https://github.com/open-telemetry/semantic-conventions-conformance/pull/140

### Why the existing Rust-contrib live check does not detect these gaps

The current Tower live-check workflow exercises a broad set of HTTP methods,
status codes, and route shapes, but it does not send a request containing a
query string.

It also runs Weaver's registry validation without the additional HTTP policies
used by the centralized conformance runner. Weaver validates attributes that
are present, but it cannot always infer the applicable HTTP span definition or
report attributes that were never emitted.

This issue is related to the broader live-check work tracked by #530.

### Suggested changes

- Emit `url.query` when the request target has a query component.
- Emit `error.type` when the server operation ends in an error.
- Populate `client.address`, `server.address`, and
  `network.protocol.version` when their values are available.
- Add focused instrumentation tests for these attributes.
- Add a query-string request to the Tower live-check workload.
- Consider running the shared HTTP conformance advice policies, or equivalent
  crate-local assertions, so missing attributes are detected in Rust-contrib
  CI.

### Acceptance criteria

- A query-string request produces a server span containing `url.query`.
- A failed server operation produces a span containing `error.type`.
- Server spans contain the recommended address and protocol attributes when
  available.
- The Tower live-check or crate tests prevent these attributes from regressing.

jan-xyz:
Fixes #777
Design discussion issue (if applicable)

## Changes

A conformance run against the published HTTP semantic conventions found that the server span lacked `url.query`, `client.address`, `server.address`, `network.protocol.version`, and `error.type`. A failed request was therefore hard
to attribute, and a span could not be matched to the client that sent the request. The middleware now reads these attributes, and it also reports `server.port`, `network.peer.address`, and `network.peer.port`.

A server can sit behind a reverse proxy, which rewrites the connection and the `Host` header. The conventions therefore prefer the forwarding headers. The middleware reads the `Forwarded` header first, then `X-Forwarded-For`, `X-Forwarded-Host`, and `X-Forwarded-Proto`, and falls back to the connection. A client can send these headers too, which the README states. A Tower service reads the request only, so the peer address comes from Axum `ConnectInfo`, which needs `Router::into_make_service_with_connect_info`.

Four further gaps came up while reading the specification for the reported ones:

* `url.scheme` was always empty. It read the request target, but a server receives a target in origin form, which carries no scheme. It now reports the forwarded scheme, and `http` when no header states one.
* `url.full` belongs to the client span, and the emitted value was a relative path rather than an absolute URL. It is gone from the server span.
* `network.protocol.name` was always `http`, while the conventions ask for it only when the protocol is not `http`. It is gone from the metrics.
* An unknown request method passed through verbatim, into both the attribute and the span name. It now reports `_OTHER`, keeps the original value in `http.request.method_original`, and produces a span name of `HTTP {route}`.

The span status of a 5xx response no longer repeats the status code in its description, because `error.type` carries it, and the description of a failed request no longer doubles as the error type.

`network.protocol.version` reports `2` and `3` in place of `2.0` and `3.0`, which matches the examples in the conventions and the actix-web instrumentation.

The live-check workload now sends a request with a query string and requests with both forms of forwarding header, so a missing attribute fails CI. Running the shared conformance advice policies here is a larger piece of work and belongs to the live-check effort in #530. An override of the known-method list through `OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS`, and the redaction of sensitive query parameters, are both still open.

The client layer in #700 shares the attribute code and needs the same sweep. It follows once this lands, as agreed in #777.

### DIFF
--- a/.github/workflows/weaver-live-check-tower.yml
+++ b/.github/workflows/weaver-live-check-tower.yml
@@ -103,6 +103,17 @@ jobs:
           echo "::group::Drive traffic"
           c "http://127.0.0.1:5000/"
           c "http://127.0.0.1:5000/users/1"
+          # A query component makes `url.query` applicable.
+          c "http://127.0.0.1:5000/users/1?fields=name&verbose=true"
+          # The forwarding headers make the request look like it came through a
+          # reverse proxy, so `client.address`, `server.address` and
+          # `url.scheme` describe the original client request.
+          c -H 'Forwarded: for=203.0.113.7;host=public.example.com:8443;proto=https' \
+            "http://127.0.0.1:5000/users/2"
+          c -H 'X-Forwarded-For: 203.0.113.8' \
+            -H 'X-Forwarded-Host: public.example.com' \
+            -H 'X-Forwarded-Proto: https' \
+            "http://127.0.0.1:5000/users/3?fields=name"
           c "http://127.0.0.1:5000/files/a/b/c/1.txt"
           c -X POST   "http://127.0.0.1:5000/items"
           c -X PUT    "http://127.0.0.1:5000/items/1"
--- a/opentelemetry-instrumentation-tower/CHANGELOG.md
+++ b/opentelemetry-instrumentation-tower/CHANGELOG.md
@@ -13,6 +13,22 @@
 * The instrumentation scope now carries the OpenTelemetry semantic conventions
   schema URL.
   [#679](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/679)
+* The server span now carries the HTTP server span attributes that it missed:
+  `url.query`, `network.protocol.version`, `client.address`, `server.address`,
+  `server.port`, `network.peer.address`, `network.peer.port`, and `error.type`.
+  The middleware reads the attributes of the original client request from the
+  `Forwarded` header, then from the `X-Forwarded-For`, `X-Forwarded-Host`, and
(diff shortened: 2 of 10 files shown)


jan-xyz:
Closing this in favour of a series of smaller pull requests. This change touched too many attributes at once to review well. I will open one pull request per attribute, starting with `url.query`.

PR #789, opened 2026-09-07
fix(tower): record the query string of a server request

jan-xyz:
Fixes part of #777

This is the first of a series of small pull requests for #777, I decided to keep the PRs small and focused to better see what changes and make it much easier to review.

## Changes

A server span left out the query string of the request. The HTTP semantic conventions require `url.query` when the request target carries a query component, and the conformance run in open-telemetry/semantic-conventions-conformance#140 reported it missing. An empty query component, as in `/users/456?`, yields `url.query` with an empty value.

The live check did not send a single request with a query string, which is why its Weaver run never noticed the gap; it does now, which also serves the CI compliance goal in #530.


opentelemetry-instrumentation-tower/src/http/server.rs
```diff
@@ -484,6 +484,16 @@ where
             KeyValue::new(semconv::attribute::URL_FULL, req.uri().to_string()),
         ];
 
+        // `url.query` is required when the request target carries a query
+        // component. An empty component still counts as one, so the presence of
+        // the value decides, not its content.
+        if let Some(query) = req.uri().query() {
+            span_attributes.push(KeyValue::new(
```
cijothomas, line 549: we need to do some redaction/omit known sensitive keys.

jan-xyz: I decided to remove `url.full` for the redaction, it isn't part of the server semantic convention and otherwise would've also needed redaction.

jan-xyz:
@cijothomas I was looking at the actix-web-instrumentation for inspiration and noticed that it doesn't cover query redaction, I think they can both share the same logic to have performance and correctness improvements benefit both crates, as well as other http instrumentations in the future. What do you think?

cijothomas:
> @cijothomas I was looking at the actix-web-instrumentation for inspiration and noticed that it doesn't cover query redaction, I think they can both share the same logic to have performance and correctness improvements benefit both crates, as well as other http instrumentations in the future. What do you think?

Agreed, sharing this would make sense. For now, let’s just duplicate the code in Tower and Actix. Before stabilizing, we can figure out a good place for the shared logic and remove the duplication.

opentelemetry-instrumentation-tower/src/common/attributes.rs
```diff
@@ -64,3 +66,192 @@ pub(crate) fn split_and_format_protocol_version(
...
 }
+
+/// Query parameter keys whose values the conventions ask instrumentations to
+/// redact by default.
+pub(crate) const DEFAULT_SENSITIVE_QUERY_PARAMETERS: &[&str] = &[
+    "X-Amz-Signature",
+    "X-Amz-Credential",
+    "X-Amz-Security-Token",
+    "sig",
+    "X-Goog-Signature",
```
cijothomas, line 77: [X-Amz-Signature](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-authentication-methods.html)
[X-Amz-Credential](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-authentication-methods.html)
[X-Amz-Security-Token](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-authentication-methods.html)
[sig](https://learn.microsoft.com/azure/storage/common/storage-sas-overview#sas-token)
[X-Goog-Signature](https://cloud.google.com/storage/docs/access-control/signed-urls)

https://opentelemetry.io/docs/specs/semconv/http/http-spans/ has 2 more that we are missing. Could we keep same default?

jan-xyz: Which one do you see? For me it only shows:

> Query string values for the following keys SHOULD be redacted by default and replaced by the value REDACTED:
> 
> X-Amz-Signature
> X-Amz-Credential
> X-Amz-Security-Token
> sig
> X-Goog-Signature
> 
> This list is subject to change over time.

cijothomas: https://github.com/open-telemetry/semantic-conventions/pull/3989
Sorry they were removed but then restored!

opentelemetry-instrumentation-tower/README.md
```diff
@@ -16,9 +16,26 @@
...
+conventions. The goal is every `Required`, `Conditionally Required`, and
+`Recommended` attribute that a Tower service can read. `Opt-In` attributes, such
+as `http.request.header.<key>`, are not emitted. Add them with
+`http::server::LayerBuilder::with_request_extractor`.
+
+## Query string redaction
+
+The `url.query` attribute holds the query string of the request, which can carry
+a credential. The middleware therefore replaces the value of every query
+parameter that the conventions name, such as `sig` or `X-Amz-Signature`, with
```
cijothomas, line 33: nit: lets document the default list?

opentelemetry-instrumentation-tower/src/common/attributes.rs
```diff
@@ -64,3 +66,192 @@ pub(crate) fn split_and_format_protocol_version(
...
+    let mut redacted: Option<String> = None;
+    // Byte offset up to which `query` has been copied into `redacted`.
+    let mut copied = 0;
+    // Byte offset of the parameter under inspection.
+    let mut pair_start = 0;
+
+    for pair in query.split('&') {
+        let key = pair.split('=').next().unwrap_or(pair);
+
+        if sensitive.iter().any(|parameter| parameter.as_ref() == key) {
```
cijothomas, line 98: would it be faster if we stored `sensitive` as hashset instead of slice?

opentelemetry-instrumentation-tower/benches/http_server.rs
```diff
@@ -104,7 +104,7 @@ fn build_requests(n: u64) -> Vec<http::Request<String>> {
         .map(|_| {
             http::Request::builder()
                 .method("GET")
-                .uri("http://example.com/users/123")
+                .uri("http://example.com/users/123?fields=name&sig=secret-signature")
```
cijothomas, line 107: lets leave the existing one as-is, and if needed, add a new one which can measure the cost, when there is one/more sensitive key in query string.

cijothomas (approved):
Looks good. Left few minor comments, okay to follow up separately than this PR.

opentelemetry-instrumentation-tower/src/common/attributes.rs
jan-xyz: ah yes, that's a good idea.

cijothomas:
Quueing to merge to unblock progress. The sggestions can be followed up separately.

PR #793, opened 2026-09-09
feat: Expose user-events trace processor

cijothomas:
## Changes

Expose a public `Processor` and `ProcessorBuilder` for the user-events trace exporter instead of adding the processor directly through a tracer-provider extension. This allows applications to wrap the processor with filtering or other custom `SpanProcessor` implementations before attaching it to an `SdkTracerProvider`. It also makes the trace exporter configuration consistent with `opentelemetry-user-events-logs`, which already exposes a processor builder for explicit provider attachment.

Invalid provider configuration is now returned to the caller. The obsolete reentrant processor and its `futures-executor` dependency are removed, and examples and benchmarks use the explicit processor API.

A Linux user_events integration test verifies that a wrapping processor suppresses filtered spans, exports accepted spans, and preserves resource mapping through `set_resource`.


opentelemetry-user-events-trace/src/trace/processor.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    fn on_end(&self, span: SpanData) {
+        let _ = self.exporter.export_span(&span);
+    }
+
+    fn force_flush(&self) -> OTelSdkResult {
+        Ok(())
+    }
+
+    fn shutdown(&self) -> OTelSdkResult {
```
lalitb, line 95: Could we make shutdown wait for exports already in progress before unregistering the tracepoint? The old processor held the same mutex during export and shutdown. Now shutdown can unregister the tracepoint while another thread is still serializing a span, causing its write to fail and the span to be dropped.

Please preserve that ordering without serializing exports against each other, and add a regression test for shutdown during an active export.

lalitb (approved):
Approving with the understanding that the shutdown race will be fixed before merge.

opentelemetry-user-events-trace/src/trace/processor.rs
cijothomas: This is intentional to avoid coordination on the hot path. If shutdown is called from another thread while an export is in progress, that record could be lost. This is also how user-events logs and ETW logs/traces behave today.

Making shutdown wait would require bookkeeping on every export, and I’m not sure that overhead is worth paying for this edge case. Applications needing lossless shutdown should stop producing telemetry before shutting down the provider.

Does that sound reasonable? 
(The previous behavior avoid this issue, but took a lock which caused contention and was not desired.)

Des-cloud (approved):

opentelemetry-user-events-trace/src/trace/processor.rs
cijothomas: resolving as per offline discussion.

PR #799, opened 2026-09-16
feat(aws): add Elastic Beanstalk resource detector

chinmaychahar:
Adds a `BeanstalkResourceDetector`

## Changes

- New `BeanstalkResourceDetector`. Reads the X-Ray `environment.conf` and sets details such as cloud.provider, cloud.platform, service.name etc.
- Reuses the shared `detector::utils` helpers
- README and CHANGELOG updated


chinmaychahar:
I have set service.name as `aws_elastic_beanstalk` to match the Python/JS detectors. Happy to drop if we prefer not to override a user config

chinmaychahar:
cc: @cijothomas if you could help with a review. We have this beanstalk detector present in other SDKs (Go/Python etc.) and collector but not in Rust

chinmaychahar:
gentle ping ^^ @cijothomas 😄

opentelemetry-aws/src/detector/beanstalk.rs
```diff
@@ -0,0 +1,196 @@
...
+        let deployment_id = conf.deployment_id.and_then(json_scalar_to_string);
+        let attribute_options = [
+            Some(KeyValue::new(semco::CLOUD_PROVIDER, "aws")),
+            Some(KeyValue::new(
+                semco::CLOUD_PLATFORM,
+                "aws_elastic_beanstalk",
+            )),
+            Some(KeyValue::new(semco::SERVICE_NAME, "aws_elastic_beanstalk")),
+            opt_kv(semco::SERVICE_INSTANCE_ID, deployment_id),
+            opt_kv(semco::SERVICE_NAMESPACE, conf.environment_name),
```
lalitb: Could we use `deployment.id` and `deployment.environment.name` here, as the current Go detector does? Multiple running instances can share a deployment ID, so using it as `service.instance.id` makes them appear to be the same instance. Please update the attribute mapping tests too.

opentelemetry-aws/src/detector/beanstalk.rs
```diff
@@ -0,0 +1,196 @@
...
+        };
+
+        let deployment_id = conf.deployment_id.and_then(json_scalar_to_string);
+        let attribute_options = [
+            Some(KeyValue::new(semco::CLOUD_PROVIDER, "aws")),
+            Some(KeyValue::new(
+                semco::CLOUD_PLATFORM,
+                "aws_elastic_beanstalk",
+            )),
+            Some(KeyValue::new(semco::SERVICE_NAME, "aws_elastic_beanstalk")),
```
lalitb: I'd leave `service.name` unset here. In the documented example, `Resource::builder()` reads `OTEL_SERVICE_NAME`, but adding this detector overwrites it with `aws_elastic_beanstalk`. `cloud.platform` already identifies Beanstalk. Could we remove this attribute and add a test that the configured service name is preserved?

opentelemetry-aws/src/detector/beanstalk.rs
chinmaychahar: Sure

opentelemetry-aws/src/detector/beanstalk.rs
chinmaychahar: Okay, ack

cijothomas (approved):

PR #801, opened 2026-09-18
fix(tower): record error.type on a failed server span

jan-xyz:
Fixes part of #777
Design discussion issue (if applicable) #777

## Changes

The HTTP semantic conventions mark `error.type` as conditionally required for a failed request. The tower server span does not carry this attribute. The conformance run in semantic-conventions-conformance#140 found this gap, tracked as the second item in issue #777.

A `5xx` response now sets `error.type` to the status code as a string. An inner-service error, returned before a status code exists, sets `error.type` to that error's Rust type name instead.


cijothomas:
Could you add a separate 500-response benchmark alongside the existing success-path scenarios? It would give us a baseline for the error-path instrumentation, including `error.type` on spans and metrics. Please keep the existing scenarios unchanged so their results remain comparable.

jan-xyz:
Added a `tracing+metrics-500` scenario in 99c7e106 (mirrors `tracing+metrics` but with a 500 response), and re-ran the full suite with fresh numbers for all scenarios. On this run, the extra `error.type` attribute, span status, and metric label cost about 35 ns over the 200 baseline for the same signals.

lalitb (approved):

jan-xyz:
@lalitb the protoc action had a 500 when downloading, can you retry it? I don't have permissions for it.

